Issue #7, opened 2019-01-16
Going back is odd/unintuitive


rcrdlbl:
Copying this directly from the twitter thread I started –

it feels kinda awkward that the back button/back gesture in the browser takes you away from bridge – I think this stems from the fact that bridge's UI takes up the entire page and therefore sends the message that each view is in fact a separate browser page

The only way to go back (during setup at least) is to click on the path at the top which isn’t totally unintuitive; but it isn’t the first thing most people would try — the thing they try leads to them leaving bridge then returning all the way back to the start screen

The most obvious way would be a back button next to the continue button imo – much easier to notice

galenwp:
Did we close because it's fixed, or being fixed elsewhere? I agree that this is confusing / annoying.

rmariani:
This is maybe the biggest problem with bridge IMO

urbitalex:
This is being worked on right now. Closed it to avoid any duplication of work.

Issue #6, opened 2019-01-16
Dropdown item spacing is off


Fang-:
<img width="266" alt="image" src="https://user-images.githubusercontent.com/3829764/51222200-90616e00-18f1-11e9-911d-0a34e241c368.png">

gavinpatkinson:
Firefox?

Fang-:
Yes. (^:

alexmatz:
bundled with other UI issues and assigned to @c-johnson

PR #10, opened 2019-01-18
Ga Patch 1

gavinpatkinson:
This PR addresses the following issues:

- [ ] refactor unlock/node selection forms for consistency
- [x] #7
- [x] #6
- [x] visual sigil bugs
- [x] visual layout bugs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,10 @@
 {
   "name": "bridge",
+<<<<<<< HEAD
   "version": "1.0.1",
+=======
+  "version": "1.0.0",
+>>>>>>> remove unnessecary righthand column
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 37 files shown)


gavinpatkinson:
Thanks for the review. Gonna release this EOW.

gavinpatkinson:
@jtobin can you elaborate on "Ticket and shard auth don't seem to set the wallet prop accurately; possibly something to do with async mumbojumbo?"

jtobin:
@gavinpatkinson When trying to authenticate via master ticket or shards, I observed that I'd get forwarded to the points view without a wallet prop actually being set.

I took a brief look at this and guessed that it might be due to React's asynchronous updating in some combination of `walletFromTicket` and `handleSubmit`, but that was just my initial suspicion.  I.e., perhaps we need to better enforce that `setWallet` do its thing before `popRoute` / `pushRoute`, or something.

I messed around with it for a few minutes but didn't get anywhere with it.  I can definitely take another look when I get a chance.

Fang-:
Pinging on this. I think #7 was also fixed in @c-johnson's recent router work, so there may be more conflicts here than just moved/renamed files.

gavinpatkinson:
Closing because this is now very out of date.

jtobin:
@Fang- I've ripped out the optional passphrase stuff and PR'd it in #94.

PR #18, opened 2019-01-19
Add WalletHdPath to Set Network Keys transaction scope

jkmallard:
This fixes a bug preventing Trezor users from being able to sign the Set Network Keys transaction.


jtobin:
@jkmallard LGTM!  Thanks for this one!

PR #24, opened 2019-01-21
Additional authentication methods

### LINKED ISSUE #8: Should support keyfiles
We should support importing keyfiles.  They're extremely common in practice, and they're the default for MyEtherWallet, geth, parity, and others.

They need a password to unlock.  keythereum has code you can look at to see how to process them.

jtobin:
Resolves #8.

Adds support for authenticating via Ethereum keystore files (w/password) and raw Ethereum private keys.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "bridge",
-  "version": "0.4.0",
+  "version": "1.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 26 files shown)


iamwillkim:
@jtobin I'm having some difficulty authenticating via keystore file. I enter contents of my keystore file and pw, then the decrypt option lights up, I click, but nothing happens.

jtobin:
@iamwillkim Hmm.  I actually can't reproduce:

iamwillkim:
@jtobin styling issue here and problem where if I don't enter the password to decrypt, the upload button turns orange and I can't decrypt after submitting the right password. 

image

image

iamwillkim:
@gavinpatkinson thanks for the styling, but when I upload my file, the button whites out, which is confusing:

image

gavinpatkinson:
this was existing behavior

gavinpatkinson:
fixed

jtobin:
Yeah that was me trying to be clever and re-use some state to display an error.. will have a fix up in a sec.

iamwillkim:
@jtobin nice, but we're displaying the private key in the URL (I must have missed this in the first pass), which is a no-go, even if this is run on client-side. With that fix, I think this is ready for merge

jtobin:
This is definitely now my least successful pull request of all time.

@iamwillkim **Very** nice catch -- the private key isn't actually passed in the URL *per se* (i.e. that mechanism for passing data is never used explicitly -- we always update React state according to the value of a given field), but the input component is wrapped in a HTML form component, and thus pressing return on that field will submit it as such.  Any entered form data will be displayed in the URL.

This is actually a broader bug throughout the codebase.  There are a lot of components -- the majority of them, by the look of it -- that use the same form-wrapping mechanic.  So: if someone is entering a mnemonic, shards, ticket, etc. at present, and they press return while editing a field, the form data will be displayed in the URL.

I'll get rid of the wrapping Form components -- I don't *think* they're actually required for anything (@gavinpatkinson check me if I'm wrong on that).

iamwillkim:
talked to @gavinpatkinson offline, he says that won't alter behavior; merging now.

PR #26, opened 2019-01-29
Manually follow EIP-155 in certain circumstances

### LINKED ISSUE #12: Generated transactions throw an "invalid sender" error
Trying to set networking keys for `~maldeb-hapben`. I'm using the management proxy, which is a Ledger, if that's important. According to Bridge, the chainId is 1 in the transaction; disassembling the raw tx seems to suggest the chainId is 28 (!).

jtobin:
Resolves #12.

Ledger does not seem to handle EIP-155 automatically.  When using a Ledger, if the block number is at least FORK_BLKNUM = 2675000, one needs to pre-set the ECDSA signature parameters with r = 0, s = 0, and v = chainId prior to signing.

The easiest way to handle this on our end is to just branch on the network type, since mainnet and Ropsten have obviously passed FORK_BLKNUM.

This is somewhat awkward when dealing with offline transactions, since we might want to test them on a local network as well.  For now we can assume that users generating offline transactions wish to later submit them to mainnet, but the best thing to do is probably to add an 'advanced' tab to offline transaction generation where one can disable the defaulted-on EIP-155 settings in that case.

See:

https://github.com/LedgerHQ/ledgerjs/issues/43#issuecomment-366984725

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -952,17 +952,28 @@
       "integrity": "sha512-5a6wqoJV/xEdbRNKVo6I4hO3VjyDq//8q2f9I6PBAvMesJHFauXDorcNCsr9RzvsZnaWi5NYCcfyqP1QeFHFbw=="
     },
     "@ledgerhq/hw-app-eth": {
-      "version": "4.32.0",
-      "resolved": "https://registry.npmjs.org/@ledgerhq/hw-app-eth/-/hw-app-eth-4.32.0.tgz",
-      "integrity": "sha512-d22WinjcsqJNoZSI+6UpTWZ7hl+UhL2dFeVeliCwtBWSj40z6F25MpoviGxPsv0WC7IUjayw+a9jIRcOJ5kkIw==",
+      "version": "4.35.0",
+      "resolved": "https://registry.npmjs.org/@ledgerhq/hw-app-eth/-/hw-app-eth-4.35.0.tgz",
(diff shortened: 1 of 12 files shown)


Fang-:
> This is somewhat awkward when dealing with offline transactions

Wait, don't we still give the user the "what network" dropdown when in offline mode? We should, because it affects the transaction as you've discovered.

juped:
Much obliged for fixing my issue!

jtobin:
@juped No sweat!

iamwillkim:
> > This is somewhat awkward when dealing with offline transactions
> 
> Wait, don't we still give the user the "what network" dropdown when in offline mode? We should, because it affects the transaction as you've discovered.

@Fang- Making a separate issue for this

jtobin:
This got out of sync with master, so I've rebased and force-pushed.  Should be good to go again.  I've also tacked on a fix for some issues around Etherscan links.

iamwillkim:
Bridge crashes when I try to sign the tx tro set my networking keys on localnet with my Ledger. Screencap below:

image

jtobin:
@iamwillkim As always, nice catch.  That was the previous Ledger fix which must have fallen out in a merge -- ditching the form components in #24 required a lot of manual stitching.

We should probably devote some effort to adding proptypes/tests at this point.  The Will Kim test continues to be extremely effective, but is no doubt expensive to run.

iamwillkim:
Hear hear! I'll make it an issue.

In other news, this PR looks good to me.

PR #63, opened 2019-02-26
Replace 'Urbit' with 'Arvo' where appropriate.

### LINKED ISSUE #39: "Urbit networking keys"
image

This cannot stand. Urbit is a forbidden word in Bridge. Get it out.

jtobin:
Resolves #39.

This preserves the word 'Urbit' when referencing the Urbit HD wallet, but other instances of the word are axed.

### DIFF
--- a/src/views/SetKeys.js
+++ b/src/views/SetKeys.js
@@ -306,8 +306,8 @@ class SetKeys extends React.Component {
             ? <Warning>
                 <h3 className={'mb-2'}>{'Warning'}</h3>
                 {
-                  'Once these keys have been set, your Urbit is considered ' +
-                  "'Linked'.  This operation cannot be undone."
+                  'Once these keys have been set, your point is considered ' +
+                  "'linked'.  This operation cannot be undone."
                 }
               </Warning>
             : <div /> }
--- a/src/views/Point.js
+++ b/src/views/Point.js
@@ -242,7 +242,7 @@ class Point extends React.Component {
             {
               displayReminder
                 ? <P>{`Before you can issue child points or generate your Arvo
-                  keyfile, you need to set your Urbit networking keys.`}</P>
+                  keyfile, you need to set your public keys.`}</P>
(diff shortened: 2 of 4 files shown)


src/views/SetKeys.js
```diff
@@ -306,8 +306,8 @@ class SetKeys extends React.Component {
             ? <Warning>
                 <h3 className={'mb-2'}>{'Warning'}</h3>
                 {
-                  'Once these keys have been set, your Urbit is considered ' +
-                  "'Linked'.  This operation cannot be undone."
+                  'Once these keys have been set, your ship is considered ' +
```
Fang-:
```suggestion
                  'Once these keys have been set, your point is considered ' +
```

Fang-:
> I'd argue that these all need to just say "public keys" instead of "Arvo (networking)". But perhaps that's too general for "Bridge as Arvo tool"?

To be clear, I do actually want to hear your input on this @jtobin, and an unsatisfactory defense of "arvo networking" will make me tell you to change it. (:

Probably also cc @msutherl, maybe others?

jtobin:
'Public keys' actually works for me bru.  Simpler, more general, probably more widely understood.

Fang-:
Oh, sick. Make the change then, quick, before anyone objects! (^:

msutherl:
If we wanted to be abundantly clear, these would be "Arvo authentication and encryption public keys," correct? Any reason not to be verbose?

Fang-:
> Any reason not to be verbose?

The verbose description you gave is technically correct.. for the Arvo case.

"authentication and encryption public keys" is fine, but also a mouthful. iirc we already list them as such in the details. If we don't, the extra verbosity can and should go there, but other places should keep it brief and just refer to them collectively as "public keys".

jtobin:
Going to merge this one for the time being anyway, lettuce (ha ha) discuss additional copy changes elsewhere.

PR #66, opened 2019-03-06
Copy wallet-generator code into bridge (+ bridge UI touchups)

c-johnson:
This PR brings in all the wallet-generator source code, and fixes a number of UI issues.

The basic paths for wallet-generator and bridge seem to work fine. To load walletgen, just replace the comment in src/index.js. 

A couple notes:

- I threw a week's effort into getting this all to build with gulp/rollup/sucrase. The dependencies defeated me for now. It's simpler to keep everything working with react-scripts until we run into problems with it.
- Should we merge this & other ongoing bridge work into an intermediate branch, or is merging into master and then testing before release fine? I'm fine with either way.
- New wallet-generator work should probably be added here. Not sure who else might work on it besides @gavinpatkinson and @jtobin


src/components/StatelessTransaction.js
```diff
@@ -15,6 +15,16 @@ const StatelessTransaction = props => {
   const { setUserApproval, userApproval } = props
   const { canSend, handleSubmit } = props
 
+  // setState doesn't seem to work in SetProxy/handleSubmit, so just
+  // modifying the DOM manually here. (https://imgur.com/a/i0Qsyq1)
```
Fang-: Linking to an image (and an indirect link at that) makes me feel weird. Is it possible to very briefly describe that error instead?  
Also, unsure, but maybe this needs a `//TODO investigate`.

src/lib/router.js
```diff
@@ -105,8 +107,8 @@ const renderRoute = (props, route) => {
         Nothing: () => 'Wallet',
         Just: (wal) =>
             wal.value instanceof EthereumWallet
-          ? wal.value.address
-          : addressFromSecp256k1Public(wal.value.publicKey)
+          ? (<span className="text-mono">{wal.value.address}</span>)
+          : (<span className="text-mono">{addressFromSecp256k1Public(wal.value.publicKey)}</span>)
```
Fang-: Can this not be wrapped outside of the `? :`? idk how jsx works.

src/views/Network.js
```diff
@@ -127,8 +127,8 @@ class Network extends React.Component {
           <P>
           {
             "Please select the Ethereum node you'd like to send " +
-            "transactions to.  If you don't know, the default should be " +
-            "fine.  For sensitive transactions, please select offline mode."
+            "transactions to.  For sensitive transactions, please select " +
+            "offline mode."
```
Fang-: I think this is more about the keys themselves rather than the transactions they sign? Probably "When using highly valuable keys, offline mode is advised." or something?

src/views/Ticket.js
```diff
@@ -112,9 +122,16 @@ class Ticket extends React.Component {
           <Button
             className={'mt-8'}
             prop-size={'lg wide'}
+            disabled={this.state.isUnlocking || !Maybe.Nothing.hasInstance(wallet)}
```
jtobin: Just a couple of notes re: `folktale/maybe` use:

* It's probably better to use `Maybe.Just.hasInstance(foo)` instead of `!Maybe.Nothing.hasInstance(foo)`.  I know there are a few instances of `!Maybe.Nothing.hasInstance` floating around from the last rush of code pre-initial release, but we should avoid introducing more of them.

* I'd also started to import folktale/maybe via `import { Just, Nothing } from ..` in recent contributions, which cuts down on a lot of the verbose `Maybe.`'s that pop up everywhere needlessly.

(Of course, if anyone wants to just rip this whole library out, go for it.  But I do find it very useful -- especially when its verbosity issues are addressed.)

jtobin:
One comment: it's probably worth ripping out the vendored `urbit-key-generation` code that wallet-generator uses.  I'm not sure why that was in wallet-generator to begin with (probably just to do with some last-second rushing before the WG release).  We should definitely be using the latest version of that because some bugs have been fixed in the interim, and Bridge uses it anyway for ticket/shard auth.

I also left from a minor comment on one use of `folktale/maybe`.  Otherwise broadly :+1: from me!

alexmatz:
@c-johnson Jared left a little feedback, but looks like it's almost ready!

src/lib/router.js
c-johnson: It's a bit uglier that way IMO; I try to keep the amount of logic inside JSX brackets minimal.

src/views/Network.js
c-johnson: I agree, done.

src/views/Ticket.js
c-johnson: Fixed them here and elsewhere, too.

c-johnson:
@jtobin Pulled out vendor/kg from walletgen code. I already hooked up the imports to use urbit-key-generation, so shouldn't be any conflicts.

PR #68, opened 2019-03-07
Send transaction spinner

c-johnson:
Added a disabled state and a loading spinner when the Send Transaction button is pressed.

This is slightly hacky; ideally we pass a state property in from whatever consumes StatelessTransaction, but calling setState from handleSubmit (eg in SetProxy) throws a strange error; curious of other devs have seen this or what to do about it: https://imgur.com/a/i0Qsyq1

### DIFF
--- a/src/components/StatelessTransaction.js
+++ b/src/components/StatelessTransaction.js
@@ -15,6 +15,11 @@ const StatelessTransaction = props => {
   const { setUserApproval, userApproval } = props
   const { canSend, handleSubmit } = props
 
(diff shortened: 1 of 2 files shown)


c-johnson:
(The other pull request is from a new github account I made. Some configuration thing got messed up and it changed all my old commits, but think I have it fixed now. Ignore the other one)

c-johnson:
Superceded by https://github.com/urbit/bridge/pull/70

PR #72, opened 2019-03-11
Range slider for gas price

c-johnson:
Left it with the default styling for now.

### DIFF
--- a/src/components/StatelessTransaction.js
+++ b/src/components/StatelessTransaction.js
@@ -65,7 +65,7 @@ const StatelessTransaction = props => {
 
   const gasPriceRangeDialogue = (
     <React.Fragment>
-      <div className="mt-8">Gas Price: <span className="ml-4 text-700 text-sm">{gasPrice} gwei</span></div>
+      <div className="mt-8">Gas Price: <span className="ml-4 text-700 text-sm">{(gasPrice / 1000000000).toFixed('9')} eth</span></div>
 
       <input
         className="mt-4"
--- a/src/components/StatelessTransaction.js
+++ b/src/components/StatelessTransaction.js
@@ -65,7 +65,16 @@ const StatelessTransaction = props => {
 
   const gasPriceRangeDialogue = (
     <React.Fragment>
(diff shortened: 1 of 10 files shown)


Fang-:
>gas price in eth, not gwei

Ehh, it's pretty hard to count the eight zeroes in the `eth` notation for `1 gwei`. Because these are such small units, gwei is super common. I think we should stick with it.

A solution that would still inform the user of actual, tangible value, would be to display the `gas price * gas limit` to them as "max tx fee in eth". This usually ends up being something with like two or three zeroes? Much more readable.

src/components/StatelessTransaction.js
```diff
@@ -59,6 +63,31 @@ const StatelessTransaction = props => {
...
   })
 
+  const gasPriceRangeDialogue = (
+    <React.Fragment>
+      <div className="mt-8">Gas Price: <span className="ml-4 text-700 text-sm">{(gasPrice / 1000000000).toFixed('9')} eth</span></div>
+
+      <input
+        className="mt-4"
+        type="range"
+        min="4"
```
Fang-: imo this should be able to go as low as 2 gwei.

c-johnson: Roger.

c-johnson:
The task recommended to change it to eth, but I agree. Added your revisions.

src/components/StatelessTransaction.js
```diff
@@ -84,6 +113,12 @@ const StatelessTransaction = props => {
       </InnerLabel>
     </Input>
 
+  const maxTxnPriceInEth = (
+    <div className="mt-8">
+      Max transaction price: <span className="text-700">{(gasPrice * gasLimit) / 1000000000} eth</span>
```
Fang-: "Transaction *fee*" is canonical terminology here. I recommend changing that in the variable name also, just to stay consistent.

Fang-: I'd also give this a slightly smaller top margin, `8` is a bit much. `2` might look best to keep it nice and tied to the slider.

c-johnson: Moved max transaction fee to the top right of the slider; what do you think?

PR #76, opened 2019-03-12
Use latest Trezor Connect

Fang-:
This makes it compatible with latest Trezor firmware.

Also took the liberty of re-enabling Trezor support in offline mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "react-scripts": "^2.1.1",
     "secp256k1": "^3.5.2",
     "sigil-js": "github:urbit/sigil-js#f7f29c5",
-    "trezor-connect": "^6.0.3",
+    "trezor-connect": "^7.0.1",
     "urbit-key-generation": "0.12.1",
     "urbit-ob": "^3.1.0",
     "web3": "1.0.0-beta.33"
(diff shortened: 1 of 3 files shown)


jtobin:
Looks good -- only comment is that IIRC we removed the Trezor auth from offline mode for a specific reason (initially we had enabled it, because yes, you can still use the thing while in offline mode).  But I actually can't find anything in issues/PRs for why we would have removed it exactly.

@iamwillkim do you remember what was up with that?

alexmatz:
hey @iamwillkim, any idea why you guys removed Trezor support in offline mode?

iamwillkim:
@fang I think we decided to do this because trezor required a server-side connection for auth. If the new trezor connection works offline though, keep it in!

Fang-:
It doesn't work in "my machine has no internet" offline mode, but it works fine in "I want to do weird things/testing so I selected" offline mode.

If the user tries to use Trezor in "my machine has no internet" mode, then it will fail at the very next step, and be obvious about the fact that it's a Trezor-side limitation. (To be honest, using a hardware wallet with an airgapped machine feels a bit too paranoid, but hey.)

alexmatz:
I don't have strong feelings about this, but @Fang-,  is seems like an edge case that could cause less than expert level users frustration?

Fang-:
It's an edge case I don't imagine less than experienced users ending up in:
1. Having a hardware wallet, and
2. User an airgapped (or otherwise disconnected) machine.

PR #80, opened 2019-03-20
Add view a point button to landing page

### LINKED ISSUE #54: Add 'View a Point' button to landing page
I.e., to look up a point's public information (keys/metadata) without needing to authenticate.

This is actually sort of screwy because it sidesteps network selection.  I suppose we can simply default to mainnet for this option, but it still feels a little weird.

pkova:
Fixes #54 

Network initialization is now done in Bridge.js, defaulting to mainnet unless NODE_ENV is set to development, in which case it's local.

I also decided to split out the possible point actions to a separate functional component and not render them at all if the user is not authenticated, ie has no wallet.

### DIFF
--- a/src/views/Landing.js
+++ b/src/views/Landing.js
@@ -34,6 +34,20 @@ const Landing = (props) =>
         </P>
 
         {
+          <Row>
+            <Col>
+              <Button
+                prop-type={'link'}
+                prop-size={'md'}
+                onClick={ () => props.pushRoute(ROUTE_NAMES.VIEW_SHIP) }
+              >
+                { 'View a point  →' }
+              </Button>
+              <P>
+                { "View a point on Azimuth." }
+              </P>
+            </Col>
+          </Row>
         //  FIXME add address view
         //  <LinkButton
         //    size={'l'}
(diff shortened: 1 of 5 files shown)


src/views/Landing.js
```diff
@@ -34,6 +34,20 @@ const Landing = (props) =>
...
 
         {
+          <Row>
+            <Col>
+              <Button
+                prop-type={'link'}
+                prop-size={'md'}
+                onClick={ () => props.pushRoute(ROUTE_NAMES.VIEW_SHIP) }
+              >
+                { 'View a point  →' }
```
Fang-: This should probably use the same size/styling as the "Unlock a Wallet" above? Looks kinda off right now.

src/views/Landing.js
```diff
@@ -34,6 +34,20 @@ const Landing = (props) =>
...
+              </Button>
+              <P>
+                { "View a point on Azimuth." }
```
Fang-:
```suggestion
                { "View an Azimuth point without signing into a wallet." }
```

src/views/Point.js
```diff
@@ -113,93 +98,11 @@ class Point extends React.Component {
 
     const online = Maybe.Just.hasInstance(web3)
 
-    const canSetTransferProxy = pointDetails.matchWith({
-      // also should check is addr is operator
-      Nothing: _ => false,
-      Just: details => eqAddr(details.value.owner, addr)
+    const authenticated = wallet.matchWith({
```
jtobin: Alternatively you can use `Maybe.Just.hasInstance(wallet)` here.

PR #82, opened 2019-03-21
Fix missing key props on sigil ReactSVGComponents

pkova:
Removes annoying console errors.

### DIFF
--- a/src/bridge/components/ReactSVGComponents.js
+++ b/src/bridge/components/ReactSVGComponents.js
@@ -2,59 +2,59 @@ import React from 'react'
 import { get, map } from 'lodash'
 
 const ReactSVGComponents = {
-  svg: p => {
+  svg: (p, key) => {
     return (
-      <svg {...p.attr} version={'1.1'} xmlns={'http://www.w3.org/2000/svg'}>
-       { map(get(p, 'children', []), child => ReactSVGComponents[child.tag](child)) }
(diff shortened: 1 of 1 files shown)


pkova:
After looking into this a bit more it looks like using indexes as key props may not be a good idea. I can't figure out a clean way to generate unique keys for the svg elements so I'm gonna close this one for now.

pkova:
I'm gonna actually reopen this.

Now that I've done a little bit more development on bridge I'm starting to be really annoyed by the console getting flooded by key prop errors every time a point is rendered. This PR would reduce console clutter a lot, even though it doesn't help React render anything more efficiently.

Fang-:
Lots of changes since this was first opened. You might want to rebase your changes onto current master to try and resolve the conflict.

pkova:
Man! Git, how does it even work! I guess the conflict had something to do with the file being renamed, cause there are no changes since last time.

Anyways this branch is rebased now.

gavinpatkinson:
```
To help you decide, I put together three conditions which these examples have in common:

the list and items are static–they are not computed and do not change;
the items in the list have no ids;
the list is never reordered or filtered.
When all of them are met, you may safely use the index as a key.
```

https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

PR #83, opened 2019-03-21
Fix spawn target address field disabling

Fang-:
This variable had been renamed and redefined during some refactoring, but this instance of it stuck behind. Now it actually properly enables the field if the child is available, allowing users to actually spawn things again.


Fang-:
Merging pre-review with permission from @alexmatz. Reviews still welcome though! (Even though this is a tiny change.)

PR #86, opened 2019-03-23
Implement back navigation

c-johnson:
Okay, I feel really stupid about this.

It turns out the history API works pretty well if you just add a new history entry every time the back button is pressed. This "clears" the forward history, so the user isn't prompted to navigate forward (which is where the problems happen).

So there's no forward navigation, but this seems to work comfortably without any major router overhauls.

### DIFF
--- a/src/views/Network.js
+++ b/src/views/Network.js
@@ -45,7 +45,7 @@ class Network extends React.Component {
     if (network === NETWORK_NAMES.LOCAL) {
       setNetworkType(network)
 
-      const endpoint = 'wss://localhost:8545'
+      const endpoint = 'ws://localhost:8545'
       const provider = new Web3.providers.WebsocketProvider(endpoint)
       const web3 = new Web3(provider)
       const contracts = azimuth.initContracts(web3, CONTRACT_ADDRESSES.DEV)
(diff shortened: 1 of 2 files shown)


jtobin:
A note: I recall messing around with this but didn't manage to get it to work "just right."  I observed that one could get into a state where the browser history was out of sync with the router history -- maybe in situations where the router does a `popRoute` immediately followed by a `pushRoute` or something similar.  Might be worth double checking that this works as expected everywhere in the application.

(I think I was trying to support forward history as well, which is definitely not necessary, so that might have messed me up.)

Otherwise -- le woop!  I like this simple router design, so it'd be nice to augment it with actual navigation. 😄

src/views/Network.js
```diff
@@ -45,7 +45,7 @@ class Network extends React.Component {
     if (network === NETWORK_NAMES.LOCAL) {
       setNetworkType(network)
 
-      const endpoint = 'wss://localhost:8545'
+      const endpoint = 'ws://localhost:8545'
```
Fang-: I've seen weird behavior around this (both wss and ws). Maybe I'm just doing something wrong in my dev environment. Let's talk about that today. (But this change specifically is fine, probably.)

msutherl:
🙌

PR #90, opened 2019-04-01
Switch between Walletgen and Bridge

c-johnson:
There's a button now in the bottom-right corner of the application that lets you switch between Bridge and Walletgen. Yay!

One note on styles:

Both bridge and walletgen have different root/element styles, and even use conflicting versions of base.css. For these to properly work together, we'd basically need to rewrite the styles of both applications to eliminate the conflicts. 

For now, our workaround is to just disable the stylesheet of the non-loaded application. This is a bit non-trivial; since webpack's css-loader doesn't allow you to (say) add an identifier (eg a class) to the outputted <style> element, our workaround is to inspect the stylehseet's first CSS rule, looking for a special "bridge" or "walletgen" directive, and only enable ones that match the currently loaded app.

This isn't too onerous to do, but should be communicated to developers working on this. I put some documentation in Root.js, but will probably add it to the brief dev guide as well.


alexmatz:
@jyng is this style conflict something indigo can solve?

c-johnson:
@alexmatz I think it'd be a good project to go through and update walletgen / bridge UIs once Indigo is ready. Probably best to tackle CSS consolidation at that point.

PR #95, opened 2019-04-03
Fix most ESLint warnings.

jtobin:
There are still a couple remaining, though fixing them is marginally more involved than gutting dead code (which is all I've done here).


c-johnson:
Ah, sorry about these. I'll look at the other ones. LGTM

Issue #93, opened 2019-04-03
Show live transaction state after submission


Fang-:
The transaction submission page/view should also show the user the state of their transaction: pending, confirmed or failed. This way, they can see how it ends up without checking outside tooling like Etherscan.

PR #96, opened 2019-04-04
Move common transaction state/mutations into statelessTransaction 

c-johnson:
Cuts down on the state mutation boilerplate for executing transactions by moving all the common transaction details into statelessTransaction.js


pkova:
I think you forgot (to commit) CreateGalaxy.js! 

Apologies if this is a WIP PR, noticed this since I'm doing some work on #93 but realized I better wait for this one to be merged.

src/bridge/views/AcceptTransfer.js
```diff
@@ -32,144 +32,17 @@ class AcceptTransfer extends React.Component {
...
-            gasPrice: fromWei(r[2], 'gwei'),
-          };
-
-          this.setState({...txMetadata})
-
-        })
-      }
-    });
+    this.createUnsignedTxn = this.createUnsignedTxn.bind(this)
+    this.statelessRef = React.createRef();
```
Fang-: The React docs say you need to add a `ref` attribute to some DOM element for this to work? I know this ends up with having a pointer to the `StatelessTransaction` in `statelessRef.current`, but how does that hook up? I don't see it happen anywhere.

c-johnson: Ah, good catch! I missed this one. It's there now.

c-johnson:
Indeed I did @pkova! It's updated now.

PR #98, opened 2019-04-08
[draft] invite onboarding

Fang-:
Draft PR to make progress more visible and reviews easier.

Right now, I'm mainly looking to get the last [two](https://github.com/urbit/bridge/commit/a08dec934a393b14d69144c67c42f0a8bb4136ac) [commits](https://github.com/urbit/bridge/commit/a2ce09c7a6f8be6f556e23516db3fe91104b9637) some visibility, since they're refactoring with broader impact than just the invite flow.

Invite flow pages are still WIP, but feedback is always welcome.


src/bridge/views/InviteWallet.js
```diff
@@ -0,0 +1,289 @@
...
+      return new Promise((resolve, reject) => {
+        gen.generate(bits, result => {
+          const chunked = lodash.chunk(result, 2)
+          const desired = chunked.slice(0, bytes) // only take required entropy
+          const more = lodash.flatMap(desired, arr => arr[0] ^ arr[1])
+          const entropy = lodash.zipWith(some, more, (x, y) => x ^ y)
+          const buf = Buffer.from(entropy)
+          const patq = ob.hex2patq(buf.toString('hex'))
+          resolve(patq)
+          reject('Entropy generation failed')
```
c-johnson: Does this `reject` ever get run? Maybe there's a condition where you want to reject it.

Fang-: I pulled this from walletgen's `Generate` view. Not sure why it was written like this, and I don't think that as-is it can ever `reject`? idk where this would ever fail, anyway.

I'll probably clean that out once I move this into a lib.

Fang-:
This thing in its entirety is never ever gonna get a proper review. People have looked at it over the course of its development, and much more work will be done over top of this still.

Merging into the `mino` branch now so we have a clear, cleanly-named dev branch to work with.

PR #99, opened 2019-04-08
Add transaction status indicator to SentTransaction

### LINKED ISSUE #93: Show live transaction state after submission

pkova:
Fixes #93

### DIFF
--- a/src/bridge/Bridge.js
+++ b/src/bridge/Bridge.js
@@ -80,7 +80,8 @@ class Bridge extends React.Component {
       pointCursor: Maybe.Nothing(),
       pointCache: {},
       // txn
-      txnHashCursor: Maybe.Nothing()
+      txnHashCursor: Maybe.Nothing(),
+      txnConfirmations: {}
     }
 
     this.pushRoute = this.pushRoute.bind(this)
@@ -95,6 +96,7 @@ class Bridge extends React.Component {
     this.setAuthMnemonic = this.setAuthMnemonic.bind(this)
     this.setPointCursor = this.setPointCursor.bind(this)
     this.setTxnHashCursor = this.setTxnHashCursor.bind(this)
+    this.setTxnConfirmations = this.setTxnConfirmations.bind(this)
     this.setNetworkSeedCache = this.setNetworkSeedCache.bind(this)
     this.addToPointCache = this.addToPointCache.bind(this)
   }
@@ -234,6 +236,12 @@ class Bridge extends React.Component {
     this.setState({ txnHashCursor })
   }
 
+  setTxnConfirmations(txnHash, txnConfirmations) {
+    this.setState((prevState, _) =>
+                  ({txnConfirmations: { ...prevState.txnConfirmations,
+                                        [txnHash]: txnConfirmations}}))
+  }
+
   render() {
     const {
       routeCrumbs,
@@ -248,6 +256,7 @@ class Bridge extends React.Component {
       pointCursor,
       pointCache,
       txnHashCursor,
+      txnConfirmations,
       web3,
       contracts
     } = this.state
@@ -300,7 +309,9 @@ class Bridge extends React.Component {
                 setNetworkSeedCache= { this.setNetworkSeedCache }
                 // txn
                 setTxnHashCursor={ this.setTxnHashCursor }
-                txnHashCursor={ txnHashCursor } />
+                txnHashCursor={ txnHashCursor }
+                setTxnConfirmations={ this.setTxnConfirmations }
+                txnConfirmations={ txnConfirmations } />
 
               <div className={'push'} />
 
--- a/src/bridge/components/StatelessTransaction.js
+++ b/src/bridge/components/StatelessTransaction.js
@@ -85,7 +85,7 @@ class StatelessTransaction extends React.Component {
 
   submit(){
     const { props, state } = this
-    sendSignedTransaction(props.web3.value, state.stx)
+    sendSignedTransaction(props.web3.value, state.stx, props.setTxnConfirmations)
       .then(sent => {
         props.setTxnHashCursor(sent)
         props.popRoute()
--- a/src/bridge/lib/txn.js
+++ b/src/bridge/lib/txn.js
@@ -137,7 +137,7 @@ const signTransaction = async config => {
   setStx(Just(stx))
 }
 
-const sendSignedTransaction = (web3, stx) => {
+const sendSignedTransaction = (web3, stx, confirmationCb) => {
   const txn = stx.matchWith({
     Just: (tx) => tx.value,
     Nothing: () => {
@@ -153,6 +153,10 @@ const sendSignedTransaction = (web3, stx) => {
         resolve(Just(Ok(hash)))
       )
       .on('receipt', txn => {
+        resolve(Just(Ok(txn.transactionHash)))
+      })
+      .on('confirmation', (confirmationNum, txn) => {
+        confirmationCb(txn.transactionHash, confirmationNum + 1)
         resolve(Just(Ok(txn.transactionHash)))
       })
       .on('error', err => {
--- a/src/bridge/views/SentTransaction.js
+++ b/src/bridge/views/SentTransaction.js
@@ -92,6 +92,8 @@ class Success extends React.Component {
           <H3>{ 'Transaction Status' }</H3>
           <P>
             { status }
+          </P>
+          <P>
             { esanchor }
           </P>
           <P>
--- a/src/bridge/views/SentTransaction.js
+++ b/src/bridge/views/SentTransaction.js
@@ -30,7 +30,7 @@ class Success extends React.Component {
     this.setState({interval: interval})
   }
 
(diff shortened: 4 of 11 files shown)


src/bridge/lib/txn.js
```diff
@@ -155,6 +155,10 @@ const sendSignedTransaction = (web3, stx) => {
       .on('receipt', txn => {
         resolve(Just(Ok(txn.transactionHash)))
       })
+      .on('confirmation', (confirmationNum, txn) => {
+        confirmationCb(txn.transactionHash, confirmationNum + 1)
```
pkova: Feels weird to have a callback like this inside a promise, but I think this allows us to implement the feature with minimal code changes to sendSignedTransaction.

src/bridge/Bridge.js
```diff
@@ -300,7 +309,9 @@ class Bridge extends React.Component {
                 setNetworkSeedCache= { this.setNetworkSeedCache }
                 // txn
                 setTxnHashCursor={ this.setTxnHashCursor }
-                txnHashCursor={ txnHashCursor } />
+                txnHashCursor={ txnHashCursor }
+                setTxnConfirmations={ this.setTxnConfirmations }
+                txnConfirmations={ txnConfirmations } />
```
Fang-: @c-johnson I continue to feel highly uncomfortable with the dumping of state into `Bridge.js` for the sole purpose of passing it into new/different screens. In a lot of these cases, especially ones like these where only a single, temporary screen needs the data, it seems much more desirable to pass it in via the `pushRoute` data argument.

How would you feel about standardizing on a pattern for that? Would it be possible to do that in such a way that data is always explicitly passed forward, instead of stored in pseudo-global state? I'm trying to steer towards a slightly more functional style here. @jtobin may also have thoughts.

src/bridge/components/StatelessTransaction.js
```diff
@@ -85,7 +85,7 @@ class StatelessTransaction extends React.Component {
 
   submit(){
     const { props, state } = this
-    sendSignedTransaction(props.web3.value, state.stx)
+    sendSignedTransaction(props.web3.value, state.stx, props.setTxnConfirmations)
```
jtobin: I just noticed that `web3` isn't being unwrapped here.  I.e., `web3` could be `Nothing`, such that `web3.value` would be `undefined`.  It should be unwrapped in the standard fashion; something like:

```
const web3 = props.web3.matchWith({
  Just: w3 => w3.value,
  Nothing: _ => { throw BRIDGE_ERROR.MISSING_WEB3 }
})

sendSignedTransaction(web3, ...)
```

That folktale lets one just wing it on the `value` property is pretty lousy, but alas, such is the nature of JS.

(This isn't actually related to your change @pkova -- I just happened to spot it while scrolling through the diff.)

src/bridge/views/SentTransaction.js
```diff
@@ -29,10 +33,16 @@ const Success = (props) => {
...
-        href={`https://${esdomain}/tx/${props.hash}`}>
+        href={`https://${esdomain}/tx/${hash}`}>
           {'View on Etherscan ↗'}
       </Anchor>
 
+  const confirmations = Maybe.fromNullable(txnConfirmations[hash]).getOrElse(0)
+
+  const requiredConfirmations = 1
+
+  const status = confirmations < requiredConfirmations ? 'Pending...' : `Confirmed x${confirmations}!`
```
Fang-: Would love to see a spinner displayed alongside the "Pending..." text, to indicate to the user that this is actively happening/being checked for.

PR looks good at first glance, I'll give this a proper test run & review tomorrow! (:

src/bridge/components/StatelessTransaction.js
Fang-: Is it not ok to use `.value` directly if pre-condition logic (ie whatever you need to do to make `submit()` accessible) already guarantees/checks for `Just`? (Of course, figuring that out as the reader isn't always trivial, but "trust the author"? (^: )

jtobin: The reason you want to be pedantic and do it literally everywhere is that it's easy to change the code (perhaps in a refactoring), remove the precondition, but forget to change the yolo `.value` attempt.  It's a recipe for introducing annoying bugs eventually.

Better would probably be to do the unwrapping once, in one place, consistently.  This is what I was thinking when I mentioned unwrapping all the necessary props in the constructor, or at least all in one place, and then to use the unwrapped values elsewhere in the component.  Maybe something like:

```
class Foo extends React.Component {
  constructor(props) {
    super(props)

    this.web3 = props.web3.matchWith({
      Nothing: _ => throw { BRIDGE_ERROR.MISSING_WEB3 },
      Just: w3 => w3.value
    })
  }

  render() {
    const web3 = this.web3

    return(
      <div>{ web3 }</div>
    )
  }
}
```

Fang-: Definitely in favor of unwrapping once in the constructor, yes. I'll put that on my informal refactoring list.

src/bridge/Bridge.js
jtobin: In general I think that's a great idea.  I would just try to ensure `routeData` maintains some sort of sane/predictable structure -- I think using a dynamically-typed blob consisting of whatever the downstream component needs doesn't necessarily scale well, for example.

src/bridge/components/StatelessTransaction.js
pkova: I don't like Maybe. Using it has the upside of 'forcing you to check' for missing values, but the tradeoff is that changes that should be non-breaking become breaking changes. If we have a function 

`foo :: x -> Maybe y` 

and we manage to improve it so that we can always return y, eg. 

`foo :: x -> y` 

existing callers of our function break. Nullable types a la Kotlin or Typescript seem better. [Rich explains this far better than I ever could.](https://github.com/matthiasn/talk-transcripts/blob/master/Hickey_Rich/MaybeNot.md)

Don't think there's a solution for this in js-land, though!

src/bridge/Bridge.js
```diff
@@ -234,6 +236,12 @@ class Bridge extends React.Component {
     this.setState({ txnHashCursor })
   }
 
+  setTxnConfirmations(txnHash, txnConfirmations) {
+    this.setState((prevState, _) =>
+                  ({txnConfirmations: { ...prevState.txnConfirmations,
+                                        [txnHash]: txnConfirmations}}))
+  }
+
```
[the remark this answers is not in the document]
pkova: `StatelessTransaction` needs to communicate state change (confirmations going up) to `SentTransaction`, which is not possible through `pushRoute`, since it only ever gets called once in the `submit` method. Only the `render` method on a component gets called on state updates.

src/bridge/Bridge.js
c-johnson: A couple things I can think of; 

- We could bundle the routeData with the actual routeCrumb entry, so it's at least coupled to its view and not a free-floating variable:
in `Bridge.js#pushRoute`

      this.setState((state, _) => ({
        routeCrumbs: state.routeCrumbs.push({
          view: symbol,
          data: routeData
        })
      }));

- Adding propType validators on views that take custom route data to shore up the structure somewhat:

in `Bridge.js#pushRoute`
SentTransaction.propTypes = {
  routeData:  PropTypes.shape({
    promptKeyfile: PropTypes.bool
  })
}

I can't think of many other ways to isolate route data from global state, since all `pushRoute` has access to is updating `Bridge.js`'s state variable. But definitely open for other suggestions if you think of anything.

src/bridge/components/StatelessTransaction.js
jtobin: I'm a Haskeller, so Rich's talk triggers the @#&! out of me.  But what I can certainly agree on is that, yes, there's no satisfactory solution for this in JS-land. ;-)

pkova:
Animations are a go.

src/bridge/views/SentTransaction.js
```diff
@@ -91,8 +91,8 @@ class Success extends React.Component {
 
           <H3>{ 'Transaction Status' }</H3>
           <P>
-            <div>{ status }</div>
-            <div>{ esanchor }</div>
+            { status }
+            { esanchor }
```
Fang-: I think you just want to give these their own separate `<P>` tags.
image

pkova: Oh shit! Fixed.

pkova: Actually, one more fix needed.

Fang-: This looks good to me rn, but let me know when you're truly done.

pkova: Allright we're good now, made 3 small commits which should explain themselves.

Fang-:
Looks clean! Thanks a bunch @pkova, this is good!

Issue #89, opened 2019-03-25
Accept incoming transfer unavailable in offline mode


Fang-:
It's greyed out in the point overview screen, but shouldn't I still be able to generate a transaction for that in offline mode, if I so choose?

<img width="411" alt="image" src="https://user-images.githubusercontent.com/3829764/54955261-141c8400-4f4d-11e9-9151-f29206bc4e62.png">

jtobin:
Definitely.  I think it's just greyed out because we currently hit the network to check if the currently-auth'd address is a transfer proxy for the point, but that hardly seems necessary.  The `disabled` condition in the point view can probably just be changed to `online && !canAcceptTransfer`.

c-johnson:
has this been done already? looks like disabled only checks canAcceptTransfer currently: https://github.com/urbit/bridge/blob/master/src/bridge/views/Point/Actions.js#L147

alexmatz:
@Fang- do you know if this has been done already?

Fang-:
<img width="398" alt="image" src="https://user-images.githubusercontent.com/3829764/55667781-3ffb0c00-5861-11e9-8440-0622577f05eb.png">

Oh no, it's getting worse. (^:

@c-johnson, that check ends up as `false` [here](https://github.com/urbit/bridge/blob/master/src/bridge/views/Point/Actions.js#L95). We'll want all of these checks to return `true` in the case of offline mode, regardless of values. [cancelTransfer](https://github.com/urbit/bridge/blob/master/src/bridge/views/Point/Actions.js#L152) has the right idea.

Also we'll need to think about how to do keyfile generation in offline mode. Probably just an input field for the revision number if logged in with an urbit wallet, falling back to the "pls reset" in all other cases?

Fang-:
Bridge v2 has no offline mode. Closing.

Issue #53, opened 2019-01-11
Checkboxes can't be un/checked by clicking their labels


Fang-:
The text next to them should be in `<label>`s and properly associated, the browser will take care of the rest.

Issue #52, opened 2019-01-13
Putting in an address with invalid checksum crashes the app


Fang-:
You can reproduce this by copy-pasting your logged-in-as address into an address field, and then changing one of the letters to/from capital.

<img width="1028" alt="image" src="https://user-images.githubusercontent.com/3829764/51089675-1df06280-1726-11e9-9350-7508bd5d0490.png">

c-johnson:
Yeah, azimuth.ecliptic throws exceptions that crash the app when in development mode. I think this is by design?

I've lowercased the address input before we submit it to ecliptic; that should handle this particular error, I think other conceivable errors should be handled in the validator function. 

We could try wrapping these in a try/catch to prevent app crashing; not sure if @jtobin has a better idea for how we should handle these exceptions.

PR #104, opened 2019-04-09
Ui improvements

c-johnson:
Fixes a couple of small UI issues, including https://github.com/urbit/bridge/issues/89, https://github.com/urbit/bridge/issues/53, and https://github.com/urbit/bridge/issues/52. 

Still need to do more work on https://github.com/urbit/bridge/issues/89 on making sure the actions actually work, but I'll do that in another branch.

### DIFF
--- a/src/bridge/components/StatelessTransaction.js
+++ b/src/bridge/components/StatelessTransaction.js
@@ -1,244 +1,426 @@
-import { Nothing } from 'folktale/maybe'
+import { Nothing, Just } from 'folktale/maybe'
 import React from 'react'
 
 import { Code, H3 } from './Base'
 import { Button } from './Base'
 import { CheckboxButton, Input, InnerLabel } from './Base'
+import {  Warning } from '../components/Base'
 
-import { renderSignedTx, signTransaction } from '../lib/txn'
+import { addressFromSecp256k1Public } from '../lib/wallet'
+import { BRIDGE_ERROR } from '../lib/error'
+import { ROUTE_NAMES } from '../lib/router'
 
-const StatelessTransaction = props => {
-  const { web3, gasPrice, gasLimit, nonce, chainId } = props
-  const { setNonce, setChainId, setGasLimit, setGasPrice, showGasDetails, toggleGasDetails } = props
-  const { txn, stx, createUnsignedTxn } = props
-  const { canSign, canGenerate, canApprove } = props
-  const { setUserApproval, userApproval } = props
-  const { canSend, handleSubmit } = props
+import {
+  sendSignedTransaction,
+  fromWei,
+  renderSignedTx,
+  signTransaction
+ } from '../lib/txn'
+
+class StatelessTransaction extends React.Component {
+
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      gasPrice: '5',
+      gasLimit: '600000',
+      showGasDetails: false,
+      userApproval: false,
+      chainId: '',
+      nonce: '',
+      stx: Nothing(),
+      txn: Nothing(),
+      txError: Nothing(),
+    }
+
+    this.createUnsignedTxn = this.createUnsignedTxn.bind(this)
+    this.submit = this.submit.bind(this)
+    this.setUserApproval = this.setUserApproval.bind(this)
+    this.setTxn = this.setTxn.bind(this)
+    this.setStx = this.setStx.bind(this)
+    this.sendTxn = this.sendTxn.bind(this)
+    this.setNonce = this.setNonce.bind(this)
+    this.setChainId = this.setChainId.bind(this)
+    this.setGasPrice = this.setGasPrice.bind(this)
+    this.setGasLimit = this.setGasLimit.bind(this)
+    this.rangeChange = this.rangeChange.bind(this)
+    this.toggleGasDetails = this.toggleGasDetails.bind(this)
+  }
+
+  componentDidMount() {
+    const { props } = this
+
+    const addr = props.wallet.matchWith({
+      Just: wal => addressFromSecp256k1Public(wal.value.publicKey),
+      Nothing: () => {
+        throw BRIDGE_ERROR.MISSING_WALLET
+      }
+    })
+
+    props.web3.matchWith({
+      Nothing: () => {},
+      Just: (w3) => {
+        const validWeb3 = w3.value
+
+        const getTxMetadata = [
+          validWeb3.eth.getTransactionCount(addr),
+          validWeb3.eth.net.getId(),
+          validWeb3.eth.getGasPrice()
+        ];
+
+        Promise.all(getTxMetadata).then(r => {
+          const txMetadata = {
+            nonce: r[0],
+            chainId: r[1],
+            gasPrice: fromWei(r[2], 'gwei'),
+          };
+
+          this.setState({...txMetadata})
+
+        })
+      }
+    });
+  }
+
+  submit(){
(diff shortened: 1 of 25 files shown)


src/bridge/views/AcceptTransfer.js
```diff
@@ -65,7 +65,7 @@ class AcceptTransfer extends React.Component {
     return Just(azimuth.ecliptic.transferFrom(
       validContracts,
       owner,
-      state.receivingAddress,
+      state.receivingAddress.toLowerCase(),
```
Fang-: idk if this is the right approach. If you enter an address with invalid checksum, the app *should* shout at you... but not in a "crash and burn" kinda way. Probably want to just do checksum checking on the address input component and have it consider input as invalid appropriately?

src/bridge/components/StatelessTransaction.js
```diff
@@ -358,7 +358,7 @@ class StatelessTransaction extends React.Component {
         onClick={ setUserApproval }
         state={ userApproval }
       >
-        <div>
+        <div onClick={ setUserApproval }>
```
Fang-: Sorry to keep hitting you with the refactoring comments, but:

Shouldn't checkbox labels just be part of the `CheckboxButton` component (and implemented as `<label>` for native click-ability)?

c-johnson: Haha, that's okay. 

The problem is that native-label-clickability works when it's attached to an `<input>` with a 'for' attribute. CheckboxButton is implemented as a `<button>` with a custom way of displaying the 'checked' state (TFW no native component styling). Refactoring CheckboxButton seems like a lot of work for the functionality we want to accomplish here.

Fang-: I just looked at the code, as I should have earlier, and... oh my, raw SVG, how lovely!

Fair re: native `label`ing. I still think there's something to be said for moving labels into the components, but I'm not dead-set of having that done right now.

It's weird though that we have both a `Checkbox` and a `CheckboxButton`, and that they're completely distinct. The `Checkbox` does have built-in labels it seems like? `subtitle` or something.  
Wake me from this nightmare.

src/bridge/views/AcceptTransfer.js
c-johnson: Left a comment about this here, but maybe there's a better solution: https://github.com/urbit/bridge/issues/52#issuecomment-479678341

IMO, the validator function should handle things like address input before we even call azimuth.ecliptic. The exception we get back from azimuth.ecliptic is a big stack trace that's hard to map to a specific validation error:

image

I'm not sure what exceptional cases that azimuth.ecliptic might throw or where to display them on the page. Maybe we just add a try/catch with a noop for now to reduce development pain?

c-johnson: Regardless of what we do about exceptions, I still think lowerCasing is a correct thing to do to fix this specific problem; users should be able to input uppercase addresses and have it go through successfully (right?)

Fang-: I think we're saying the same thing, at least in your first comment!

What I mean to say, is that the component that takes an Ethereum address and updates `receivingAddress` is the one that should do checksum checking, and consider the input invalid if the input is checksummed but not correct, not setting `receivingAddress` in that case.

Fang-: > users should be able to input uppercase addresses and have it go through successfully (right?)

To answer this more directly: the point of addresses with uppercase characters in them is that they're checksummed. If that checksum doesn't check out, we should inform the user that they may have made a typo, or something went wrong during copy-pasting. Just always lowercasing their input defeats the purpose of having checksummed addresses. See also rationale for the standard [here](https://eips.ethereum.org/EIPS/eip-55).

c-johnson: Oh, okay, that makes sense! Definitely the wrong solution here then, hah.

c-johnson: So we should probably update isValidAddress in lib/wallet's validation logic then, right? https://github.com/urbit/bridge/blob/master/src/bridge/lib/wallet.js#L72

Fang-: That's probably the one, yeah. Except that checksum validation logic lives in web3, and I'm not sure where else we have it accessible, so that may not be trivial to pull into that lib... ):

c-johnson: Looks like we've already implemented it right below where it's supposed to be called, haha. https://github.com/urbit/bridge/blob/master/src/bridge/lib/wallet.js#L75

Still trying to understand exactly what we do here

1) User types a lowercase address
   - This is okay, and validation should return true (?)
2) User types a mixed case address
   - Run it through the checksum validator, and if the checksum address matches up, validation should return true
3) User types an uppercase address 
  - This is okay, and validation should return true (?)

Fang-: Oh, sorry for the delay on this.

1 and 2 are correct. 3 should be treated as identical to case 2. "Fully uppercase address" is not a common/valid notation, unless that happens to be the checksum version of the address. (Not that that's ever the case, but "string contains uppercase" is a good trigger for "need to verify checksum".)

c-johnson: That's what I landed on. Pushing a version that has this, just confirming it works real quick

```
const isValidAddress = (address) => {
  let validLength = /^0x[0-9a-fA-F]{40}$/.test(address)
  let isLowercase = address === address.toLowerCase()
  let validChecksum = address === toChecksumAddress(address)

  if (isLowercase) return validLength
  else return validChecksum && validLength
}
```

Fang-: Oh, actually, I just discovered web3 has [`web3.utils.isAddress`](https://web3js.readthedocs.io/en/1.0/web3-utils.html#isaddress), which seems to do exactly what we need.

c-johnson: We don't always have access to Web3 though, and passing it around a bunch of utility functions seems kinda meh...

Weirdly, you can't seem to use Web3.utils without initializing a provider first, so "offline" functions like this are hard to import directly >.<

jtobin: @c-johnson There's one trick to get at those bad boys that's used in `lib/txn.js'.  Agreed that it's weird, though.. :-)

``` js
const dummy = new Web3()
const toHex = dummy.utils.toHex
const toWei = dummy.utils.toWei
const fromWei = dummy.utils.fromWei
```

Fang-: Oh, oof, you're right. That sucks almost as much as reinventing the wheel.

Would it be weird to suggest a `/lib/ethUtils.js` that uses @jtobin's suggested method to expose "web3-less web3 functions"?

c-johnson: Oh, interesting. It looks like the latest beta of Web3 actually breaks this; see this comment here (https://github.com/ethereum/web3.js/issues/2573#issuecomment-476231814). 

I didn't catch this during my beta.52 testing, but looks like this does actually break existing functionality. My love for the web3 library grows every day.

Looks like recommended usage is to just install web3-utils, which works fine with beta.52. Doing that now.

c-johnson: This PR should address it; when it gets merged, I'll swap out our implementation for web3's isAddress

https://github.com/urbit/bridge/pull/112

Issue #102, opened 2019-04-09
"Error: no point found" on setKeys page in offline mode


Fang-:
In offline mode, if you go to view a point, and then try the "set keys" page, it shouts the "no point found" error at you. Other pages don't do this. In offline mode, setting keys should just work with the "randomly generated seed" logic.

Fang-:
Bridge v2 has no offline mode. Closing.

PR #109, opened 2019-04-10
Generate keyfile in offline mode

c-johnson:
This PR allows users to enter their own revision number on the genKeyfile view, allowing them to generate a keyfile that we assume is valid.

This only works if:
  - They log in as a valid management proxy for the point they're generating for 
  - They log in with an Urbit wallet 
  - They enter the correct revision number

Fixes https://github.com/urbit/bridge/issues/102, https://github.com/urbit/bridge/issues/89

### DIFF
--- a/src/bridge/lib/keys.js
+++ b/src/bridge/lib/keys.js
@@ -62,9 +62,13 @@ const genKey = (networkSeed, point, revision) => {
 
 // 'next' refers to setting the next set of keys
 // if false, we use revision - 1
-const attemptSeedDerivation = async (next, args) => {
+const attemptSeedDerivation = async (next, args, assumedRevision) => {
   const { walletType, wallet, pointCursor, pointCache } = args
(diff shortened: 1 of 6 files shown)


src/bridge/lib/keys.js
```diff
@@ -62,10 +62,14 @@ const genKey = (networkSeed, point, revision) => {
 
 // 'next' refers to setting the next set of keys
 // if false, we use revision - 1
-const attemptSeedDerivation = async (next, args) => {
+const attemptSeedDerivation = async (next, args, assumedRevision) => {
```
Fang-: The fact that this can't reasonable be part of `args` makes me think that's a misleading variable name. Just noting that for the record, no action required here.

Fang-:
Ping @c-johnson!

shrugs:
closing due to staleness, will be tracked as part of https://github.com/urbit/bridge/issues/183

PR #112, opened 2019-04-11
Fix web3

c-johnson:
Turns out, web3@beta.52 does break some things; it requires the Web3 object to pass in a provider, whereas before `new Web3()` worked without one. 

This PR includes web3-utils directly in cases where we don't want a web3 object. In offline mode, we do need a web3 object to build the contracts, so we pass in a dummy localhost address during initialization.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2008,6 +2008,11 @@
         "web3": "1.0.0-beta.33"
       },
       "dependencies": {
+        "bn.js": {
+          "version": "4.11.6",
+          "resolved": "https://registry.npmjs.org/bn.js/-/bn.js-4.11.6.tgz",
+          "integrity": "sha1-UzRK2xRhehP26N0s4okF0cC6MhU="
+        },
         "web3": {
           "version": "1.0.0-beta.33",
           "resolved": "https://registry.npmjs.org/web3/-/web3-1.0.0-beta.33.tgz",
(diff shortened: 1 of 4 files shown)


c-johnson:
@Fang- Same :(

Fang-:
This never should've been merged. (^:

As you've observed, this breaks transaction sending. "invalid opcode" errors can mean any number of things. Posting this comment to notify you and others that I'm looking into what it is we're actually doing wrong/different. Pretty sure that it isn't an issue in web3, since another codebase I've been working on uses beta.52 and does transaction signing and sending just fine. I'll update when I know more.

(Also, this should've pinned `web3-utils` just like it did `web3`.)

Fang-:
Oh, it's because we're using transaction objects that have `to: undefined`. Of course that's invalid!

Still not quite sure why the `to` address isn't being set (during azimuth-js's calls to contract methods), but that does very much feel like a web3 issue... Maybe something about the interface there changed?

Having a rough time confirming that. Trying the `contract.methods.thing().call()` pattern from node cli just straight-up sends the transaction instead of giving an object, but it does also shout invalid opcode at me...

Fang-:
Found it! At some point, `contractObject._address` got changed to `contractObject.address`. Azimuth-js should be updated to use latest web3 and account for that change. Upgrading web3 there breaks tests though, so I'll have to do a slightly deeper dive on that tomorrow.

c-johnson:
Ah, great find! I'll go ahead and revert Web3 back to beta.33 until we can fix this. Happy to help (actually) test it again when we fix azimuth-js. Sorry about that :/

Issue #111, opened 2019-04-11
In offline mode, can still built transactions without entering nonce or chain ID


Fang-:
It doesn't warn you if you leave these at their empty defaults. Simplest fix here is to just give them defaults of `0` and `mainnnet` respectively.

Additionally, when you've signed a transaction, then change the nonce, the signed output disappears (as it should), but that doesn't happen when you change the chain ID.

PR #113, opened 2019-04-12
add defaults for nonce and chainId, and clear signed transaction

c-johnson:
Fixes https://github.com/urbit/bridge/issues/111


c-johnson:
Yep!

PR #114, opened 2019-04-16
Revert web3 back to beta.33 until we fix issues

c-johnson:


Fang-:
The only thing that depended on beta.52 was #104 right? Doesn't that need any of its changes reverted now?

c-johnson:
Hm, I think it should be okay? I left web3-utils at beta.52, it doesn't seem to make much difference (I tested the utility functions to make sure)

Issue #33, opened 2019-02-04
Instructions should be distributed with the release


msutherl:
The release folder does not contain instructions for running the software. It's another thing to keep updated, but seems advisable!

msutherl:
@jtobin should we revisit this?

Fang-:
This seems completely fair. New release Soon™, I'll put this in with it.

msutherl:
@Fang- some additional detail: working with a user currently who may be running the standard instructions on the release page and the `mkcert` instructions in the README simultaneously. I'd suggest presenting both of these sets of instructions in the same place and making it clear that it's an either/or choice.

Fang-:
Resolved by putting instructions in the readme. #117 should take care of the rest.

PR #117, opened 2019-04-17
Clarify mkcert instructions and mention nss dependency

msutherl:
When I installed `mkcert` with `brew` on Mac OS, it mentioned that I would need to install `nss` in order for it to work with Firefox. I can't verify that this is true, and true on every platform, but it seems worth mentioning.

Also, today myself and (probably) a user made the mistake of glossing over that you need to go to http*s*://localhost:4443. Just listing the URI in full I think minimizes the chance of missing this, and also might clarify the instructions for non-developers who could be unfamiliar with localhost.

This is pretty nitpicky, so open to modifications or rejection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,8 @@ To authenticate and sign transactions with a Ledger, Bridge must be serving
 over HTTPS on localhost. This requires self-signed certificates. You can do
 this via the following:
 
-* Install [mkcert](https://github.com/FiloSottile/mkcert), and [nss](https://github.com/nss-dev/nss) if you're using Firefox
+* Install [mkcert](https://github.com/FiloSottile/mkcert)
+* If you're using Firefox, additionally install [nss](https://github.com/nss-dev/nss)
 * Install a local certificate authority via `mkcert -install`
 * In your `bridge` directory, generate a certificate valid for localhost via
   `mkcert localhost`.  This will produce two files: `localhost.pem`, the local
--- a/README.md
+++ b/README.md
@@ -10,22 +10,24 @@ Python 3.7.2
 
 ### Instructions
 
-Before using Bridge, you'll need access to whatever media you're using to store the Ethereum wallet that owns or will own your Azimuth points. This could be an Urbit HD wallet containing a Master Ticket, a BIP39 mnemonic, a Ledger or Trezor hardware wallet, or an Ethereum private key or keystore file.
+To use Bridge:
 
-With that in hand, to use Bridge, [download a release](https://github.com/urbit/bridge/releases) and unzip it (`bridge-$version.zip`), then do one of the following:
+- [Download a release](https://github.com/urbit/bridge/releases)
+- Unzip it (`bridge-$version.zip`)
+- Open up your command line interface (Terminal on MacOS, Command Prompt on Windows)
+- Follow the instructions below to run Bridge
 
-#### Non-Ledger
+#### Running Bridge
 
 If you plan to authenticate and sign transactions with a Master Ticket, BIP39 mnemonic, Ethereum private key, or keystore file:
 
-1. Open up your command line interface (Terminal on MacOS, Command Prompt on Windows)
-2. `cd` into the `bridge-$version` directory
-3. Run this command: `python3 -m http.server 5000 --bind 127.0.0.1`
-4. Navigate to http://localhost:5000 using a web browser to access Bridge
+1. `cd` into the `bridge-$version` directory
+2. Run this command: `python3 -m http.server 5000 --bind 127.0.0.1`
+3. Navigate to http://localhost:5000 using a web browser to access Bridge
 
-#### Ledger support
+#### Running Bridge with Ledger support
 
-To authenticate and sign transactions with a [Ledger](https://www.ledger.com/), Bridge must be serving over HTTPS on localhost. This requires self-signed certificates. To do this, follow these steps:
+If you plan to authenticate and sign transactions with a [Ledger](https://www.ledger.com/), Bridge must be serving over HTTPS on localhost. This requires self-signed certificates. To do this:
 
 1. Install [mkcert](https://github.com/FiloSottile/mkcert)
 2. If you're using Firefox, additionally install [nss](https://github.com/nss-dev/nss)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,14 +38,13 @@ To authenticate and sign transactions with a Ledger, Bridge must be serving
 over HTTPS on localhost. This requires self-signed certificates. You can do
 this via the following:
 
-* Install [mkcert](https://github.com/FiloSottile/mkcert)
+* Install [mkcert](https://github.com/FiloSottile/mkcert), and [nss](https://github.com/nss-dev/nss) if you're using Firefox
```
Fang-:
```suggestion
* Install [mkcert](https://github.com/FiloSottile/mkcert)
* If you're using Firefox, additionally install [nss](https://github.com/nss-dev/nss)
```

msutherl:
@Fang- ;)

Also moved the Ledger instructions out of the Development section – let me know what you think. Sorry for the merge commit.

README.md
```diff
@@ -13,10 +13,23 @@ Python 3.7.2
 1. To use Bridge, [download a release](https://github.com/urbit/bridge/releases) and unzip it (`bridge-$version.zip`).
 2. Open up your command line interface (Terminal on MacOS, Command Prompt on Windows).
 3. `cd` into the `bridge-$version` directory.
-4. Run this command: `python3 -m http.server 5000 --bind 127.0.0.1`
+4. Run this command: `python3 -m http.server 5000 --bind 127.0.0.1`. If you're using Bridge with a Ledger hardware wallet, skip this step and see the Ledger Support subsection below.
 5. You can then use Bridge by navigating to http://localhost:5000 using a web browser.
 
-### Verify checksums
+#### Ledger Support
```
Fang-: Since this is a more rare usage case than "verify checksums", I'd put it below that.

Also, this needs to be a h3, not h4, to match up with the other headers here.

README.md
```diff
@@ -31,7 +44,7 @@ To validate your downloaded file's integrity, compare the lines in checksum.txt
 Clone the repo, and use a simple `npm install`.  You can then use a `npm run
 build` to create an optimised static build (serve it with e.g. [serve][serv]).
 
-### General notes
+### General Notes
```
Fang-: *grumbles about capitalists*

imo capitalizing Every Single Word in a header is redundant with the fact that it's a header. They're not proper names, or titles, and we aren't writing German, so headers should only have a capital at the start.  
Same for the other two headers you changed.

If you care a lot, I'll give you this. It's not my hill to die on. But if you're lukewarm on the issue I'd prefer reverting to less capitals.

msutherl: I almost added "fix Euro mixed caps O_0" to the commit message. This is an *American* company. Sorry.

In seriousness, I see we don't use all caps in other READMEs, so I'll accept the revert for consistency. I may die on this hill, however.

README.md
msutherl: Disagree with this change. I had explicitly made it a subsection – running the software for Ledger is a special case of running the software. All instructions for running the software should be grouped together.

I debated whether verify checksums should actually come first. I'm open to that, but I also suspect that verifying checksums is something only people who already know how to do it, or are looking for it, are going to do.

Fang-: Oh, hmm, I suppose that's fair then.

If you can come up with nice copy for "do these steps instead of the above" for this Ledger section, by the way, then go for it. (As per your comment on #33.)

msutherl: Done – have a look

README.md
```diff
@@ -10,11 +10,31 @@ Python 3.7.2
...
-1. To use Bridge, [download a release](https://github.com/urbit/bridge/releases) and unzip it (`bridge-$version.zip`).
-2. Open up your command line interface (Terminal on MacOS, Command Prompt on Windows).
-3. `cd` into the `bridge-$version` directory.
-4. Run this command: `python3 -m http.server 5000 --bind 127.0.0.1`
-5. You can then use Bridge by navigating to http://localhost:5000 using a web browser.
+Before using Bridge, you'll need access to whatever media you're using to store the Ethereum wallet that owns or will own your Azimuth points. This could be an Urbit HD wallet containing a Master Ticket, a BIP39 mnemonic, a Ledger or Trezor hardware wallet, or an Ethereum private key or keystore file.
+
+With that in hand, to use Bridge, [download a release](https://github.com/urbit/bridge/releases) and unzip it (`bridge-$version.zip`), then do one of the following:
+
+#### Non-Ledger
```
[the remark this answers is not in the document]
msutherl: That's a nice suggestion – I'll go with something like that

README.md
```diff
@@ -10,11 +10,31 @@ Python 3.7.2
 
 ### Instructions
 
-1. To use Bridge, [download a release](https://github.com/urbit/bridge/releases) and unzip it (`bridge-$version.zip`).
-2. Open up your command line interface (Terminal on MacOS, Command Prompt on Windows).
-3. `cd` into the `bridge-$version` directory.
-4. Run this command: `python3 -m http.server 5000 --bind 127.0.0.1`
-5. You can then use Bridge by navigating to http://localhost:5000 using a web browser.
+Before using Bridge, you'll need access to whatever media you're using to store the Ethereum wallet that owns or will own your Azimuth points. This could be an Urbit HD wallet containing a Master Ticket, a BIP39 mnemonic, a Ledger or Trezor hardware wallet, or an Ethereum private key or keystore file.
```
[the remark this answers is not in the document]
msutherl: Made it simpler in the current commit

msutherl:
@Fang- I'm up for it 🤼‍♂️

README.md
```diff
@@ -10,31 +10,34 @@ Python 3.7.2
 
 ### Instructions
 
-Before using Bridge, you'll need access to whatever media you're using to store the Ethereum wallet that owns or will own your Azimuth points. This could be an Urbit HD wallet containing a Master Ticket, a BIP39 mnemonic, a Ledger or Trezor hardware wallet, or an Ethereum private key or keystore file.
+To use Bridge:
```
Fang-: lol I guess this works (^:

README.md
```diff
@@ -10,31 +10,34 @@ Python 3.7.2
...
 
 1. Install [mkcert](https://github.com/FiloSottile/mkcert)
 2. If you're using Firefox, additionally install [nss](https://github.com/nss-dev/nss)
 3. Install a local certificate authority via `mkcert -install`
 4. In the `bridge-$version` directory, generate a certificate valid for localhost via
   `mkcert localhost`.  This will produce two files: `localhost.pem`, the local
   certificate, and `localhost-key.pem`, its corresponding private key
-5. Run `python bridge-https.py` from the `bridge-$version` directory
-6. Navigate to https://localhost:4443 in a web browser to access Bridge
+5. `cd` into the `bridge-$version` directory
```
Fang-: But step 4 already assumes you're in that dir? This should just swap places with step 4 I think. (And then you can be less verbose in step 6 about "from this dir".)

msutherl: Ah yes – of course. Good catch.

README.md
msutherl: ¯\_(ツ)_/¯

Issue #125, opened 2019-04-25
setTxnConfirmations keep being called after exiting "transaction submitted" screen


Fang-:
When you submit a transaction, you get an overlay screen that tells you the transaction hash, and displays its confirmation status. Confirmation count gets updated for every new confirmation, up to 25 confirmations.

All good, except that keeps happening even after you leave that screen. And this calls `setState` in `Bridge.js`, which causes a re-render for whatever page you're viewing.

Expected behavior here is that it either stops listening for, or stops updating on, new confirmations.

jtobin:
Should be able to use a `componentWillUnmount` to fix that one.

Fang-:
I figured as much, but a very quick look didn't make it clear to me what would need to be unset. The callback is passed into a web3 transaction event handler [here](https://github.com/urbit/bridge/blob/master/src/bridge/lib/txn.js#L160), and I think it's kinda out of our hands from then on? Unless there's some thing you can do here with mutating `confirmationCb` from the outside.

An alternative would be to use a `waitForTransactionConfirm` that doesn't rely on web3.js mechanisms as deeply, such as the one implemented on my dev branch [here](https://github.com/urbit/bridge/blob/mino-onboarding-invites/src/bridge/lib/txn.js#L170).

(Overall, while the functionality web3 offers is nice, I think we can become stronger/more flexible by rolling our own variants to fit our needs, but further thought and discussion is needed.)

jtobin:
This one appears to be resolved, so I'll take the liberty of closing it.

PR #126, opened 2019-04-25
Fix bug where entire application re-renders on setTxnConfirmations

pkova:
Sorry for introducing this bug my dudes!

This PR is a quick fix for #125, but is obviously not very nice. The state flow for the transaction confirmations is very convoluted at the moment so a real fix will need some refactoring. I'll do it straight away when I have time, probably using some polling logic like @Fang- suggested in the issue comments.

I still think it's worth having a PR like this fixing the bug until then.


jtobin:
pkova = stud

PR #127, opened 2019-04-30
Count 0x0 as valid address input

Fang-:
For easier input, instead of having to write many `0`s by hand. This just like works with transaction generation logic.

### DIFF
--- a/src/bridge/components/Base.js
+++ b/src/bridge/components/Base.js
@@ -37,6 +37,9 @@ import DropdownItem from './DropdownItem';
 import DropdownDivider from './DropdownDivider';
 import InnerLabelDropdown from './InnerLabelDropdown';
 
+// Option selectors
+import HorizontalSelector from './HorizontalSelector';
+
 // Path Control
 import Breadcrumb from './Breadcrumb';
 import BreadcrumbItem from './BreadcrumbItem';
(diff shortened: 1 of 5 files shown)


eglaysher:
Drive by, and I understand I have no context here, but doesn't this seem dangerous? Under what circumstances do you want to send things to the 0 address? I don't think we want people to be able to burn address space.

Fang-:
Contract logic prevents sending assets to the zero address, and azimuth-js still checks for that, so nothing changes there. This is just a convenience for cases where you want to unset your proxies. You'd need to explicitly input `0x0000000000000000000000000000000000000000` otherwise.

jtobin:
I don't think I like this.  How about adding an 'unset proxy' button that just sets it to the zero address?  Might make for a little nicer UX.

Fang-:
>How about adding an 'unset proxy' button

On the proxy configuration page? I guess you could have a "set proxy"|"unset proxy" toggle thing that hides the input field for the "unset" case, sure.

Fang-:
I did the thing! @jtobin better?

image

eglaysher:
(FWIW, I'm happier now that we aren't accepting `0x0`. I have no real UI opinions on this tho.)

c-johnson:
LGTM

PR #128, opened 2019-04-30
Add autocomplete to IssueChild

### LINKED ISSUE #121: "Issue child" page doesn't check if the ship is actually a child
For example, if I'm ~marzod, I can generate and send a transaction for spawning ~palfun-foslup, even though that's no child of mine.

It should just use urbit/azimuth-js#25 to get a list of unspawned children, and either provide a dropdown or "randomize" button. This would allow us to also cut out the "check availability" step.

pkova:
Closes #121 

Tear me apart boys.

Alt Text

### DIFF
--- a/src/bridge/views/IssueChild.js
+++ b/src/bridge/views/IssueChild.js
@@ -20,6 +20,15 @@ import {
   eqAddr
 } from '../lib/wallet'
 
+const setFind = (set, pred) => {
+  for (const e of set) {
+    if (pred(e)) {
+      return e
+    }
+  }
+  return undefined
+}
+
 class IssueChild extends React.Component {
   constructor(props) {
     super(props)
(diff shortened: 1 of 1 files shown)


src/bridge/views/IssueChild.js
```diff
@@ -250,6 +224,11 @@ class IssueChild extends React.Component {
             value={ state.desiredPoint }
             onChange={ this.handlePointInput }>
             <InnerLabel>{ 'Point to Issue' }</InnerLabel>
+            <div
+              style={{marginTop: '8.8rem', marginLeft: '4.6rem', fontSize: '5rem', color: '#757575', whiteSpace: 'pre-wrap'}}
```
pkova: I have no idea the classes you're using work, except that tl-0 seems to mean top and left 0! Anyways it's probably terrible to have custom css here but it works.

Fang-: This seems to be doing some very careful positioning and sizing of the text to make it overlap just right. I'm personally fine with leaving this as is, save for maybe splitting the style out over multiple lines.

Fang-:
@pkova I'd love to get in touch more directly! Can you email mark@tlon.io ? If you want, that is!

PR #129, opened 2019-04-30
Help with hardware wallet derivation path selection

Fang-:
Derivation path details for Ledger sourced from [here](https://www.reddit.com/r/ledgerwallet/comments/9nykl8/hd_derivation_path_question/e7qifav/). The Trezor case is more straightforward.

For the Ledger view, I can't get the styling on this to Just Like Work though. The `mt-8` gets overwritten by the `mp-0` that's set inside of `InnerLabelDropdown`'s logic, which is weird and stupid. See screenshot.

Also in the category weird and stupid: having to manually find and submit the title of the currently selected item to the dropdown.

image

### DIFF
--- a/src/bridge/views/Ledger.js
+++ b/src/bridge/views/Ledger.js
@@ -96,15 +96,7 @@ class Ledger extends React.Component {
         value: 'custom'
       }
     ];
-    //NOTE this is dumb
-    let basePathTitle = '';
-    for (let i in pathOptions) {
-      let option = pathOptions[i];
-      if (option.value === basePath) {
-        basePathTitle = option.title;
-        break;
-      }
-    }
+    let basePathTitle = pathOptions.find(o => (o.value === basePath)).title;
 
     let accountOptions = [];
     for (let i = 0; i < 20; i++) {
@@ -113,15 +105,7 @@ class Ledger extends React.Component {
         value: i
       });
     }
-    //NOTE this is dumb
-    let accountTitle = '';
-    for (let i in accountOptions) {
-      let option = accountOptions[i];
-      if (option.value === account) {
-        accountTitle = option.title;
-        break;
-      }
-    }
+    let accountTitle = accountOptions.find(o => (o.value === account)).title;
 
     const basePathSelection = (
       <InnerLabelDropdown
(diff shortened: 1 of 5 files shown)


src/bridge/views/Ledger.js
```diff
@@ -60,7 +81,81 @@ class Ledger extends React.Component {
...
+      { title: 'Ledger Legacy',
+        value: LEDGER_LEGACY_PATH
+      },
+      {
+        title: 'Custom path',
+        value: 'custom'
+      }
+    ];
+    //NOTE this is dumb
+    let basePathTitle = '';
```
c-johnson: You could try using `pathOptions.find(o => o.value === basePath).title` here.

src/bridge/views/Ledger.js
```diff
@@ -60,7 +81,81 @@ class Ledger extends React.Component {
...
+    let accountOptions = [];
+    for (let i = 0; i < 20; i++) {
+      accountOptions.push({
+        title: 'Account #' + (i+1),
+        value: i
+      });
+    }
+    //NOTE this is dumb
+    let accountTitle = '';
+    for (let i in accountOptions) {
```
c-johnson: same `.find` comment as above

src/bridge/views/Trezor.js
```diff
@@ -53,7 +63,57 @@ class Trezor extends React.Component {
...
+    }];
+    for (let i = 0; i < 20; i++) {
+      accountOptions.push({
+        title: 'Account #' + (i+1),
+        value: i
+      });
+    }
+    //NOTE this is dumb
+    let accountTitle = '';
+    for (let i in accountOptions) {
```
c-johnson: same `.find` comment

Fang-:
Good call @c-johnson, thanks! Good now?

c-johnson:
Ah. Looks like className isn't being actually implemented in InnerLabelDropdown. Just adding it in should do the trick: 

```
<div className={`rel table ${fullWidthClass} ${props.className}`}>
```

Fang-:
It was implemented, but on the inner `button` instead of the outer `div`. The latter seems more appropriate, and changing this doesn't seem to break any existing usage.

image

c-johnson:
Looks good then! No further comments.

PR #130, opened 2019-05-01
Support flagging for discontinuity

### LINKED ISSUE #25: Need to be able to signal breaches while setting networking keys
When you set new networking keys, there's an argument called "discontinuous", which should be set if you're creating a new pier, so all your old connection state is gone.  Currently, we default this to false, but we should add a checkbox for it.

Arguably, we should default it to true since most of the time if someone's changing their keys it's because they're breaching their ship.  Counter-argument is that a breach is irreversible, while if you forget to set it you can just cycle the keys again with it set.

Not sure what's the best description for the checkbox that isn't Urbit-specific, but I suggest the following text:

- [ ] This is a discontinuous breach.  This should be set if you're going to clear network state.

Fang-:
Closes #25. The copy was taken from @alexmatz's suggestion there. Let me know if we're still cool with the neutral-ish wording there.

(I'll try to work on a separate PR that merges clickable label into the `CheckboxButton` component.)

### DIFF
--- a/src/bridge/views/SetKeys.js
+++ b/src/bridge/views/SetKeys.js
@@ -190,7 +190,7 @@ class SetKeys extends React.Component {
             onToggle={ this.toggleDiscontinuity }
             state={ state.discontinuity }
             label={`I have lost important off-chain data relating to this point
-                    and need to do a hard-reset.
+                    and need to do a hard reset.
                     (For example, rebooting an Arvo ship.)`}
           ></CheckboxButton>
 
--- a/src/bridge/views/SetKeys.js
+++ b/src/bridge/views/SetKeys.js
@@ -187,13 +187,12 @@ class SetKeys extends React.Component {
 
           <CheckboxButton
             className='mt-8'
-            onClick={ this.toggleDiscontinuity }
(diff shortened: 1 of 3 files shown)


src/bridge/views/SetKeys.js
```diff
@@ -185,6 +185,16 @@ class SetKeys extends React.Component {
               </Warning>
             : <div /> }
 
+          <CheckboxButton
+            className='mt-8'
+            onClick={ this.toggleDiscontinuity }
+            state={ state.discontinuity }
+          >
+            <div style={{cursor: 'pointer'}} onClick={ this.toggleDiscontinuity }>
+              { 'I have lost important off-chain data relating to this point and need to do a hard-reset. (For example, rebooting an Arvo ship.)' }
```
c-johnson: I feel like "hard reset" is better than "hard-reset", but not super important.

Fang-: You're probably right. I'll change it. Good to go otherwise? (:

Fang-: @c-johnson please confirm?

PR #131, opened 2019-05-01
Include label in CheckboxButton component

Fang-:
For more consistent experience, styling, etc.

This PR targets #130 because that creates a `CheckboxButton` that we want updated to fit this. If that gets merged before this, we can just target master instead.


c-johnson:
Looks good to me.

PR #133, opened 2019-05-04
Rename INFURA_KEY environment variable.

jtobin:
`REACT_APP_INFURA_KEY` is not actually a key, and certainly not a secret key, but the public Infura endpoint for the app.  The name can give the impression that we're leaking secret data in the application, but we're not.  `REACT_APP_INFURA_ENDPOINT` is more accurate.

(Also tacks a license badge onto the README.)


jtobin:
@Fang- Literally because I find it to be aesthetically pleasing. 😉😝

PR #135, opened 2019-05-07
Incorporate gas tank into StatelessTransaction

Fang-:
Oof, this was a struggle alright.

This PR makes StatelessTransaction do more or less what the "claim invite" flow does for transactions. If the account can't fund the transaction itself, reach out to the gas tank for help.

Most of the weirdness here deals with the situation where the gas tank submits the transaction for us, causing our submission to fail with "nonce too low". For some reason web3 isn't capable of detecting we're submitting the exact same transaction...

To work around that, we mimic invite flow logic and, observing a nonce error, realize what's going on. We wait for the first confirmation to finalize the flow.

This replaces the existing `sendSignedTransaction` logic, which felt about as jank as this does, for seemingly less necessary reasons.

### DIFF
--- a/src/bridge/components/StatelessTransaction.js
+++ b/src/bridge/components/StatelessTransaction.js
@@ -25,6 +25,7 @@ import {
 
 const SUBMISSION_STATES = {
   PROMPT: 'Send transaction',
+  PREPARING: 'Preparing...',
   FUNDING: 'Finding transaction funding...',
   SENDING: 'Sending transaction...'
 };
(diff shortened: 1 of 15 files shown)


src/bridge/components/StatelessTransaction.js
```diff
@@ -194,16 +179,95 @@ class StatelessTransaction extends React.Component {
...
-  // setState doesn't seem to work in SetProxy/submit;
-  //   - TypeError: Cannot read property 'updater' of undefined
-  // so just modifying the DOM manually here. (https://imgur.com/a/i0Qsyq1)
+  async sendTxn() {
+    const { props, state } = this;
+    const web3 = props.web3.value;
+
+    this.setState({ txStatus: SUBMISSION_STATES.FUNDING });
+
+    const rawTx = hexify(state.stx.value.serialize());
```
jtobin: Should unwrap `stx` here with `matchWith` or similar.

Fang-:
> with state being tracked in multiple components and transaction semantics being tracked in multiple places

I don't think it's that bad. May or may not want to move some of the `sendTxn` logic into `/lib/txn`, but then you'd need a way to get the display state out. Looking forward to the full review!

Fang-:
@jtobin I pushed some more commits to this, since they also touch the StatelessTransaction in non-trivial ways. Primarily, I added a `beforeSend` hook. For an example of how this is used, see [here](https://github.com/urbit/bridge/commit/992f0c39061b5808c5601a027cd537e02c079ead).

PR #138, opened 2019-05-16
Ensure we generate the keyfile for latest-set keys

### LINKED ISSUE #137: Old keyfile gets downloaded after cycling keys
After resetting public keys (either breach in 1.4.0, or non-breach in 1.3.1), there's an option to download a keyfile. If you click on it, you wind up downloading the old keyfile instead of the new one. If you go back to the point right afterwards and select "Generate Arvo keyfile" again, the new one is downloaded.

Fang-:
...rather than latest-seen. Chain might not keep up with user actions, and even if it does, if user navigates to keyfile screen straight away, it won't know about the change yet. Fixes #137.

This was only a problem for UHD-derived networking keys, since the randomly generated ones were pulled out of cache.

This PR also includes an updated azimuth-js, which we actually need for the fancy new auto-complete on the "issue child" page. Currently, it just crashes there. (^:

Will push out a release once this gets approved/merged.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1994,47 +1994,28 @@
       }
     },
     "azimuth-js": {
-      "version": "0.10.0",
-      "resolved": "https://registry.npmjs.org/azimuth-js/-/azimuth-js-0.10.0.tgz",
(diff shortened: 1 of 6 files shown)


vvisigoth:
Is that...a steamboat?
—
~poldec-tonteg
http://urbit.org

jtobin:
_ship it_



.
.
.


_via steamboat_

PR #141, opened 2019-05-20
Update wording around key configuration

### LINKED ISSUE #119: "Set public keys" vs "Set network keys"
This should probably say "set network keys" rather than "set public keys," which is non-specific.

Screenshot 2019-04-19 11 41 3

Fang-:
Fixes #119


msutherl:
LGTM

PR #142, opened 2019-05-28
Make suggestions better in IssueChild.js

pkova:
Suggestion gif

The problem this PR is trying to solve is that the three random point suggestions in IssueChild.js (see gif) do not come from azimuth and may therefore be already spawned. That's annoying, but getting stuff from azimuth can be slow so we don't want any shitty waiting either.

This PR changes the suggestion logic in the following way: instead of generating random children it generates the n first children. When we get a response back from azimuth regarding the actual unspawned children we fix the suggestions with the sick effect pictured above.

Now all this might be too fancy, especially since it introduces a new npm dependency, but I think this looks pretty cool!

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "react": "^16.6.0",
     "react-dom": "^16.6.0",
     "react-router-dom": "^4.3.1",
+    "react-scramble": "^0.4.2",
     "react-scripts": "^2.1.1",
     "secp256k1": "^3.5.2",
     "sigil-js": "github:urbit/sigil-js#f7f29c5",
--- a/src/bridge/lib/child.js
+++ b/src/bridge/lib/child.js
@@ -4,17 +4,11 @@ const MIN_PLANET = Math.pow(2, 16)
 const MAX_GALAXY = MIN_STAR - 1
 const MAX_STAR = MIN_PLANET - 1
 
-const randomStarName = galaxy =>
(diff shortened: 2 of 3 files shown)


Fang-:
I feel like there comes a point where we have to say no to extra code for cool effects. But, when the effects are cool enough, that feeling goes away! (^:

Yet, we must stay vigilant about not letting Bridge get out of hand. I do like the "actually correct" suggestions, but changing them to a new value (instead of hiding, then showing them), especially with an effect like this, feels a bit noisy.

Regardless, I feel the suggestions are kinda deprecated in favor of your autocomplete mechanism anyway? At least, the autocomplete is much closer to what I think the ideal "issue child" page looks like. (If you ask me, just slap a "randomize" button on there and call it a day.)

We don't have designs yet for the "new" issue child page. At this point, I'm not sure if it's worth accepting these kinds of changes, no matter how cool they are.

Thoughts @c-johnson, @alexmatz?

alexmatz:
naturally adverse to new dependencies and yes to your point about new design work on this page. what do you think @vvisigoth?

alexmatz:
paging @vvisigoth

vvisigoth:
Whoa this is great! I think we should definitely put this thinking into the new Spawn Child affordance, but we're not going to merge this PR since this part of the code base is moving on.

cc @jyng what think you of this suggestion mechanism?

PR #150, opened 2019-06-10
import walletgen styles for walletgen appview

shrugs:
this PR imports the walletgen styles and uses the existing mechanism of enabling/disabling the respective stylesheets for each app view.


shrugs:
I wouldn't be surprised at all if there're problems using the super imperative code for swapping stylesheets. Works for me based on a few manual tests. I'm planning on refactoring this as part of the larger goal of updating onboarding UI, though, which will probably land before the next release.

PR #152, opened 2019-06-11
Simpler Routing

shrugs:
The goal of this PR is a few fold:

1. Remove the "prop passing" pattern that's going on across the codebase. The classical solution to this, which got a lot better in React 16, is to use Context to deep-pass state and functions that children might need. Routing is a perfect use-case. I'd also like to explore using [unstated-next](https://github.com/jamiebuilds/unstated-next) which is basically what I already wrote, but with a bit more structure.
2. Refactor the history and routing stuff into their own logical files so that it's not tied up in the Bridge.js stateful monolith component.
3. Begin introducing hooks, which are pretty fantastic for 80% of the stuff we'll be doing in this app. I expect hooks will be really annoying around async state & talking to ethereum, so I don't expect to use them at all in those components.

- updated react and react-dom to latest stable
- removed dangling react-router-dom dependency
- a `route` is now an object with a `name` and `data` property. the `name` is from `ROUTE_NAMES` and is a `Symbol` and `data` is arbitrary, but usually an object.
- history is now accessed with a `withHistory` higher order component or a `useHistory` hook. it offers the same/similar API we expected across the app already
- I made some nested ternarys switch-cases instead because multiline nested ternarys scare me and switch-case is a perfect operator, even if it's more characters (which isn't a problem anywau because minimization/uglification)
- the `routeNames` file has to be separate to avoid a circular dependency between views that use `withHistory/useHistory` and the `router` file which imports those views
- also moves tx confirmations into a separate context store


Fang-:
As discussed over call, this all sounds reasonable and good. I'll want to see this in code also though, so I'll wait that out. In the mean time:

>I expect hooks will be really annoying around async state & talking to ethereum

But that's all Bridge does? Nearly every view does async state/data loading...

Also pinging @jtobin to let us know his thoughts here.

shrugs:
re: async, I think I figured out a pattern for that, actually, modeled after apollo's hooks. I'll be playing around with abstracting the new ethereum loading logic you mentioned later on. Previously, though, I was referencing all of the UI-specific state that we're working on (with mnemonics, networks, contracts, navigation history, that sort of stuff) as good candidates for the hooks API.

src/bridge/lib/wallet.js
```diff
@@ -5,7 +5,7 @@ import * as lodash from 'lodash';
 import Maybe from 'folktale/maybe';
 import * as secp256k1 from 'secp256k1';
 import * as ob from 'urbit-ob';
-import * as kg from '../../../node_modules/urbit-key-generation/dist/index';
+import * as kg from 'urbit-key-generation/dist/index';
```
shrugs: I changed all of these imports - is there a reason we had a relative link into node_modules previously? based on my understanding these should have 0 effect on the code (and I haven't been able to find a difference with manual testing), so I'm curious.

shrugs:
@Fang- it's ready for review! I'm going to keep manually testing this locally.

src/bridge/views/SentTransaction.js
```diff
@@ -146,7 +143,7 @@ const SentTransaction = props => {
           prop-type={'link'}
           onClick={() => {
             setPointCursor(pointCursor);
-            popRoute();
+            history.pop(2);
```
shrugs: @Fang- what's the correct behavior for when a user executes a stateless transaction, finds themselves at SentTransaction, and then presses "ok"? right now, with a single pop, I go back to the statelesstransaction component instead of the point view.

the behavior in the current release goes back to the point view, but the implementation of popRoute was only popping a single entry: https://github.com/urbit/bridge/blob/master/src/bridge/Bridge.js#L179

so there's a bug (?) here where statelesstransaction is in the history when it shouldn't be, I think

shrugs: figured it out — the functions we're passing to these consumers close-over the current value of `history` so when we have code like:

```
history.pop()
history.push(...)
```

the second function call's result is persisted in state, but since it closes over the state during the previous render, we end up ignoring the `.pop` call. The fix is to use the reducer-esq functionality of setState, which is coming in a next commmit

src/bridge/store/txnConfirmations.js
```diff
@@ -0,0 +1,39 @@
...
+      {txnConfirmations => (
+        <Component ref={ref} {...txnConfirmations} {...props} />
+      )}
+    </TxnConfirmationsContext.Consumer>
+  ));
+
+// Hook version
+export function useTxnConfirmations() {
+  return useContext(TxnConfirmationsContext);
+}
```
Fang-: "HOC version" is "do this as/with a component", whereas "hook version" is "do this as hook function", right?

The hook version here doesn't get used anywhere yet, but we'll be working towards that?

src/bridge/views/Network.js
```diff
@@ -34,9 +35,7 @@ class Network extends React.Component {
     if (network === NETWORK_NAMES.ROPSTEN) {
       setNetworkType(network);
 
-      const endpoint = `https://ropsten.infura.io/v3/${
-        process.env.REACT_APP_INFURA_ENDPOINT
-      }`;
+      const endpoint = `https://ropsten.infura.io/v3/${process.env.REACT_APP_INFURA_ENDPOINT}`;
```
Fang-: Again, there may be some `prettier` weirdness here.

src/bridge/views/SentTransaction.js
Fang-: Correct behavior is definitely to kick you back to the point view. So popping twice seems right.

src/bridge/components/DropdownItem.js
```diff
@@ -10,9 +10,7 @@ const DropdownItem = props => {
     <button
       disabled={props.disabled}
       style={props.style}
-      className={`h-10 m-0 p-0 ph-2 bg-white flex align-center items-start bs-solid brw-2 blw-2 btw-0 bbw-0 sans ${_disabled} ${
-        props.className
-      }`}
+      className={`h-10 m-0 p-0 ph-2 bg-white flex align-center items-start bs-solid brw-2 blw-2 btw-0 bbw-0 sans ${_disabled} ${props.className}`}
```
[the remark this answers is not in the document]
shrugs: ah, my editor integration wasn't respecting the ignore file. I've fixed my editor and will revert this file to `mino`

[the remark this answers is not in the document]
shrugs: actually, looks like the `components/` declaration in `.prettierignore` isn't there! I've added it back. and I don't think any other component files have been formatted in the meantime 😅

src/bridge/lib/keys.js
```diff
@@ -21,15 +21,14 @@ const b64 = buf => {
 
   // prettier-ignore
   const trans = j =>
-      10 > j
-    ? j + 48
+    10 > j ? j + 48
```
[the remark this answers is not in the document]
shrugs: oops, that was me manually screwing things up earlier — fixed in a recent commit

src/bridge/store/txnConfirmations.js
shrugs: @Fang- correct. a HOC wraps a component and provides some extra props. the HOC method is the most compatible with the existing code, but hooks are a replacement for the same behavior, so we should move towards that if we end up happy with them.

src/bridge/views/Network.js
shrugs: fixed my editor 👍

shrugs:
@Fang- thanks for the great review - new commits should handle most/all comments

shrugs:
scratch that, one moment - my components reset was a little too aggressive

shrugs:
@Fang- the original reason to not format the components folder was to preserve the components for comparison with indigo, but I don't think that tradeoff is worth it when we're trying to edit files that are also in the components folder. We could either move all of the generic indigo components to a new folder and ignore that one or just enable formatting for the entire directory.

Fang-:
Yeah, I was about to comment that excluding the components folder seemed like a weird move. I'm fine styling them and breaking whatever ties to Indigo they still have. When it comes time to Actually use Indigo we should pull them in as a submodule, or leave them untouched in some other way.

Fixes all seem good!

shrugs:
will wait on an LGTM from @jtobin 👍

src/bridge/lib/wallet.js
jtobin: Yeah I suspect it had something to do with getting things to work w/the rollup build for wallet generator, but can't remember exactly what was up.  Regardless: hacky relative link, begone!

src/bridge/views/Landing.js
```diff
@@ -3,70 +3,75 @@ import { Button } from '../components/Base';
...
+                prop-type={'link'}
+                prop-size={'lg'}
+                className={'mb-4 mt-8'}
+                onClick={() => history.push(ROUTE_NAMES.VIEW_SHIP)}>
+                {'View a point  →'}
+              </Button>
+              <P>{'View an Azimuth point without signing into a wallet.'}</P>
+            </Col>
+          </Row>
+          //  FIXME add address view
```
jtobin: Presumably this old commented-out stuff here can be safely gutted.

src/bridge/components/Header.js
```diff
@@ -31,28 +32,30 @@ const Crumbs = props => {
...
-        pointCursor={props.pointCursor}
-      />
-    </div>
-  );
+}
+
+function Header(props) {
+  const history = useHistory();
+
+  const showCrumbs = !history.includes(ROUTE_NAMES.INVITE_TICKET);
```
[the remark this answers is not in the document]
shrugs: @Fang- which inner view do mean?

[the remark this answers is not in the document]
Fang-: ie specify "should this page display breadcrumbs or not" on the view/component as found in `/src/bridge/views` (the different "pages" we route to), rather than having a check like this in `Header`.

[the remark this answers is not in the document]
shrugs: ah, yeah, that's totally something we should decide on, pattern-wise. are we making a distinction between "dumb" components that just interact with their props and isolated state and "connected" components that know about higher-level state and use our HOCs/Hooks for connecting data sources to dumb components. In our case, views would be connected components and ideally all of the indigo components and re-usable bridge-specific things are "dumb".

If that's a distinction we want to maintain, I can make it a priority to separate the concerns in the next few pull requests.

Personally in recent projects I've ended up giving up on dumb vs connected component distinctions (except for isolated libraries like our indigo-react) just because it was a pain to handle all of the boiler plate and maintain multiple component definitions & files per component.

here's dan's thoughts on this concept from 2015 for context https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0

PR #153, opened 2019-06-11
fix: reformat files

shrugs:
these files seem to have escaped formatting, so I ran `npm run lint:fix` on them


shrugs:
just going to merge this right away cause it's blocking my diffs and I don't think it'll be controversial 👍

shrugs:
should have waited the extra 5 minutes 😅

Issue #140, opened 2019-05-19
Can't continue after authenticating with Ledger


Fang-:
```
~satdef+ One of my friends is trying to get started, but can't auth with
         Bridge with his Ledger Nano S (1.5.5).  Are there any known issues
         with this?
~satdef+ After authing, 'continue' button never turns blue
...
~satdef+ palfun, he managed to auth when downgrading to 1.3.1
```

Fang-:
Hmm, cannot reproduce with my own Ledger Nano S, which runs SE firmware 1.4.2 and MCU firmware 1.5. Continue button work just fine, regardless of the selected account. Contract/display data settings don't seem to matter.

I suspect ~satdef's friend was accessing Bridge over HTTP instead of HTTPS. Ledger refuses to authenticate in that case, and the continue button does stay greyed out. The Ledger login page gives instructions for this, and they're included in the repo readme as well.

msutherl:
I ran through this flow recently and it was pretty fiddly. I can't remember exactly, but we ran into the above issue a few times until we got the settings and/or timing right.

jtobin:
A Chrome update broke Ledger support awhile back -- I kind of figure it has been resolved since, but am not sure.  I last used a Ledger successfully w/Brave.

shrugs:
_should_ be fixed as part of #155, but will wait for double-confirmation before closing this

PR #155, opened 2019-06-12
Update Dependencies

shrugs:
- react-scripts -> 3.x
- ledger libs -> 4.61.0
- lodash -> lodash-es

this also happens to bring our audit results back down to 0 vulnerabilities 🎉 

I updated react-scripts to 3.0 [following the migration guide](https://github.com/facebook/create-react-app/blob/master/CHANGELOG.md#migrating-from-21x-to-300)

I removed babel-polyfill, intending to replace it with `regenerator-runtime/runtime`, [according to the migration guide](https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#babelpolyfill) but found that I didn't get an error when loading the ledger page (afaik, the error would have been thrown immediately because of the implicit global dependency on the regenerator-runtime features).

I'm going to fully test the ledger flow on Monday using the office ledger, unless you happen to have one, @Fang-.

I can't tell if this is related to this PR or not, but my hotreloading also broke in firefox, and [this](https://github.com/facebook/create-react-app/issues/6720) seems like the relevant issue.

I also added an `analyze` script that shows bundle sizes. The azimuth-js check.js file takes up a _massive_ amount of space, so probably worth debugging that. Likewise for urbit-key-generation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,26 +13,36 @@
       }
     },
     "@babel/core": {
-      "version": "7.1.6",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.1.6.tgz",
-      "integrity": "sha512-Hz6PJT6e44iUNpAn8AoyAs6B3bl60g7MJQaI0rZEar6ECzh6+srYO1xlIdssio34mPaUtAb1y+XlkkSJzok3yw==",
+      "version": "7.4.3",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.4.3.tgz",
+      "integrity": "sha512-oDpASqKFlbspQfzAE7yaeTmdljSH2ADIvBlb0RwbStltTuWa0+7CCI1fYVINNv9saHPa1W7oaKeuNuKj+RQCvA==",
       "requires": {
         "@babel/code-frame": "^7.0.0",
-        "@babel/generator": "^7.1.6",
-        "@babel/helpers": "^7.1.5",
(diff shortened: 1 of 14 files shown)


shrugs:
> lodash

`lodash-es` is the same as good ole' `lodash`, but is just an ES module. by using an ES module instead of the normal commonJS, the bundler should be able to avoid bundling functions that we don't use, but after actually testing, that doesn't seem to be the case. not a big deal at the moment, so I've reverted to normal lodash

> Interesting! It's pulling in all the other contract-specific libs, but I'd hope that gets deduplicated or something?

it doesn't have a chance to get deduplicated because we're just pulling in the `/dist/` compiled blob. If we exported an ES module I think webpack might dedupe? Not entirely sure

Fang-:
_puts a dollar in the js-ecosystem-headache jar_

Merge at will!

shrugs:
@Fang- let me know if this branch works with ledger? just in case, can you also `rm -rf ./node_modules` and then `npm install` to get a clean slate?

Fang-:
I _always_ burn `node_modules` after doing checkouts. Been burned by that too many times already!

It was initially failing silently, but after updating my Ledger firmware it behaves normally. So, all good!

shrugs:
@Fang- great! merged! did you test chrome, specifically? #140 seems chrome related

Fang-:
I did now! Can authenticate just fine.

PR #164, opened 2019-06-17
Extract network + bugfixes

shrugs:
- extract network seed / revision into wallet store since it's identity-specific
- fix some bugs I uncovered:
  1. setPointCursor(string) was being used, but pointCursor expects number
  2. local networkSeed was not respected within GenKeys.js

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -2,3 +2,4 @@ REACT_APP_DEV_MNEMONIC='benefit crew supreme gesture quantum web media hazard th
 REACT_APP_DEV_GALAXY_TICKET='~wacfus-dabpex-danted-mosfep-pasrud-lavmer-nodtex-taslus-pactyp-milpub-pildeg-fornev-ralmed-dinfeb-fopbyr-sanbet-sovmyl-dozsut-mogsyx-mapwyc-sorrup-ricnec-marnys-lignex'
 REACT_APP_ROPSTEN_KEY='b7d2af9f01534031ba773374f766ef65'
 REACT_APP_HD_PATH="m/44'/60'/0'/0/0"
+REACT_APP_STUB_LOCAL='false'
--- a/.env.development
+++ b/.env.development
@@ -2,4 +2,3 @@ REACT_APP_DEV_MNEMONIC='benefit crew supreme gesture quantum web media hazard th
 REACT_APP_DEV_GALAXY_TICKET='~wacfus-dabpex-danted-mosfep-pasrud-lavmer-nodtex-taslus-pactyp-milpub-pildeg-fornev-ralmed-dinfeb-fopbyr-sanbet-sovmyl-dozsut-mogsyx-mapwyc-sorrup-ricnec-marnys-lignex'
 REACT_APP_ROPSTEN_KEY='b7d2af9f01534031ba773374f766ef65'
 REACT_APP_HD_PATH="m/44'/60'/0'/0/0"
-REACT_APP_STUB_LOCAL='false'
--- a/README.md
+++ b/README.md
@@ -52,13 +52,14 @@ To validate your downloaded file's integrity, compare the lines in checksum.txt 
 
 ### Install / Build
 
-Clone the repo, and use a simple `npm install`. You can then use a `npm run build` to create an optimised static build (serve it with e.g. [serve][serv]).
+Clone the repo, and use a simple `npm install`.  You can then use a `npm run
+build` to create an optimised static build (serve it with e.g. [serve][serv]).
 
 ### General notes
 
-For development, use `npm run pilot` to get going after a `npm install`. This
+For development, use `npm run pilot` to get going after a `npm install`.  This
 will boot up a Ganache node in the background, deploy the Azimuth contracts to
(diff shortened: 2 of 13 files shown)


src/bridge/views/SentTransaction.js
```diff
@@ -148,7 +146,6 @@ function SentTransaction(props) {
         <Button
           prop-type={'link'}
           onClick={() => {
-            setPointCursor(pointCursor);
```
shrugs: this line didn't seem like it was doing anything, since `setPointCursor(pointCursor)` would just produce the same state. @Fang- any insight here?

Fang-: That does indeed look like a no-op. And one that's been around since forever at that! Seems fine to take it out if it doesn't break anything.

src/bridge/views/ViewPoint.js
```diff
@@ -71,7 +71,8 @@ class ViewPoint extends React.Component {
             className={'mt-8'}
             disabled={valid === false}
             onClick={() => {
-              setPointCursor(Maybe.Just(ob.patp2dec(pointName)));
+              setPointCursor(Maybe.Just(parseInt(ob.patp2dec(pointName), 10)));
```
shrugs: the latest version of urbit-ob has a `patp2bn` function that we could use `ob.patp2bn(pointName).toNumber()` if we think that's better than using `parseInt` on the string value here

Fang-: Maybe `2dec` should just give you a number like it says it'll do. (^:

But I'm fine using `2bn` until that happens.

shrugs: @Fang- looks like [2bn isn't exported](https://github.com/urbit/ob-js/blob/master/src/internal/co.js#L143) so I think this is fine to leave for now.

Fang-: (Just remembered it's not giving you a number because it supports working with numbers larger than Javascript's supported limit. Luckily, in the Bridge case, we're bounded to 32-bit numbers thanks to not caring about moons and comets.)

shrugs: hmm, in that case I'll create an issue to help us migrate to 100% bignumber, since it's sorta lucky that we won't be bitten by that for planets. It's odd to me that the azimuth-js call returns javascript numbers, which is probably why we're using the native number type everywhere else in bridge.

Fang-: > migrate to 100% bignumber

Inside Bridge? Yuck, that doesn't feel comfy at all. We don't really need bignums anywhere, do we? All web3 stuff that could need it is strings, and that's it?

> It's odd to me that the azimuth-js call returns javascript numbers

iirc most of these just return whatever came out of the under-the-hood web3 calls. Though there may be weird cases. (And in azimuth-js, too, we don't expect to be dealing with >32-bit numbers.)

shrugs: ah, are moons and comets never actually on-chain?

Fang-: Nope. Azimuth has a strictly 32-bit namespace, and doesn't have a concept of moons or comets. They're an Arvo thing, and "owned" by their planet and nobody respectively.

shrugs: sounds great, then! will ignore bignum problems 👍

.env.development
```diff
@@ -2,3 +2,4 @@ REACT_APP_DEV_MNEMONIC='benefit crew supreme gesture quantum web media hazard th
 REACT_APP_DEV_GALAXY_TICKET='~wacfus-dabpex-danted-mosfep-pasrud-lavmer-nodtex-taslus-pactyp-milpub-pildeg-fornev-ralmed-dinfeb-fopbyr-sanbet-sovmyl-dozsut-mogsyx-mapwyc-sorrup-ricnec-marnys-lignex'
 REACT_APP_ROPSTEN_KEY='b7d2af9f01534031ba773374f766ef65'
 REACT_APP_HD_PATH="m/44'/60'/0'/0/0"
+REACT_APP_STUB_LOCAL='false'
```
Fang-: This is good to have in here, but can we get unrelated changes as separate commits next time? Same goes for some minor cleanup you did in this PR. The more single-purpose a commit is, the better imo. (:

README.md
```diff
@@ -52,14 +52,13 @@ To validate your downloaded file's integrity, compare the lines in checksum.txt
 
 ### Install / Build
 
-Clone the repo, and use a simple `npm install`.  You can then use a `npm run
-build` to create an optimised static build (serve it with e.g. [serve][serv]).
+Clone the repo, and use a simple `npm install`. You can then use a `npm run build` to create an optimised static build (serve it with e.g. [serve][serv]).
```
Fang-: I think you accidentally linted the readme?

.env.development
shrugs: yeah totally  — will move those changes to a new PR

README.md
shrugs: fixed!

.env.development
Fang-: "Will do in the future" would've been fine too, but thanks. (;

PR #167, opened 2019-06-17
Extract txcursor

shrugs:
the final one! moving the txn cursor into its own container

### DIFF
--- a/src/bridge/components/StatelessTransaction.js
+++ b/src/bridge/components/StatelessTransaction.js
@@ -25,6 +25,7 @@ import { withTxnConfirmations } from '../store/txnConfirmations';
 import { withNetwork } from '../store/network';
 import { withHistory } from '../store/history';
 import { withWallet } from '../store/wallet';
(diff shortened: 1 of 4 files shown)


src/bridge/components/StatelessTransaction.js
```diff
@@ -238,7 +239,7 @@ class StatelessTransaction extends React.Component {
         props.setTxnConfirmations
       )
         .then(txHash => {
-          props.onSent(Just(Ok(txHash)), state.stx.value);
```
shrugs: the second argument here was just ignored by the `onSent` function, so I'm removing it here.

Fang-: The reason the transaction is included here is that some flow implementations may depend on data from it. We previously had one such case, but I think that was refactored out?

I guess since it's not actively in use we should just remove it. If anything ever needs this, it can be added back in. (Though, better to be consistent about how transaction submission flows work.)

PR #169, opened 2019-06-17
stub env with environment variable, update README

shrugs:
instead of using a hard-coded boolean to toggle stubbing, let's use an env variable

```bash
REACT_APP_STUB_LOCAL=true npm run start
```

is how I personally enable the auto-login functionality now

my editor also auto-formatted the README while I was adding a note about the env variable. ¯\\\_(ツ)_/¯


shrugs:
> yet

😱

PR #171, opened 2019-06-18
Mino login

Fang-:
Still WIP at the moment, but this PR implement a new login flow which hides network selection, visually simplifies login methods, and intelligently chooses between "points list" and "point page".

Opening this as a draft so that @shrugs can tell me whether or not my pattern for integrating the login methods here is sane. tl;dr: we swap in sub-views from `/views/Login/` based on the tab/dropdown the user has selected, and defer back to the `Login` view/component for finalization.

The things in `/views/Login/` are initially going to be near-identical to their pre-existing counterparts in `/views/`, but should diverge to whatever degree the designs desire.  
(I also like how this (eventually) leads to a slightly more organized `/views/` directory, with login-related views grouped under `/views/Login/`.)

Also how do I get at `initialNetworkType` [here](https://github.com/urbit/bridge/blob/11c42d9489bdc35554bb6bf69701784d6334b5aa/src/bridge/views/Login.js#L54-L55)?

Remaining work for me on this is to integrate the remaining login methods into the advanced tab and clean up some of the TODOs I put in.


shrugs:
for initialNetworkType, because you're using the `withNetwork` HOC you'll have `props.networkType` available in that component — would that work?

Fang-:
Ah! I somehow hadn't noticed that it also sets `.networkType` to the initial one right at the beginning. That makes a lot more sense now. (:

src/bridge/views/Login.js
```diff
@@ -0,0 +1,138 @@
...
+          contracts,
+          wallet.address
+        );
+        if (canOwn.length === 1) {
+          this.props.setPointCursor(Just(canOwn[0]));
+        }
+      }
+    }
+
+    if (Just.hasInstance(this.props.pointCursor)) {
```
shrugs: is the intention to have `this.props.pointCursor` available after `this.props.setPointCursor` above? If so, I don't know that the `this.props.pointCursor` would be equal to the new value until after the next render.

src/bridge/views/Login.js
```diff
@@ -0,0 +1,138 @@
...
+  handleTabChange(currentTab) {
+    this.setState({ currentTab });
+  }
+
+  async continue() {
+    const wallet = need.wallet(this.props);
+    const contracts = need.contracts(this.props);
+
+    // if no point cursor set by login logic, try to deduce it
+    if (Nothing.hasInstance(this.props.pointCursor)) {
```
shrugs: could use

```js
this.props.setPointCursor(this.props.pointCursor.matchWith({
  Nothing: () => { ... },
  Just: p => p.value
}))
```

if it makes it cleaner

src/bridge/views/Login.js
Fang-: Hmm you might be right. The whole "when to props/state update" thing still feels not smooth/intuitive to me.  
Should we have a local `Maybe(pointCursor)` that we also check for, then? ie, "if it was already set, _or_ we just deduced it, ..."

shrugs: I would just include this logic in the same branch above. if we're setting pointcursor, also popAndPush POINT_HOME, otherwise it'll fallthrough to SHIPS

Fang-:
@shrugs Let's try and get this merged/combined with #172. I see you already did some work on Passport component and related views there, would love to start getting that into PointHome.

We'll probably talk about this merge more during call today. In the mean time, this is ready for review, though there's really not that much interesting stuff here.

src/bridge/views/PointHome.js
```diff
@@ -0,0 +1,148 @@
...
+          azimuth.delegatedSending
+            .getTotalUsableInvites(ctrcs, point)
+            .then(count => this.setState({ invites: Just(count) }));
+          this.updateSpawned(ctrcs, point);
+        })
+      )
+    );
+  }
+
+  async updateBirthday(blockNumber) {
```
shrugs: Note to self: include this birthday logic in `indigo/Point.js`

src/bridge/views/Login.js
```diff
@@ -0,0 +1,131 @@
...
+      currentTab: TABS.TICKET,
+      wallet: Nothing(),
+    };
+
+    this.handleTabChange = this.handleTabChange.bind(this);
+    this.continue = this.continue.bind(this);
+  }
+
+  componentDidMount() {
+    this.props.setNetworkType(this.props.networkType);
```
shrugs: this line won't actually do anything, since the initial network type is already the same as `props.networkType`

src/bridge/views/Login.js
```diff
@@ -0,0 +1,131 @@
...
+    } else if (Just.hasInstance(this.props.pointCursor)) {
+      this.props.history.popAndPush(ROUTE_NAMES.POINT_HOME);
+    } else {
+      //TODO to new overview (maybe should be merged into point home?)
+      this.props.history.popAndPush(ROUTE_NAMES.SHIPS);
+    }
+  }
+
+  render() {
+    const tabOptions = [
```
shrugs: move this outside of the component to where TABS is?

src/bridge/views/Login.js
```diff
@@ -0,0 +1,131 @@
...
+  ADVANCED: Symbol('ADVANCED'),
+};
+
+class Login extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      currentTab: TABS.TICKET,
+      wallet: Nothing(),
```
shrugs: do you think it's possible to use the global wallet state setter/getter instead of storing it locally and then committing it to the global state once the user presses continue?

src/bridge/views/Login/Keystore.js
```diff
@@ -0,0 +1,162 @@
...
+  handlePasswordInput(password) {
+    this.setState({ password });
+  }
+
+  constructWallet() {
+    const { setWallet } = this.props;
+    const { keystore, password } = this.state;
+
+    try {
+      const text = keystore.matchWith({
```
shrugs: could be a  `need` guard

src/bridge/views/Login.js
shrugs: like how mnemonic uses `setWallet`

src/bridge/views/Login.js
shrugs: nitpick: can we use `key` instead of `value` for the tab key?

src/bridge/views/Login.js
```diff
@@ -0,0 +1,131 @@
...
+
+    //TODO probably needs a dedicated component because styling
+    const tabs = (
+      <HorizontalSelector
+        options={tabOptions}
+        onChange={this.handleTabChange}
+      />
+    );
+
+    const login = (() => {
```
shrugs: this tabs/login pair could be its own component like

```js
function Tabs({ tabs, options, currentTab, onChangeCurrentTab, ...rest }) {
  const Component = tabs[currentTab];
  return <>
    <HorizontalSelector options={options} onChange={onChangeCurrentTab} />
    <Component {...rest} />
  </>
}
```

src/bridge/views/Login/Mnemonic.js
```diff
@@ -0,0 +1,160 @@
...
+        passphrase
+      );
+      return {
+        passphrase,
+      };
+    });
+  }
+
+  handleMnemonicInput(mnemonic) {
+    this.setState((state, _) => {
```
shrugs: this code duplication should definitely be its own separate function, but I also already fixed this in the indigo branch, so no need to duplicate the work here. just an fyi

src/bridge/views/Login.js
Fang-: You're right. I'm imagining cases though where we return to Login after having been in a different flow that may have used a different network. Very unlikely of course, but it feels nice having this guarantee, instead of needing to think about how Login might interact with other functionality.

Does than sound sane?

src/bridge/views/Login.js
Fang-: I just mimicked what was done in other cases for `HorizontalSelector`. But `key` does also work. I'm fine changing this, but can you explain why you prefer it?

Fang-: Wait no I didn't reload properly. Changing this actually does break `HorizontalSelector`, which checks options for `.value` instead of `.key`.

src/bridge/views/Login.js
Fang-: I think that's what we already do, this line can just be purged.

src/bridge/views/Login.js
shrugs: even then, it wouldn't reset to the initial network type, but only re-set the current network

if we want a "reset to existing network" behavior

1) let's not add that until we actually need it (the figma designs don't have anything about networks, actually - jimmy confirmed that it isn't something they want normal users to toggle)
2) we'd put that within the `_useNetwork` hook itself as a function like `resetNetwork()` that sets the state to the `initialNetwork` that was provided to the hook.

Fang-:
Good feedback, thanks! I'll get on it.

> will be refactoring these into hooks as part of indigo

I might try and take a stab at that prior to merge if you don't object. There's plenty examples on how these should work around now, and should help me get a better feel for them.

src/bridge/views/Login.js
shrugs: I was suggesting to also change horizontal selector - if we don't want to change it that's cool. the semantics are wrong, since that's a key (not a value) is all

shrugs:
@Fang- go for it! check out indigo/Mnemonic for a specifically relevant example

src/bridge/views/Login.js
Fang-: _cleans glasses_ Oh, I see what you're saying. This is a straight no-op. (^:

Fair on both points!

src/bridge/views/Login.js
shrugs: there are a few `setState({ wallet })` calls in the other components as well

src/bridge/views/PointHome.js
Fang-: Note to self: birthday should go into point cache also, this one's especially slow.

Fang-:
Going ahead and merging on the assumption that this is fine. @shrugs feel free to still review whenever, if you feel like it. I'll be proceeding with dev as discussed.

PR #172, opened 2019-06-19
WIP: Indigo / Multipass Overhaul

shrugs:
WIP for @Fang- to 👀 at while it's in heavy flux. Doing my best to make the project perfectly buildable at every commit.


src/lib/need.js
```diff
@@ -1,75 +1,32 @@
...
-      })()
-    : point in obj.pointCache
-    ? obj.pointCache[point]
-    : (() => {
-        throw BRIDGE_ERROR.MISSING_POINT;
-      })();
-}
-
-export { web3, contracts, wallet, address, pointCursor, fromPointCache };
+const needBuilder = error => obj => {
```
Fang-: Neat. (:

A little sad about needing to think about what argument to pass in again, just blindly throwing `props` in felt really nice. But I guess this is more flexible wrt hook pattern support?

src/views/CancelTransfer.js
```diff
@@ -56,20 +57,18 @@ class CancelTransfer extends React.Component {
...
-          <P>{`This action will cancel the transfer to ${proxy}.`}</P>
-          <StatelessTransaction
-            canGenerate={canGenerate}
-            createUnsignedTxn={this.createUnsignedTxn}
-          />
-        </Col>
-      </Row>
+      <View>
+        <H1>
+          {'Cancel Transfer of '}{' '}
```
Fang-: (this was already here but) what's up with this? No reason for this to exist, right? (The extra `{' '}`, that is.)

src/views/InvitesManage.js
```diff
@@ -140,44 +139,42 @@ class InvitesManage extends React.Component {
...
-          </Input>
-
-          <StatelessTransaction
-            canGenerate={this.state.canGenerate}
-            createUnsignedTxn={this.createUnsignedTxn}
-            ref={this.statelessRef}
+      <View>
+        <p>
+          {
+            'manage invites here, for stars. can only give invites to child planets'
```
Fang-: Just a heads up, this is description is outdated. Stars can now grant invites to anyone.

src/components/Inputs.js
```diff
@@ -0,0 +1,81 @@
...
+
+import { validateNotEmpty, validateMnemonic } from 'lib/validators';
+import { DEFAULT_HD_PATH } from 'lib/wallet';
+
+const kExampleMnemonic = bip39.generateMnemonic();
+
+// const RequiredInput = advancedInput({
+//   WrappedComponent: Input,
+//   validators: [validateNotEmpty],
+// });
```
Fang-: What are these commented blocks doing here?

src/store/wallet.js
```diff
@@ -36,45 +42,48 @@ function _useWallet(
   const [networkSeed, setNetworkSeed] = useState(Maybe.Nothing());
   const [networkRevision, setNetworkRevision] = useState(Maybe.Nothing());
 
-  const setWalletType = walletType => {
+  const setWalletType = useCallback(walletType => {
```
Fang-: Can you tldr me on what this achieves?

src/components/Buttons.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import { Button } from 'indigo-react';
+
+// TODO: use svg icons
+export const ForwardButton = props => <Button icon="→" {...props} />;
+export const DownloadButton = props => <Button icon="⬇" {...props} />;
```
Fang-: In some cases I imagine we may want to swap out the icons on buttons dynamically (ie show a spinner, then show an arrow). Just displaying a different component seems fine, but is that the most sane way to achieve that?

Fang-:
> Doing my best to make the project perfectly buildable at every commit.

Hmm, should we maybe merge the `mino-login` branch with this, so that we can work on the new login pages directly? Otherwise I'll hold off on pulling existing login pages into `/views/Login/` until you're done touching those here.

src/lib/need.js
shrugs: agreed on both, yeah

src/views/CancelTransfer.js
shrugs: in some cases you need a `{' '}` to emulate a `&nbsp;` (nonbreaking space character) because jsx will collapse new lines when rendering. There are a _ton_ of uses of `<div>{'Some text'}</div>` in the codebase where `<div>Some text</div>` is semantically identical, so I'm not sure how/when/why that pattern was adopted. I'll be changing them as I go and will make sure I don't break any expectations the UI has for these spaces.

src/views/InvitesManage.js
shrugs: cool, will file that away for a copy pass 👍

src/components/Inputs.js
shrugs: These are the future inputs I've yet to convert to the new Input component — just kept them around for documentation.

src/store/wallet.js
shrugs: useCallback memoizes a function so that it has referential equality between renders where the dependencies don't change.

pre-`useCallback`, any time a component `_useWallet`ed, it would get a new value for `setWalletType`. If it passed that `setWalletType` to a component as props, that component would re-render every time, since `setWalletType` is a unique reference every render. With useCallback, only when the dependencies of `setWalletType` (which is only the `_setWalletType` function) change does `setWalletType` become a new function, meaning we can avoid a re-render in any child we pass this function to.

In this case, `_setWalletType` is the only dependency, and the state setting functions from `useState` have stable identities between renders, so it's safe to omit.

> Even in that case, dispatch identity is still guaranteed to be stable between re-renders. So you may omit it from the effect deps if you want. It’s not going to cause the effect to re-run.

(from dan's blog post below)

I will include the dependencies anyway, though, since being explicit is a great idea.

would recommend https://overreacted.io/a-complete-guide-to-useeffect/ and https://kentcdodds.com/blog/usememo-and-usecallback for reference

src/components/Buttons.js
shrugs: I think I'll be creating another component called `LoadableButton` that supports hot-swapping icons and the background loading behavior from figma. For the static buttons this feels right, though.

shrugs:
the login merge might be pretty tough, depending on which components were touched. tbh I think the move might be to rebase this branch on top of mino-login — git should three-way merge the existing changed files and then I'll have to manually apply my updates around the `View` component to the new view you created, but that should be it?

src/views/InvitesManage.js
Fang-: I wouldn't worry too much about copy if I were you! Your time is better spent elsewhere. (:

Fang-:
Alright, I'll go and put in the remaining mino-login work then!

src/lib/routeNames.js
```diff
@@ -3,7 +3,7 @@ export const ROUTE_NAMES = {
   //
   INVITE_TICKET: Symbol('INVITE_TICKET'),
   INVITE_TRANSACTIONS: Symbol('INVITE_TRANSACTIONS'),
-  INVITES_SEND: Symbol('INVITES_SEND'),
+  INVITE: Symbol('INVITE'),
```
Fang-: Why the renaming? I liked the explicitness, since there's a good few pages relating to invites. (Though this should probably be renamed still per the next commit, to `INVITE_EMAIL` or similar.)

src/indigo-react/components/Flex.js
```diff
@@ -0,0 +1,61 @@
...
+  row = false,
+  wrap = false,
+  col = false,
+  align,
+  justify,
+  className,
+  ...rest
+}) {
+  if (row && col) {
+    throw new Error('Only one of row or col must be true, not both.');
```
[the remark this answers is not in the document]
shrugs: both of them being false(/falsy) is actually ok (and the default state is both falsy) — the default behavior is just to apply the `flex` class and not the `flex-col` or `flex-row` class.

src/lib/routeNames.js
shrugs: In the latest commit I pulled the router out into a separate hook and made it more generic so that it can be used in a nested fashion as well. In the mocks we have a navigation to invite but then within the Invite page a subsection of the page navigates between types of invites. So for the toplevel router, the only route it needs to know about is the "Invite" route and then the Invite view has its own router to navigate between states. Only the primary router captures the browser back button, etc.

shrugs:
TODO: make useSyncPoint only trigger on point change + have a manual invalidation function for key cycling

Fang-:
Whew, merge is in!

At first:
<details>
<summary>Whaaat, Bridge looks like this now? (:</summary>
image
</details>

But then:

<details>
<summary>Whaaat, Bridge looks like <i>this</i> now?</summary>
image
</details>

The "shiny new login flow" pages I merged in are, of course, even worse. (^:  
But this is all feeling pretty close already!

shrugs:
lmao yeah the old components without styling are super jarring. glad the merge worked out! I'm excited to build off of this on monday. I'll sync with you over email on how we should work off this branch

PR #174, opened 2019-06-24
Functional-ify Login views

Fang-:
Fairly straightforward conversions. Switched to newer components where I could, but some (like `InnerlabelDropdown` don't have obvious replacements yet). The hardware wallet login pages are especially hideous, and Bridge Team should think about how to present those going forward.

I think we're fine to completely tear out the old login flow and its files now? RIP network selection.


src/views/Login/Keystore.js
```diff
@@ -69,83 +63,65 @@ class Keystore extends React.Component {
...
+        {`Please upload your Ethereum keystore file.  If your keystore
+         file is encrypted with a password, you'll also need to enter
+         that below.`}
+      </InputCaption>
+
+      <UploadButton
+        className={`${uploadButtonClass} mt3`}
+        onChange={handleKeystoreUpload}>
+        <div className={'flex-center-all fs-4 h-11 pointer'}>
+          {'Upload Keystore file'}
```
shrugs: You don't need the `{''}` syntax here—`<div>Upload Keystore file</div>` is semantically identical. Likewise across all of these files. the lone `{' '}` is only required when a formatted newline should be a space instead of being automatically collapsed by react.

src/views/Login/Keystore.js
```diff
@@ -69,83 +63,65 @@ class Keystore extends React.Component {
...
+        className={`mt3`}
+        disabled={Nothing.hasInstance(keystore)}
+        onClick={constructWallet}>
+        Decrypt
+      </ForwardButton>
+
+      {decryptMessage}
+
+      <ForwardButton
+        className={'mt3'}
```
shrugs: here and around these files, this can be `className="mt3"` since they're not variables or interpolated.

src/views/Login/Keystore.js
```diff
@@ -1,66 +1,60 @@
...
-import { Input, InnerLabel, InputCaption } from '../../components/old/Base';
-import { Row, Col, H1, H3, Warning } from '../../components/old/Base';
+
+import View from '../../components/View';
+import { PassphraseInput } from '../../components/Inputs';
+import { ForwardButton } from '../../components/Buttons';
+import {
+  InputCaption,
+  UploadButton,
+  H3,
```
shrugs: Go ahead and grab H3 from `indigo-react` if it makes sense

src/views/Login/Ledger.js
```diff
@@ -1,69 +1,55 @@
...
-import { Button } from '../../components/old/Base';
-import { Row, Col, H1, P, H2 } from '../../components/old/Base';
+import Transport from '@ledgerhq/hw-transport-u2f';
+import Eth from '@ledgerhq/hw-app-eth';
+import * as secp256k1 from 'secp256k1';
+
+import View from '../../components/View';
+import { ForwardButton } from '../../components/Buttons';
 import {
+  H1,
```
shrugs: we can use `indigo-react` here if it makes sense

src/views/Login/Ledger.js
```diff
@@ -72,153 +58,135 @@ class Ledger extends React.Component {
...
-        setWalletHdPath(addHdPrefix(hdpath));
+        setWalletHdPath(addHdPrefix(hdPath));
       },
       _ => {
         setWallet(Maybe.Nothing());
       }
     );
+  };
+
+  const pathOptions = [
```
shrugs: can pull these path options out of render as well

src/views/Login/Ledger.js
```diff
@@ -72,153 +58,135 @@ class Ledger extends React.Component {
...
+    { title: 'Ledger Live', value: LEDGER_LIVE_PATH },
+    { title: 'Ledger Legacy', value: LEDGER_LEGACY_PATH },
+    {
+      title: 'Custom path',
+      value: 'custom',
+    },
+  ];
+  const basePathTitle = pathOptions.find(o => o.value === basePath).title;
+
+  let accountOptions = [];
```
shrugs: we can write this as 

```js
import {times} from 'lodash';

const accountOptions = times(20, (i) => ({ title: `Account #${i + 1}`, value: i }))
```

and also pull it out of render

src/views/Login/Ledger.js
```diff
@@ -72,153 +58,135 @@ class Ledger extends React.Component {
...
-            onClick={this.props.loginCompleted}>
-            {'Continue →'}
-          </Button>
-        </Col>
-      </Row>
-    );
-  }
+  return (
+    <View>
+      <H1>{'Authenticate With Your Ledger'}</H1>
```
shrugs: so many `{'...'}` going on here... only refactor if you think it's worth it ¯\\\_(ツ)_/¯

src/views/Login/Ledger.js
```diff
@@ -72,153 +58,135 @@ class Ledger extends React.Component {
...
+            transactions, you'll also need to enable the "contract data"
+            option.`}
+      </P>
+
+      <P>
+        {`If you'd like to use a custom derivation path, you may enter
+             it below.`}
+      </P>
+
+      <H2>{'Running on HTTP?'}</H2>
```
shrugs: we can actually detect if the browser is in http or https by checking `location.protocol` if we want to be clever about this screen

src/views/Login/PrivateKey.js
```diff
@@ -1,80 +1,50 @@
...
-  InputCaption,
-} from '../../components/old/Base';
-import { Row, Col, H1 } from '../../components/old/Base';
+import { Just, Nothing } from 'folktale/maybe';
+import React, { useState } from 'react';
 
-import { EthereumWallet } from '../../lib/wallet';
-import { compose } from '../../lib/lib';
-import { withWallet } from '../../store/wallet';
+import View from '../../components/View';
```
shrugs: we can actually get rid of these relative imports now that we look for modules in `src`: `import View from 'components/View'` should work perfectly. likewise across these files.

src/views/Login/Shards.js
```diff
@@ -86,129 +36,58 @@ class Shards extends React.Component {
...
+  };
+
+  const shards = [shard1, shard2, shard3];
+  const ready = shards.filter(x => x !== '').length > 1;
+
+  return (
+    <View>
+      Enter your point and at least two of your three Urbit master ticket shards
+      here. The index of the input field needs to match the index of the shard.
+      <PointInput
```
shrugs: check out the PointInput + InputSigil from ViewPoint for how to integrate a validated sigil into here

src/views/Login/Ticket.js
```diff
@@ -29,174 +25,125 @@ const INPUT_STATUS = {
...
     if (!ob.isValidPatq(ticket) || !ob.isValidPatp(pointName)) {
-      this.props.setUrbitWallet(Nothing());
-      this.setState({ ticketStatus: Nothing() });
+      setTicketStatus(Nothing());
       return;
     }
-    this.setState({ ticketStatus: Just(INPUT_STATUS.SPIN) });
-    const contracts = need.contracts(this.props);
+    setTicketStatus(Just(INPUT_STATUS.SPIN));
+    const realContracts = need.contracts(contracts);
```
shrugs: likewise here, `_contracts` or whatever we figure out

src/views/Login/Trezor.js
```diff
@@ -1,146 +1,118 @@
...
 import Maybe from 'folktale/maybe';
-import { Button } from '../../components/old/Base';
-import { Row, Col, H1, P } from '../../components/old/Base';
+import TrezorConnect from 'trezor-connect';
+import * as secp256k1 from 'secp256k1';
+
+import View from '../../components/View';
+import { ForwardButton } from '../../components/Buttons';
 import {
+  H1,
```
shrugs: likewise here, H1 and P from `indigo-react`

src/views/Login/Trezor.js
```diff
@@ -1,146 +1,118 @@
...
     });
+  };
+
+  let accountOptions = [
+    {
+      title: 'Custom path',
+      value: 'custom',
+    },
+  ];
+  for (let i = 0; i < 20; i++) {
```
shrugs: looks like we can share this code with ledger a bit?

shrugs:
very pretty, love the `-400` lines

src/views/Login/Keystore.js
```diff
@@ -69,83 +63,65 @@ class Keystore extends React.Component {
...
+          {'Upload Keystore file'}
+        </div>
+      </UploadButton>
+
+      <PassphraseInput
+        name="password"
+        label="Keystore password"
+        initialValue={password}
+        onValue={setPassword}
+        autoFocus></PassphraseInput>
```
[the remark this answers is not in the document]
Fang-: Huh, no, it's done that autoatically for me before, so I didn't pay attention to it. Weird.

src/views/Login/Ledger.js
Fang-: Yeah I just carried them over. But you're right that leaving them out is probably better. At the very least, it's easier on the eyes.

src/views/Login/Shards.js
Fang-: Shouldn't that just be part of PointInput though?

src/views/Login.js
```diff
@@ -1,69 +1,65 @@
...
+    const realContracts = need.contracts(contracts);
 
     // if no point cursor set by login logic, try to deduce it
     let deduced = Nothing();
-    if (Nothing.hasInstance(this.props.pointCursor)) {
+    if (Nothing.hasInstance(pointCursor)) {
       const owned = await azimuth.azimuth.getOwnedPoints(
-        contracts,
-        wallet.address
+        realContracts,
```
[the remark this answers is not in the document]
Fang-: Very fair! I like the `_` prefix best. `$` is definitely way too weird.

src/views/Login/Shards.js
shrugs: @Fang- yeah, good idea!

src/views/Login/Trezor.js
Fang-: Eh, the Trezor one is subtly different from Ledger, in that it also has a "custom path" option prepended. Makes this kind of awkward, since JS doesn't let you do `[head-item, rest-of-the-array]`, does it?

shrugs: it does! what about some code like

```js
const buildHardwarePathOptions = (head) => [head, ...times(20, i => ({ ... })]
```

src/views/Login/Shards.js
Fang-: Hmm, I tried taking a stab at this, but couldn't get it to work. No console error, just didn't render... Also probably dealt with the error, pass and focussed props in `ViewPoint` in the wrong/inelegant way. Could we `useState` inside of `components/Inputs`, to factor that out?

Fang-:
Thanks for the feedback, I've pushed some fixes/cleanup!

src/views/Login/Shards.js
shrugs: hmm, I'll take a look at it once we get my `useForm` changes into mino

Fang-: nvm, I was overlooking something dumb!

This isn't as pretty everywhere as it is in `ViewPoint` because not everyone uses the `error` etc props yet. I'll try poking at it a bit, but take a look at the commit I just pushed and lmk your thoughts.

Fang-:
Gonna go ahead and merge this in, further `PointInput` work would be better at home in a separate PR anyway.

PR #176, opened 2019-06-25
refactor email list & invite flow logic

shrugs:
- [x] figure out why my invite txs are failing
- [x] fully test invite flow
- [x] need to put the accessory inside the input
- [ ] pull as much of the logic as I can out of this gnarly function & clean up
  - although the line count is basically identical to previous version, i'm not happy with the refactor yet
- [x] need to test with a real gas tank - haven't yet set that up on my computer


shrugs:
I just rebased against origin/mino so I have no idea why it says out of date with the base branch, maybe just needs to bust the cache somehow? previous commit in this branch is definitely the same as `mino` ¯\\\_(ツ)_/¯

src/views/Invite/InviteEmail.js
```diff
@@ -45,13 +51,90 @@ const EMAIL_STATUS = {
   FAIL: '×',
 };
 
+// world's simplest uid
+let id = 0;
+
+const buildInputConfig = (extra = {}) =>
+  buildEmailInputConfig({ name: (id++).toString(), ...extra });
```
Fang-: Probably want `'email' + (id++).toString()`?

src/views/Invite/InviteEmail.js
```diff
@@ -45,13 +51,90 @@ const EMAIL_STATUS = {
...
+  ] = useArray(
+    [buildInputConfig({ placeholder: 'Email Addresss' })],
+    buildInputConfig
+  );
+
+  const [hovered, _setHovered] = useState({});
+  const setHovered = (name, value) => () =>
+    _setHovered((value = { ...value, [name]: value }));
+
+  console.log(inputConfigs.length, inputConfigs.map(c => c.name));
```
Fang-: Do we want to keep this in for whatever reason?

src/views/Invite/InviteEmail.js
```diff
@@ -494,8 +524,8 @@ class InviteEmail extends React.Component {
...
-export default compose(
-  withNetwork,
-  withWallet,
-  withPointCursor
-)(InviteEmail);
+// export default compose(
+//   withNetwork,
+//   withWallet,
+//   withPointCursor
+// )(InviteEmail);
```
Fang-: Pretty sure this can just go?

src/views/Mnemonic.js
```diff
@@ -2,16 +2,26 @@ import React, { useCallback, useEffect, useState } from 'react';
...
+} from 'components/Inputs';
 import { ForwardButton } from 'components/Buttons';
 
 import { ROUTE_NAMES } from 'lib/routeNames';
 import { useHistory } from 'store/history';
 import { walletFromMnemonic } from 'lib/wallet';
 import { useWallet } from 'store/wallet';
+import { Input } from 'indigo-react';
 
 export default function Mnemonic() {
```
Fang-: Changes here should apply to `/views/Login/mnemonic` instead, I think. Or both, both is good too.

src/views/Network.js
```diff
@@ -8,6 +7,7 @@ import { ROUTE_NAMES } from '../lib/routeNames';
 import { useHistory } from '../store/history';
 import { useNetwork } from '../store/network';
 import View from 'components/View';
+import { ForwardButton } from 'components/Buttons';
 
 const kNetworkOptions = [
```
Fang-: Don't worry about touching up these "deprecated" screens too much.

src/lib/validators.js
```diff
@@ -3,6 +3,10 @@ import * as ob from 'urbit-ob';
 
 import { isValidAddress } from './wallet';
 
+// via: https://emailregex.com/
+const emailRegExp = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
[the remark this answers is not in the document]
shrugs: ah, _that's_ why the official W3C regex allows that! I thought it was a bug and went for this more restrictive version. All things being equal, I prefer the more restrictive requirement of a `sub.domain`, since 99% of `m@tld` will be a typo. That's more of a product decision, but we can just call it ourselves, imo. What do you think?

src/store/pointCache.js
```diff
@@ -19,6 +19,8 @@ const kEmptyInvites = {
   acceptedInvites: Maybe.Nothing(),
 };
 
+// TODO(shrugs): this hook is too big -> refactor into individual hooks for
+// details, birthday, invites, etc
```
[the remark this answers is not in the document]
shrugs: @Fang- ah, we can still do that — I meant putting the individual birthday/details/invites logic into their own hooks and then including those hooks within this `pointDetails` one. just composition of logic, not changing the outside API

src/store/pointCache.js
```diff
@@ -50,33 +56,31 @@ function _usePointCache() {
     [invitesCache]
   );
 
-  const fetchPoint = useCallback(
+  const syncDetails = useCallback(
```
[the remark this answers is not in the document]
shrugs: nothing strong: I had to name something else `sync*` and figured I'd standardize the verb for fetching from the blockchain and storing locally. ¯\\\_(ツ)_/¯

src/views/Invite/InviteEmail.js
shrugs: @Fang- not necessary, since ids only need to be valid within their own loop, unless I misinterpreted your reason

src/views/Network.js
shrugs: @Fang- yeah, will nuke those for #174 once that's merged today

src/lib/validators.js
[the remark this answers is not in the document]
shrugs: that said, we _are_ working with moderately technical users, though, so me@domain might actually happen here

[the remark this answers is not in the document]
shrugs: yeah, screw it, W3C here I come

src/views/Invite/InviteEmail.js
Fang-: My reasons are purely semantic. This is what ends up in the element's `name` attribute, right? It seems cleaner to have something meaningful in there, rather than just a number. But you're right that it won't, technically, matter.

src/lib/validators.js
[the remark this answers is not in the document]
Fang-: Eh, none of our users actually own a TLD, and I highly doubt anyone who does isn't used to needing to work around it. I'm fine requiring `@sub.domain`.

src/store/pointCache.js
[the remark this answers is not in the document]
Fang-: "sync" implies a two-way street, but it's fine as long as we standardize on that for "get from chain" terminology.

src/views/Invite/InviteEmail.js
shrugs: ah, totally - will add it

src/lib/useIsEclipticOwner.js
```diff
@@ -0,0 +1,20 @@
+import { useWallet } from 'store/wallet';
+import { usePointCache } from 'store/pointCache';
+
+import { eqAddr } from 'lib/wallet';
+
+export default function useIsEclipticOwner() {
```
Fang-: Do we really need a helper thing for this? It's only ever gonna be used in one place...

src/lib/useSyncPoints.js
```diff
@@ -0,0 +1,24 @@
+import { useEffect } from 'react';
+import { isEqual } from 'lodash';
+
+import { usePointCache } from 'store/pointCache';
+import usePreviousValue from 'lib/usePreviousValue';
+
+// sync a known point to the cache (i.e display values)
+export function useSyncKnownPoints(points = []) {
+  //
```
Fang-: TODO?

src/lib/useSyncPoints.js
```diff
@@ -0,0 +1,24 @@
...
+// sync an owned point to the cache
+export function useSyncOwnedPoints(points = []) {
+  const { syncOwnedPoint } = usePointCache();
+  const prevPoints = usePreviousValue(points);
+
+  useEffect(() => {
+    if (!isEqual(prevPoints, points)) {
+      // if the actual set of points changes, sync the whole array
+      // TODO: we should get clever about caching
+      points.forEach(point => syncOwnedPoint(point));
```
Fang-: Yeah this _looks_ like it re-builds the entire cache every time a new point is added? But maybe I misunderstand.

shrugs:
yeah, I haven't really figure out for myself where the line with hooks are. I feel like I'm finding it right now as I'm updating the email list stuff.

src/lib/useSyncPoints.js
shrugs: correct — to be fair, the old code was doing the same thing every time the points page was mounted. this code should only update the entire list of points when a new point is added. the clever code can be part of the actual `sync*` functions that decide whether or not to fetch from the chain

src/lib/useSyncPoints.js
shrugs: right now there are no values we need to display a point card beyond their actual point.

in the future we'll need the info to display "key type" in the list of points

src/lib/useIsEclipticOwner.js
shrugs: I guess not, but it also feels cleaner when it's nice and isolated like this

src/components/Blinky.js
```diff
@@ -0,0 +1,22 @@
...
+export const kLoadingCharacter = '▓';
+export const kInterstitialCharacter = '░';
+
+export const matchBlinky = obj =>
+  obj.matchWith({
+    Nothing: () => <Blinky />,
+    Just: p => p.value,
+  });
+
+export default function Blinky({
```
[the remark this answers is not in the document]
shrugs: imo things feel broken without the blink (loading can take 5+ seconds locally) so _some_ indication is super necessary imo

shrugs:
yeah, so hook refactor of this email component was definitely a hassle I shouldn't have taken on, but it's becoming cleaner, slowly.

use `REACT_APP_STUB_MAILER=true` to avoid making actual http calls to the mailer server (and print the generated ticket instead)

I still need to:
- [ ] figure out why my invite txs are failing — @Fang- any ideas? this is what i'd also emailed you about
  - the star set its spawn proxy and gave 5 invites to my planet, and now my planet is struggling to send transactions
- [ ] pull as much of the logic as I can out of this gnarly function & clean up
  - although the line count is basically identical to previous version, i'm not happy with the refactor yet
- [ ] need to test with a real gas tank - haven't yet set that up on my computer
- [ ] need to put the accessory inside the input

shrugs:
and I still don't know why this is out of date with the base branch. I'm going to try rebasing on github and see if that busts it

src/components/Blinky.js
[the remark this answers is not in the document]
shrugs: although after rebooting my ganache, load times are way faster, so idk what that was about.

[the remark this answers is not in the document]
Fang-: The search times for this on local nets scale linearly with chain size. On mainnet, we're behind an Infura node, which indexed event logs, making this result appear almost instantly. In most mainnet cases, it'll barely have time to blink once.

Fang-:
> figure out why my invite txs are failing — @Fang- any ideas?

How about now?

> and I still don't know why this is out of date with the base branch.

Maybe try merging instead of rebasing? Github is clearly too dumb to deal with it like you'd want it to.

src/components/Blinky.js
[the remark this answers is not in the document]
shrugs: ohhhhh that makes sense, yeah. hmm. in that case should we go back to the raw character without blinking, or keep blinking in the case that a query does actually take that long?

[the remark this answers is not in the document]
Fang-: I tried this out today, just the static "mask" isn't that bad. I'm fine with starting the blink after 2-3 idle seconds, sure.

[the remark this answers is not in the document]
shrugs: cool, I'll code that in 👍

shrugs:
marking it as ready for review got github to update the base!

shrugs:
@Fang- also, my txs are passing, so looking good

src/components/Blinky.js
```diff
@@ -17,7 +18,14 @@ export default function Blinky({
...
   const [value, setValue] = useState(true);
-  useInterval(() => setValue(val => !val), 1000);
+  const now = useRef(new Date());
+
+  useInterval(() => {
+    // only start blinking if we've elapsed enough time
+    if (new Date() - now.current > kBlinkAfterMs) {
+      setValue(val => !val);
+    }
+  }, 1000);
```
Fang-: Hmm, doesn't this only fire once a second, reducing the precision of `kBlinkAfterMs`?

shrugs: yup! figured precise timing wasn't a big deal here. with a blink after of `2500`, it should fire on the 3rd second.

shrugs:
just tested the full flow, getting into an authenticated state with my new invited ticket, so things are looking up! gonna move my todo list into the PR body and check things off as I wrap them up today. will merge this soon and then start a new PR that polishes UI for the entire flow

src/lib/txn.js
```diff
@@ -9,6 +9,9 @@ import { ledgerSignTransaction } from './ledger';
 import { trezorSignTransaction } from './trezor';
 import { WALLET_TYPES, addHexPrefix } from './wallet';
 
+const kAverageBlockTimeMs =
+  process.env.NODE_ENV === 'development' ? 1000 : 13000;
```
Fang-: I'd probably just check every five seconds in a live environment, to reduce response times?

shrugs: sgtm - was using the old 13000 magic number, but I like the idea of 5000

shrugs:
@Fang- ran into a hung up - in the app we receive strings as a response from `web3.eth.getBalance`, but then go compare them with numbers (`cost`, `minBalance`) all over the place. There are probably a _ton_ of bugs hiding here, but we're probably getting lucky because javascript is dumb and auto-parses the string as a base 10 number and does the comparison 'correctly'

anyway, the solution is to ship `BigNumber` or `BN` and make sure we're doing math with that lib. I'm going to ignore that as part of this PR, but it's something that absolutely needs to happen sooner rather than later, since it's begging to affect someone's on-chain ETH in unexpected ways.

Fang-:
> in the app we receive strings as a response from web3.eth.getBalance, but then go compare them with numbers (cost, minBalance) all over the place
> javascript is dumb and auto-parses the string as a base 10 number and does the comparison 'correctly'

Oof. Why is this allowed? ):

You're right that we should be using a bignum lib (probably whichever one Web3 also uses). I've made an Asana task for it!

shrugs:
@Fang- because javascript was designed in 14 days 20 years ago 😭

this is a _big_ reason why I'm such a fan of typescript - this code wouldn't have even compiled under typescript and we can catch entry level bugs like this one and the one fixed in https://github.com/urbit/bridge/pull/176/commits/df46115b705f2dbd2407af629da9aa925a39640c much easier

shrugs:
merged! last commit shouldn't have changed review status (just comments and cleaning up), but let me know if you find an issue with it @Fang- (no rush)

PR #178, opened 2019-06-27
invite flow ui

shrugs:
taking a crack at the Activate flow now

- Q: is `Activate` the homepage? or is it the Login page and we only jump into Activate when we load a `some-point.azimuth.network` domain?


Fang-:
> Q: is Activate the homepage? or is it the Login page and we only jump into Activate when we load a some-point.azimuth.network domain?

Pretty sure Activate (which we should name "ClaimInvite" or similar internally for clarity, it has overlap with Azimuth internal terminology otherwise) is where you end up when you click the "claim this invite" link from the email, or hop into that flow ("activate") from the default page. The default page should be login, and it can should do ship name deduction based on URL.

The point name doesn't matter for invite activation, since the invite ticket is just `[~zod ~the-ticket]`, not `[~the-planet ~the-ticket]`. We deduce the specific point we're dealing with by looking at the points which that ticket is the transfer proxy for.

src/lib/walletgen.js
```diff
@@ -66,9 +67,12 @@ export const generateWallet = async (point, ticket, boot) => {
 export const generateOwnershipWallet = (ship, ticket) =>
   kg.generateOwnershipWallet({ ship, ticket });
 
+// ~zod is a constant in temporary wallets used for invitations
```
Fang-:
```suggestion
// temporary wallets need to be derivable from just the ticket,
// so we always use ~zod as the wallet point.
```

shrugs:
@Fang- latest commit has the entire flow working on indigo. some styling tweaks needed on the passport screen, but should be good enough for our monday deadline. still need to implement retry/restart for claiming invite

src/views/Activate/ActivateCode.js
```diff
@@ -134,20 +134,17 @@ export default function ActivateCode() {
           as={ForwardButton}
           className="mt4"
           disabled={!pass || deriving}
-          accessory={deriving ? <Blinky /> : null}
+          accessory={deriving ? <Blinky /> : undefined}
```
Fang-: Does react care about this difference when it comes to rendering?

src/views/Activate/ActivateDisclaimer.js
```diff
@@ -1,12 +1,15 @@
...
+import cn from 'classnames';
+import { Grid, H3, B, Text, CheckboxInput } from 'indigo-react';
 
 import View from 'components/View';
 import { useCheckboxInput } from 'components/Inputs';
 import { ForwardButton } from 'components/Buttons';
 import { useLocalRouter } from 'lib/LocalRouter';
 import WarningBox from 'components/WarningBox';
 
+const kTextStyle = 'f5';
```
Fang-: Still a -1 from me on this naming style for constants. I wasn't going to say anything, but you used `BIG_CASE` in your change to PassportVerify, so... (;

src/views/Activate/ActivateCode.js
shrugs: react doesn't, but javascript does — in the following syntax

```
<Component accessory="default" {...props} />
```

passing `undefined` gives the component "default" and passing `null` gives the component `null`

(the idea being that `undefined` is specifically the case where "nothing at all" was passed

src/views/Activate/ActivateDisclaimer.js
shrugs: the pattern I'm going with is that config-level stuff is `CONSTANT_CASE` and local constants that aren't designed to be flags/config are `kConstant` — thoughts?

shrugs:
ok, finally ready to merge this big boi

src/views/Activate/ActivateDisclaimer.js
Fang-: I would personally forego special naming entirely in that case, and just group them directly under the imports consistently.

Naming prefixes just look ugly to me when you're not going 110% Hungarian notation.

shrugs: fixed in the latest

src/views/Activate/ActivatePassport.js
```diff
@@ -44,6 +50,21 @@ export default function ActivatePassport() {
   const leftHalf = useBreakpoints([false, false, 1]);
   const rightHalf = useBreakpoints([false, false, 2]);
 
+  // addresses are not derived until we set[Urbit]Wallet(), so do that inline
```
Fang-: Pretty sure you can just say `wallet.ownership.keys.address` though, since it's an Urbit Wallet?

shrugs:
merging because of the previous approval & assuming you're asleep 👍

src/views/Activate/ActivateCode.js
```diff
@@ -0,0 +1,150 @@
...
+      const _contracts = need.contracts(contracts);
+
+      setDeriving(true);
+      // when the ticket becomes valid, derive the point
+      (async () => {
+        const { seed } = await generateTemporaryOwnershipWallet(
+          ticketInput.data
+        );
+
+        //TODO isn't all this accessible in the ownership object?
```
Fang-: Yes! You can just `{ keys }` above. [ref](https://github.com/urbit/urbit-key-generation/blob/master/src/index.js#L172-L175) (This reads like one of my own TODOs though. (^: )

src/views/Activate/ActivateCode.js
```diff
@@ -0,0 +1,150 @@
...
+        } else {
+          setGeneralError(
+            'Invite code has no claimable point.\n' +
+              'Check your invite code and try again?'
+          );
+        }
+
+        setDerivedPoint(realPoint);
+        setDerivedWallet(wallet);
+        setInviteWallet(inviteWallet);
```
Fang-: This is a bit confusing. `wallet` wasn't derived, but randomly generated just now! Putting it into `derivedWallet` (even just having _both_ `derived` and `invite` wallets) seems not correct... Am I missing something?

Fang-:
> & assuming you're asleep 👍

Now _this_ I take issue with! What do you take me for, a weakling?

PR #180, opened 2019-07-02
refactor: indigo-ifying login components

shrugs:
- [x] convert all the login components to indigo stuff
  - [ ] pending frozen spec from jimmy
- [x] implement a basic dropdown
  - [ ] accessibility pass
- [x] clean up the `useForm`/`use*Input` APIs - they're looking mighty pretty right now, afaik
- [x] merge shards and ticket inputs
- [x] manually test all of these flows

@Fang- can you manually test the ledger flow? again, logic shouldn't have changed but ¯\\\_(ツ)_/¯


src/views/Login.js
```diff
@@ -1,104 +1,145 @@
...
+  }, [goToPoints, push, names]);
+
+  const doContinue = useCallback(async () => {
     const _wallet = need.wallet(wallet);
     const _contracts = need.contracts(contracts);
 
     // if no point cursor set by login logic, try to deduce it
     let deduced = pointCursor;
-    if (Nothing.hasInstance(deduced)) {
+    if (Maybe.Nothing.hasInstance(deduced)) {
```
Fang-: I think @jtobin and I decided some time ago to just consistently use them outside of their `Maybe.` namespace, and I believe all newer code adheres to that?

src/views/Login/Ledger.js
```diff
@@ -1,190 +1,199 @@
...
+  );
 
   return (
-    <View>
-      <H1>Authenticate With Your Ledger</H1>
-      {body}
-    </View>
+    <Grid className={className}>
+      {isHTTPS && renderHTTPS()}
+      {!isHTTPS && renderHTTP()}
```
Fang-: What about `{isHTTPS ? renderHTTPS() : renderHTTP()}`?

shrugs:
> This is kind of awkward. Is there any way we can do the on-unfocus logic _after_ handling the click? Clicks should always do exactly what they look like they'll do (unless it hasn't looked that way for <300ms, but Bridge doesn't do surprise popups etc, so all good).

yeah this was pissing me off during dev as well. what's the right behavior here? the thought I had was to only show an error if the input has been "touched" before (aka, they've focused the form _and_ changed the value). Not sure how that matches with jimmy's instincts, but I think I'll implement that because the reflow is agonizing.

Another alternative is to always include padding where the error text would be so that no reflow actually occurs when the error shows, but that would mess with the layouts (figma frames don't account for this). Will ask Jimmy in the sync tomorrow.

src/components/Inputs.js
```diff
@@ -35,10 +41,41 @@ import { AccessoryIcon } from 'indigo-react';
 //   validators: [validateEthereumAddress, validateNotEmpty],
 // });
 
-const firstOf = state => state.inputs[0];
+// const GalaxyInput = advancedInput({
+//   WrappedComponent: Input,
+//   validators: [validateGalaxy, validateNotEmpty],
+//   transformers: [prependSig],
+// });
```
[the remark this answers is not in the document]
shrugs: yeah definitely — will eventually just be `usePointInput` with an additional validator on the size of the point.

src/components/Inputs.js
```diff
@@ -102,17 +154,24 @@ export function useTicketInput({ validators = [], ...rest }) {
...
-  return { ...input, accessory: input.pass ? <AccessoryIcon.Success /> : null };
 }
 
 const kPointValidators = [validatePoint, validateNotEmpty];
 const kPointTransformers = [prependSig];
 export function usePointInput(rest) {
   const [lastValidPoint, setLastValidPoint] = useState('');
-  const input = firstOf(
+  return useFirstOf(
     useForm([
```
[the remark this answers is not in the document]
shrugs: sure!

useForm returns a _list_ of inputs (which is required to handle dynamic forms like the InviteEmail one). In the case where we just want a single input we still use the logic from useForm but make the output much easier to consume for the components that just have the single input to care about.

truthfully we could probably convert all of the components with multiple inputs on them into a single `useForm()` state, but unless there are dynamic fields or more complex interactions between individual inputs the effort won't be worth it.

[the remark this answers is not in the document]
shrugs: firstOf was originally just a simple function that pulled the first input from the useForm (`useForm().inputs[0]`) but now that I've added the `setValue` callback I converted it to a hook so that we could use useCallback, etc.

src/indigo-react/components/Button.js
```diff
@@ -23,7 +24,7 @@ export default function Button({
   };
   return (
     <Grid
-      as="a"
+      as={As}
```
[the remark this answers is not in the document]
shrugs: this lets us change the actual component that the Button (and Grid) renders as. Normally a button is an `a` tag (see the default value `as: As = 'a'`) but we could change it to something else if it's important for the markup (like `<Button as={MyStyledButton} />` or `<Button as="button" />`

src/views/Activate/ActivateCode.js
```diff
@@ -36,12 +36,11 @@ export default function ActivateCode() {
     setDerivedPoint,
   } = useActivateFlow();
 
-  const ticketInput = useTicketInput({
+  const [ticketInput, { pass: validTicket }] = useTicketInput({
```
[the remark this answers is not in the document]
shrugs: it is pretty standard for hooks, since we frequently need to rename generic arguments to contextual ones (like `const [myState, setMyState] = useState()`)

in this case I'm just making it easier/more semantic to do the "create the input state and also give me a local reference to some of it" pattern that we were doing in every instance of an input. the destructuring & renaming syntax can be verbose, but I don't know that there's a more useful alternative

src/views/Login.js
shrugs: oh no, I've been doing literally the opposite this whole time 😓

I'll probably have to do a pass against `Maybe.` and convert all of those inputs across the project

Fang-:
> the thought I had was to only show an error if the input has been "touched" before

Agreed that seems sane. Padding would just get ugly.

Heads up, Jimmy's out, and the 4th is a company holiday, so I don't think there's a sync.

shrugs:
> Heads up, Jimmy's out, and the 4th is a company holiday, so I don't think there's a sync.

thank god, I slept past the meeting anyway 😂

will implement `touched` behavior and see if anyone notices

Fang-:
This keeps you from logging in when "ticket is not owner or transfer proxy for this point", but it shouldn't. Displaying the warning is fine, but you should still be allowed to log in. You may own other points.

Also, that error message doesn't appear until you un-focus the input field. I feel like `touched` behavior should only apply to "this is required" errors?

shrugs:
> This keeps you from logging in when "ticket is not owner or transfer proxy for this point", but it shouldn't. Displaying the warning is fine, but you should still be allowed to log in. You may own other points.

that was also the behavior before. good catch tho

> Also, that error message doesn't appear until you un-focus the input field. I feel like `touched` behavior should only apply to "this is required" errors?

I wasn't able to replicate this - the error shows up as soon as it finishes trying to derive the wallet (using zod and the placeholder ticket as an example), before I've unfocused the field. can you go into more detail?

Fang-:
> I wasn't able to replicate this - the error shows up as soon as it finishes trying to derive the wallet 

Hmm, you're right. Seems to have been an issue on my end. It works properly now!

src/components/Inputs.js
```diff
@@ -137,16 +139,20 @@ export function useHdPathInput(props) {
   );
 }
 
-const kTicketValidators = [validateTicket, validateNotEmpty];
-const kTicketTransformers = [prependSig];
+const kTicketValidators = [
+  validateTicket,
+  validatePatpByteLength(8), // tickets are 8 bytes of patp
```
Fang-: This isn't strictly true. In development we may user shorter tickets, in production stars and galaxies usually have 16- and 32-byte tickets respectively.  
(Also tickets are patq, not patp, but more or less the same thing.)

src/lib/lib.js
```diff
@@ -34,6 +34,11 @@ const randomPatq = len => {
   return ob.hex2patq(hex);
 };
 
+// TODO: is there a function for this calculation in urbit-ob or similar?
```
Fang-: There's `patp2syls`, but it isn't exported and doesn't do _exactly_ what you want... ):

src/components/Inputs.js
shrugs: we probably want to allow 32 byte tickets, so `validateMaximumPatpByteLength(32)`?

I'll rename it to `validatePatpqByteLength` as well, since I guess it'll work for both encodings?

Fang-: We should just allow any valid patq, regardless of length. I wrote my bank an angry letter once about their needless limit on password length, Bridge definitely shouldn't make the same mistake. (^:

src/components/Inputs.js
```diff
@@ -35,10 +42,41 @@ import { AccessoryIcon } from 'indigo-react';
 //   validators: [validateEthereumAddress, validateNotEmpty],
 // });
 
-const firstOf = state => state.inputs[0];
+// const GalaxyInput = advancedInput({
+//   WrappedComponent: Input,
+//   validators: [validateGalaxy, validateNotEmpty],
+//   transformers: [prependSig],
+// });
```
Fang-: Do we care about keeping this commented code in?

src/views/Activate/ActivatePassport.js
```diff
@@ -68,7 +67,7 @@ export default function ActivatePassport() {
     <LocalRouterProvider value={router}>
       <View full={fullView}>
         <Grid gap={gap} className="mt8 mb10">
-          <Grid.Item half={leftHalf} full={full}>
+          <Grid.Item half={half && 1} full={full}>
```
Fang-: This seems like a downgrade, readability-wise, from the earlier version.

src/components/Inputs.js
shrugs: ¯\\\_(ツ)_/¯ guess not, will remove and rename this file for clarity

src/views/Activate/ActivatePassport.js
shrugs: I actually like it, since I can look at the components here and know that this specific Grid.Item will be the left side and the other will be the right, without having to go check what the values of `leftHalf` and `rightHalf` will be.

PR #181, opened 2019-07-02
feat: upgrade to latest hotreload api and add better hook support

shrugs:
title explains: following the new api from https://github.com/cdharris/react-app-rewire-hot-loader and https://github.com/gaearon/react-hot-loader to get better hooks support - looking 👌 to me


shrugs:
well, all this does is remove the console.warn that mentions it... so 10/10 for effectiveness

PR #182, opened 2019-07-02
Admin action pages + reticketing

Fang-:
This adds most of the logic for the admin action overview pages and links them through to their existing implementations. Those still need (more than) a little styling work, but are functionally all there? Seems like it's time to figure out what to do with StatelessTransaction...

Also implements reticketing, which uses `/lib/invite` under the hood and makes some changes to that as previously discussed.

Beware, Jank JS Inside, as usual!


src/lib/invite.js
```diff
@@ -38,22 +38,14 @@ const TRANSACTION_STATES = {
...
   },
-  FUNDING_INVITE: {
-    label: 'Funding Invite Wallet...',
+  FUNDING: {
+    label: 'Funding Transactions...',
     progress: 0.3,
   },
   CLAIMING: {
     label: 'Claiming Invite...',
     progress: 0.55,
```
shrugs: perhaps we should update the progress percent here, since there are fewer steps?

src/lib/invite.js
```diff
@@ -120,62 +113,28 @@ async function claimPointFromInvite({
...
   //
-  const newAddress = wallet.ownership.keys.address;
 
-  // configure management proxy
-  const managementTx = azimuth.ecliptic.setManagementProxy(
+  progress(TRANSACTION_STATES.GENERATING);
+
+  // transfer to invite wallet, so it can configure
+  //NOTE doing this in the invite wallet instead of the final wallet makes
+  //     failure cases "solvable" by just telling the user to fully retry
```
shrugs: add note: "i.e. the invite flow is effectively idempotent"

src/lib/invite.js
```diff
@@ -120,62 +113,28 @@ async function claimPointFromInvite({
...
+
+  //TODO no harm done if we already owned it, but should still get a bool arg
+  //     for skipping this, if it isn't too big a burden for callers
+  let transferTmpTx = azimuth.ecliptic.transferPoint(
```
shrugs: `const transferTmpTx`?

src/lib/invite.js
```diff
@@ -120,62 +113,28 @@ async function claimPointFromInvite({
...
     contracts,
     point,
-    wallet.management.keys.address
+    inviteAddress,
+    false
   );
-  managementTx.gas = 200000;
-  managementTx.nonce = 0;
+  transferTmpTx.gas = 500000; //TODO can maybe be lower?
```
shrugs: can we make this `toWei('500', 'gwei')` or whatever the more semantic value is? the magic numbers for gas will definitely confuse readers in the future.

shrugs: (let's just update the magic gas numbers in this file only, for now?) or at least make them constants like `TRANSFER_TX_GAS` or something.

src/lib/invite.js
```diff
@@ -202,41 +161,109 @@ async function claimPointFromInvite({
...
+    point,
+    wallet.management.keys.address
+  );
+  managementTx.gas = 200000;
+
+  // set spawn & voting proxies situationally
+
+  let txs = [transferTmpTx, keysTx, managementTx];
+  if (
+    azimuth.azimuth.getPointSize(point) !== azimuth.azimuth.PointSize.Planet
```
shrugs: `azimuth.azimuth.getPointSize(point)` will return a Promise (right?) which means we should probably `await` it here or we'll always get `false`

src/lib/invite.js
```diff
@@ -202,41 +161,109 @@ async function claimPointFromInvite({
...
+    let spawnTx = azimuth.ecliptic.setSpawnProxy(
+      contracts,
+      point,
+      wallet.spawn.keys.address
+    );
+    spawnTx.gas = 200000;
+    txs.push(spawnTx);
+
+    if (
+      azimuth.azimuth.getPointSize(point) === azimuth.azimuth.PointSize.Galaxy
```
shrugs: likewise here — is the return value of `getPointSize` not a promise?

src/lib/invite.js
```diff
@@ -202,41 +161,109 @@ async function claimPointFromInvite({
...
     totalCost = totalCost + tx.gasPrice * tx.gas;
     return tx;
   });
 
   let txPairs = txs.map(tx => {
     let stx = new Tx(tx);
-    stx.sign(Buffer.from(wallet.ownership.keys.private, 'hex'));
+    stx.sign(Buffer.from(inviteWallet.privateKey, 'hex'));
     return {
       raw: '0x' + stx.serialize().toString('hex'),
```
shrugs: let's use `hexify(stx.serialize())` instead of the manual hex string formation

shrugs: we do this all over the place, but let's at least update this file

src/lib/invite.js
```diff
@@ -202,41 +161,109 @@ async function claimPointFromInvite({
...
+  //
+  // sending and awaiting transactions
+  //
 
-  progress(TRANSACTION_STATES.CONFIGURING);
+  progress(TRANSACTION_STATES.CLAIMING);
 
+  console.log('sending txpairs', txPairs.length);
   await sendAndAwaitConfirm(web3, txPairs.map(p => Just(p.signed)), usedTank);
+  console.log('all done');
```
shrugs: probably worth removing this console.log

src/lib/invite.js
```diff
@@ -108,6 +100,7 @@ async function claimPointFromInvite({
   web3,
   contracts,
   onUpdate,
+  transferEth,
```
shrugs: let's default this to false (assuming that not transferring eth is the reasonable default)

src/views/Admin.js
```diff
@@ -0,0 +1,83 @@
...
+
+export default function Admin() {
+  const history = useHistory();
+  const { urbitWallet, wallet } = useWallet();
+  const { pointCursor } = usePointCursor();
+  const { pointCache } = usePointCache();
+
+  const point = need.point(pointCursor);
+  const pointDetails = need.fromPointCache(pointCache, point);
+  const userAddress = need.wallet(wallet).address;
```
shrugs: `const address = need.addressFromWallet(wallet)`

src/views/Admin.js
```diff
@@ -0,0 +1,83 @@
...
+        <Grid.Item full>
+          <ForwardButton
+            disabled={!canDownloadPassport}
+            onClick={goRedownload}
+            detail="Re-download your paper wallet">
+            Download Passport
+          </ForwardButton>
+        </Grid.Item>
+        <Grid.Divider />
+        <Grid.Item full>
```
shrugs: `{isOwner && ...}`?

src/views/Admin.js
```diff
@@ -0,0 +1,83 @@
...
+    history,
+  ]);
+
+  const canDownloadPassport = Just.hasInstance(urbitWallet);
+  const isOwner = eqAddr(userAddress, pointDetails.owner);
+
+  return (
+    <View>
+      <Grid className="pt3">
+        <Grid.Item full>
```
shrugs: For these, I would recommend a syntax like

```jsx
<Grid.Item full as={ForwardButton} disabled={!canDownloadPassport} onClick={...} {...theRestOfTheButtonProps} />
```

this way the button itself becomes the grid item (using the grid layout classes), which makes it automatically stretch the width of the grid and respect the other grid-specific layouts. buttons _do_ automatically stretch, but other elements might not, so I've standardized on making sure that all of the actual elements we care about behave as grid or flex items, which removes any uncertainty around how they'll be laid out.

src/views/Permissions.js
```diff
@@ -0,0 +1,167 @@
...
+
+import useLifecycle from 'lib/useLifecycle';
+import { useNetwork } from 'store/network';
+import { useHistory } from 'store/history';
+import { useWallet } from 'store/wallet';
+import { usePointCursor } from 'store/pointCursor';
+import { usePointCache } from 'store/pointCache';
+
+//TODO consolidate with azimuth-js' getActivationBlock
+const getRekeyDate = async (web3, contracts, point) => {
```
shrugs: should this belong as part of the point details store?

src/views/Permissions.js
```diff
@@ -0,0 +1,167 @@
...
+  ]);
+
+  const pointDetails = need.fromPointCache(pointCache, point);
+
+  const isOwner = eqAddr(userAddress, pointDetails.owner);
+  const canManage =
+    isOwner || eqAddr(userAddress, pointDetails.managementProxy);
+
+  const proxyAction = (name, address, onClick) => {
+    if (address === '0x0000000000000000000000000000000000000000')
```
shrugs: we have a constant for that `0x0....` address in `ETH_ZERO_ADDR` and should probably make a `isZeroAddress` helper function

src/views/Point.js
```diff
@@ -44,7 +46,7 @@ export default function Point() {
       <Passport point={Maybe.Just(point)} />
       <Grid className="pt2">
         <Grid.Item full>
-          <ForwardButton disabled>Admin</ForwardButton>
+          <ForwardButton onClick={goAdmin}>Admin</ForwardButton>
```
shrugs: only allow going to the Admin page if the pointCache has been set, otherwise the Admin page will throw because of the `need` requirements.

src/views/Admin.js
```diff
@@ -0,0 +1,83 @@
...
+  const goTransfer = useCallback(() => history.push(ROUTE_NAMES.TRANSFER), [
+    history,
+  ]);
+
+  const canDownloadPassport = Just.hasInstance(urbitWallet);
+  const isOwner = eqAddr(userAddress, pointDetails.owner);
+
+  return (
+    <View>
+      <Grid className="pt3">
```
shrugs: note to self: add crumbs, header during indigo-ification

src/views/Permissions.js
```diff
@@ -0,0 +1,167 @@
...
+          rekeyDate.value
+        ) : (
+          <Blinky a={LOADING_CHARACTER} b={INTERSTITIAL_CHARACTER} delayed />
+        )}
+      </>
+    );
+
+  return (
+    <View>
+      <Grid className={`pt${items}`}>
```
shrugs: why the dynamic padding here, and why is it based on the number of items?

src/views/Permissions.js
```diff
@@ -0,0 +1,167 @@
...
+    );
+  }
+
+  const netKeysText =
+    pointDetails.keyRevisionNumber === 0 ? (
+      'Not yet configured'
+    ) : (
+      <>
+        {`Revision #${pointDetails.keyRevisionNumber}, since `}
+        {Just.hasInstance(rekeyDate) ? (
```
shrugs: a good usecase for `rekeyDate.getOrElse(<Blinky ... />)` or even better `matchBlinky(rekeyDate)` which does exactly what you want

src/views/Permissions.js
```diff
@@ -0,0 +1,167 @@
...
+          <Blinky a={LOADING_CHARACTER} b={INTERSTITIAL_CHARACTER} delayed />
+        )}
+      </>
+    );
+
+  return (
+    <View>
+      <Grid className={`pt${items}`}>
+        <Grid.Item full>
+          <ForwardButton disabled="true" detail={pointDetails.owner}>
```
shrugs: just `disabled` is correct — we want a boolean, not string

src/views/Redownload.js
```diff
@@ -0,0 +1,49 @@
...
+  const [paper, setPaper] = useState(Nothing());
+
+  const doDownload = () => {
+    downloadWallet(paper.value);
+  };
+
+  const paperRenderer = Just.hasInstance(paper) ? null : (
+    <PaperCollateralRenderer
+      wallet={{ [point]: urbitWallet.value }}
+      className={'extremely-hidden'}
```
shrugs: `super-hidden` is the new class name and let's avoid extra interpolation with `className='super-hidden'`

src/views/Redownload.js
```diff
@@ -0,0 +1,49 @@
...
+  const { pointCursor } = usePointCursor();
+  const point = need.point(pointCursor);
+
+  const [paper, setPaper] = useState(Nothing());
+
+  const doDownload = () => {
+    downloadWallet(paper.value);
+  };
+
+  const paperRenderer = Just.hasInstance(paper) ? null : (
```
shrugs: good use case for `paper.matchWith`

src/views/Redownload.js
```diff
@@ -0,0 +1,49 @@
...
+      mode={'REGISTRATION'}
+    />
+  );
+
+  return (
+    <View>
+      <DownloadButton
+        disabled={Nothing.hasInstance(paper)}
+        onClick={doDownload}>
+        {Nothing.hasInstance(paper)
```
shrugs: good usecase for `paper.matchWith`

src/views/Reticket.js
```diff
@@ -0,0 +1,99 @@
...
+import useRouter from 'lib/useRouter';
+
+import { useHistory } from 'store/history';
+
+import Confirm from './Reticket/Confirm';
+import Download from './Reticket/Download';
+import Verify from './Reticket/Verify';
+import DoReticket from './Reticket/DoReticket';
+
+const STEP_NAMES = {
```
shrugs: let's just call them `NAMES` and `VIEWS` since they're file-local, which is what I've done with the other nested routers

src/views/Reticket.js
```diff
@@ -0,0 +1,99 @@
...
+
+  return (
+    <LocalRouterProvider value={innerRouter}>
+      <View>
+        <Crumbs
+          routes={[
+            {
+              text: name,
+              action: () => {
+                history.pop();
```
shrugs: you can `return history.pop(2)` here

src/views/Reticket.js
```diff
@@ -0,0 +1,99 @@
...
+            {
+              text: 'Admin',
+              action: () => history.pop(),
+            },
+          ]}
+        />
+        <MiniBackButton onClick={innerRouter.pop} />
+        Step {stepNumber} of {STEP_MAX}
+        <Route
+          STEP_NAMES={STEP_NAMES}
```
shrugs: no need to pass STEP_NAMEs — you can access them in children with `const { names } = useLocalRouter()`

src/views/Reticket/Confirm.js
```diff
@@ -0,0 +1,63 @@
...
+import { ForwardButton } from 'components/Buttons';
+
+import useLifecycle from 'lib/useLifecycle';
+import { generateWallet } from 'lib/invite';
+import { useLocalRouter } from 'lib/LocalRouter';
+
+import { isDevelopment } from 'lib/flags';
+
+export default function Confirm({ STEP_NAMES, newWallet, storeNewWallet }) {
+  const { push } = useLocalRouter();
```
shrugs: here's where you'd get `names` and do something like 

```
const goNext = useCallback(() => push(names.VIEW_NAME, [push, names]);
```

src/views/Reticket/Confirm.js
```diff
@@ -0,0 +1,63 @@
...
+  useLifecycle(() => {
+    generateWallet(point).then(wal => {
+      if (isDevelopment) {
+        console.log('ticket for', point, wal.ticket);
+      }
+      setGeneratedWallet(Just(wal));
+    });
+  });
+
+  const next = () => {
```
shrugs: let's always `useCallback` our route handlers for consistency

src/views/Reticket/Confirm.js
```diff
@@ -0,0 +1,63 @@
...
+    push(STEP_NAMES.DOWNLOAD);
+  };
+
+  const paperRenderer = Nothing.hasInstance(generatedWallet) ? null : (
```
shrugs: `matchWith`

src/views/Reticket/Confirm.js
```diff
@@ -0,0 +1,63 @@
...
+    <PaperCollateralRenderer
+      wallet={{ [point]: generatedWallet.value }}
+      className={'extremely-hidden'}
```
shrugs: same notes from paper collateral renderer above

shrugs: actually, can we abstract this even more somehow, since these code blocks are basically identical?

src/views/Reticket/DoReticket.js
```diff
@@ -0,0 +1,129 @@
...
+  const [generalError, setGeneralError] = useState();
+  const [{ label, progress }, setState] = useState({
+    label: 'Starting...',
+    progress: 0,
+  });
+  const [needFunds, setNeedFunds] = useState();
+
+  // start reticketing transactions on mount
+  useLifecycle(() => {
+    claimPointFromInvite({
```
shrugs: would prefer an anonymous async-await here and above, but it's not a big deal

```
(async () => {
  try {
      await claimPointFromInvite(...)
      setUrbitWallet(...)
  } catch (error) {
    setGeneralError(error)
  }
})()
```

src/views/Reticket/DoReticket.js
```diff
@@ -0,0 +1,129 @@
...
+        case 'gotFunding':
+          return setNeedFunds(false);
+        default:
+          console.error(`Unknown update: ${type}`);
+      }
+    },
+    [setState, setNeedFunds]
+  );
+
+  const next = () => {
```
shrugs: if the `next` fn is just `completed` we can just shove `completed` into whatever components expect it. especially since there's no `useCallback` here, they're semantically identical without the extra function creation

src/views/Reticket/DoReticket.js
```diff
@@ -0,0 +1,129 @@
...
+    completed();
+  };
+
+  const renderAdditionalInfo = () => {
```
shrugs: yeah we definitely want to de-dupe this somehow. not necessary for this review, though

src/views/Reticket/Verify.js
```diff
@@ -0,0 +1,40 @@
...
+  const displayTicketStatus = matches ? (
+    <AccessoryIcon.Success />
+  ) : (
+    <AccessoryIcon.Failure />
+  );
+
+  return (
+    <View>
+      Prove that you downloaded the new secrets!
+      <Input {...ticketInput} accessory={displayTicketStatus} />
```
shrugs: use the same pattern as the `PassportVerify` component for adding a custom validator (which will then automatically set `accessory` for you)

src/lib/invite.js
Fang-: I did actually change these values around a bit to try and fat-finger estimate the time they would take, but ended up at the same values as they initially had. Reasoning here for funding taking 25% while claiming takes 40% is that funding is _at most_ a small ETH transfer transaction with high gas price, so should be included fairly quickly. The same cannot be said for multiple larger (but still high gas price) transactions.

src/lib/invite.js
Fang-: Nope, [this is just a local mirror of the on-chain logic](https://github.com/urbit/azimuth-js/blob/master/azimuth.js#L333).

src/views/Point.js
Fang-: Oh, that commit's sitting unpushed, haha.

src/views/Permissions.js
Fang-: Because I have no idea what I'm doing. (^:

I had seen a couple Grids with `pt` numbers equal to their item count, and figured there was a relation there. There's... not, hah.

src/lib/invite.js
Fang-: But we modify its `.gas` property directly below?

src/lib/invite.js
Fang-: These aren't prices, but limits. Gwei doesn't exist in this context. Strong agree on the constants though. I should probably do some transaction analysis to actually find these values.

src/lib/invite.js
shrugs: sgtm!

src/lib/invite.js
shrugs: oh shit yeah totally 😅

src/views/Permissions.js
shrugs: what was the prior art you were looking at, for context?

src/lib/invite.js
shrugs: that's ok — the reference to `transferTmpTx` is still constant - you're not reassigning it or anything.

src/views/Permissions.js
Fang-: Might've been Point? Unsure...

shrugs: ¯\\\_(ツ)_/¯

Fang-:
Thanks for the feedback @shrugs. Just pushed fixes for everything I've resolved so far. Hope to get to the rest tonight, but highly doubt it. Soon™ will do!

src/components/PaperRenderer.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import PaperCollateralRenderer from 'PaperCollateralRenderer';
+
+export default function PaperRenderer({ point, wallet, callback, ...props }) {
```
shrugs: no need to spread `props` unless we want to pass them to `PaperCollateralRenderer` (although I think those 4 are the only props it supports?)

Fang-: whoops yes I was going to have this pass `...props` to `PaperCollateralRenderer`. Seems good practice?

shrugs: sure thing - no reason not to here ¯\\\_(ツ)_/¯

Fang-:
Gonna go ahead and merge this!

PR #184, opened 2019-07-05
Add tool for finding real-world gas usage

Fang-:
Also updates the constants to reflect results gained from it.

It ain't the prettiest, but it works. It filters for sane and successful transactions. Goes off events on the Azimuth contract initially because not all calls were made against the current instance of the Ecliptic contract (we had subtly different versions during launch ceremony).

In most cases, I just rounded up whatever was the highest gas usage that rolled out of this. In cases where we have low sample size, I rounded up a bit more aggressively.

The data you get from this is pretty interesting. Transactions tend to cluster around certain gas amounts, but there's almost always outliers on either side. Wonder why this is the way it is.

image

### DIFF
--- a/find-gas-usage.js
+++ b/find-gas-usage.js
@@ -1,0 +1,121 @@
+// usage: tweak constants at head of file, then run "node find-gas-usage.js"
+//        for very common events/calls, you may need to use a smaller TO-FROM
+//        range to keep web3 from yelling at you
+
+const ORIGIN = 6784800; // launch ceremony begins
+const PUBLIC = 7033765; // azimuth becomes public
+const LATEST = 'latest';
+
+const FROM = ORIGIN;
+const TO = LATEST;
+
+const VENT = 'ChangedKeys';
+const FUNC = '0x4447e48c'; // configureKeys
+const ARGS = 5;
+
+// const VENT = 'Spawned';
+// const FUNC = '0xa0d3253f'; // spawn
+// const ARGS = 2;
+
+// const VENT = 'OwnerChanged';
+// const FUNC = '0x1e79a85b'; // transferPoint
+// const ARGS = 3;
+
+// const VENT = 'ChangedSpawnProxy';
+// const FUNC = '0xae326221'; // setSpawnProxy
+// const ARGS = 2;
+
+// const VENT = 'ChangedManagementProxy';
+// const FUNC = '0x8866bb2c'; // setManagementProxy
+// const ARGS = 2;
+
+// const VENT = 'ChangedTransferProxy';
+// const FUNC = '0x2c7ba564'; // setTransferProxy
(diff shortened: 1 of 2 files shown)


find-gas-usage.js
```diff
@@ -0,0 +1,121 @@
...
+
+const LENT = 64 * ARGS + FUNC.length;
+
+const Web3 = require('web3');
+const azimuth = require('azimuth-js');
+
+(async () => {
+  const web3 = new Web3(
+    new Web3.providers.HttpProvider(
+      'https://mainnet.infura.io/v3/196a7f37c7d54211b4a07904ec73ad87'
```
shrugs: this API key needs to be an env variable, and we should probably cycle it now that it's been public on github

find-gas-usage.js
```diff
@@ -0,0 +1,121 @@
...
+    )
+  );
+  const contracts = await azimuth.initContractsPartial(
+    web3,
+    '0x223c067f8cf28ae173ee5cafea60ca44c335fecb'
```
shrugs: no idea what this address is - make it a named constant at the top?

find-gas-usage.js
```diff
@@ -0,0 +1,121 @@
...
+
+  const keyEvents = await contracts.azimuth.getPastEvents(VENT, {
+    fromBlock: FROM,
+    toBlock: TO,
+  });
+
+  const txsBatch = new web3.BatchRequest();
+  keyEvents.map(vent => {
+    txsBatch.add(
+      web3.eth.getTransaction.request(vent.transactionHash, (err, res) => {})
```
shrugs: is the empty callback block needed here and in the other `.request` calls?

find-gas-usage.js
```diff
@@ -0,0 +1,121 @@
...
+  });
+  const recs = await receiptsBatch.execute();
+
+  let usage = {};
+  recs.response.map(rec => {
+    if (rec.status !== true) {
+      console.log('rejecting', rec.gasUsed);
+      return;
+    }
+    if (rec.gasUsed > 480000) {
```
shrugs: let's move this 480000 magic number into a constant so we know what its purpose is

find-gas-usage.js
Fang-: Yeah, it falls over (with "is not a function" error) if you don't include a callback here.

find-gas-usage.js
Fang-: It's not a secret, we include it in the urbit source code also. It's just an endpoint URL. Fair on making is a constant though!

shrugs: ah, cool 👍

find-gas-usage.js
shrugs: lmao good software ಠ_ಠ

PR #187, opened 2019-07-09
WIP: Indigo-ify Admin Views

shrugs:


shrugs:
stopping for tonight, will pick up on monday. worth a pre-emptive review for sure.

src/lib/txn.js
```diff
@@ -209,6 +181,9 @@ const isTransactionConfirmed = async (web3, txHash) => {
   return receipt !== null;
 };
 
+const sendTransactionsAndAwaitConfirm = async (web3, signedTxs, usedTank) =>
+  Promise.all(signedTxs.map(tx => sendSignedTransaction(web3, tx, usedTank)));
+
```
Fang-: This resolves on-submit, not on-confirm, as per the multiple `resolve` calls in `sendSignedTransaction`. Fine with punting on that until the refactoring of this lib, but we shouldn't name this misleadingly in the meantime.

src/store/lib/useOptimisticDetailsStore.js
```diff
@@ -0,0 +1,38 @@
+import { useCallback } from 'react';
+import { Just, Nothing } from 'folktale/maybe';
+
+import useSetState from 'lib/useSetState';
+
+const DEFAULT_OPTIMISTIC_DETAILS = {
+  keyRevisionNumber: Nothing(),
```
Fang-: Hmm, not sure if this is still needed, if we always make the user wait for transaction confirm. If we assume we're fully synchronous with the blockchain, we can get rid of all optimistic logic entirely.

src/views/Admin.js
```diff
@@ -1,84 +1,67 @@
...
+import Crumbs from 'components/Crumbs';
 
-  const goRedownload = useCallback(() => history.push(ROUTE_NAMES.REDOWNLOAD), [
-    history,
-  ]);
+import AdminHome from './Admin/AdminHome';
+import AdminEditPermissions from './Admin/AdminEditPermissions';
+import AdminRedownload from './Admin/AdminRedownload';
+import AdminReticket from './Admin/AdminReticket';
+import AdminSetProxy from './Admin/AdminSetProxy';
```
Fang-: I'd personally not prefix these with `Admin` if they're already in the `/Admin` folder, but I guess I can see how this is nice for identifiability by just bare filename. They're still unique names regardless though.

Fang-: Also, in the future, might be good to `git mv` the renaming before/after making changes, so that there's a legible diff rather than completely "new" files. Looks like it did pick up on that for _some_ files, but wasn't smart enough to catch them all.

src/views/Admin/Reticket/ReticketExecute.js
```diff
@@ -96,33 +113,51 @@ export default function DoReticket({ newWallet, completed }) {
...
+                this page until the process is complete.
+              </Text>
+            </Grid.Item>
+          </Grid.Item>
+
+          {renderAdditionalInfo()}
+        </>
+      )}
+
+      {isDone && (
```
Fang-: I'd prefer a proper if-else/? : here.

src/views/Point.js
```diff
@@ -16,33 +15,26 @@ import { matchBlinky } from 'components/Blinky';
...
-        const address = need.addressFromWallet(wallet);
-        return (
-          eqAddr(address, pointDetails.owner) ||
-          eqAddr(address, pointDetails.managementProxy)
-        );
-      },
-    });
-  }
+  const { canManage } = usePermissionsForPoint(
+    // using empty string should be ok here
```
Fang-: The fact that we can't "just like use" pointCursor on a view that's centered around it is ridiculous. Maybe has gone too far. (But this is of later concern...)

src/lib/invite.js
```diff
@@ -43,8 +30,8 @@ const TRANSACTION_STATES = {
     label: 'Funding Transactions...',
     progress: 0.3,
   },
-  CLAIMING: {
-    label: 'Claiming Invite...',
+  TRANSFERRING: {
+    label: 'Transferring Point...',
```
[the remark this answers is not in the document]
shrugs: Totally - we are using the same flow for reticketing, though, and 'claiming invite' is super out of place in that context. any ideas?

src/lib/keys.js
```diff
@@ -64,68 +76,90 @@ const genKey = (networkSeed, point, revision) => {
...
-    const networkSeedDerivable = eqAddr(walProxy, chainProxy);
+/**
+ * @param {string} seed
+ * @param {number} revision
+ * @return {Promise<Maybe<string>>}
+ */
+const deriveNetworkSeedFromSeed = async (seed, revision) => {
+  const networkSeed = await kg.deriveNetworkSeed(seed, '', revision);
+  // apparently deriveNetworkSeed can return empty string
+  // and perhaps even malformed
```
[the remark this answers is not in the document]
shrugs: I haven't reproduced either — the previous code always checked the result of deriveNetworkSeed against empty string, so I'm doing it too — I'd love to remove this check!

src/lib/keys.js
```diff
@@ -50,9 +51,20 @@ const jam = seed => {
...
-  const bnsec = new BN(pair.crypt.private + pair.auth.private + '42', 'hex'); // '42' is curve parameter
+// simple alias to avoid importing kg into component scope
+export const deriveNetworkKeys = seed => kg.deriveNetworkKeys(seed);
+
+/**
+ * @param {object} pair
+ * @param {number} point
+ * @param {number} revision
+ * @return {string}
+ */
```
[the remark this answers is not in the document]
shrugs: If you feel like it — I was mostly doing this to help me remember what types things are as I'm refactoring them, but I don't feel strongly about enforcing docstrings across the library files.

src/lib/txn.js
shrugs: ohhh that's why the errors aren't being caught. this code is (attempting to be) semantically the same as the previous version, so update this to wait for each of those transactions

src/store/lib/useOptimisticDetailsStore.js
shrugs: that's true, we can just make them wait and then just refetch the point details etc. I prefer doing that, since getting into optimistic state is a super headache. can we confirm with alex/jimmy that blocking is a-ok?

src/views/Admin.js
shrugs: totally - I prefix them so that fuzzy search in editors works better and the chances of a collision between folders (like what could happen with `Activate/Passport` and `components/Passport`) to avoid confusion. I don't feel _super_ strongly about this though, so I'm willing to revert that pattern. lmk

src/views/Point.js
shrugs: yeah totally, this is a `need.wallet` situation 👍

src/lib/invite.js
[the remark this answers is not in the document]
Fang-: Oh, that's right, fug... I get the feeling we maybe want a little flag for that? Invite flow should be baby gloves, but I'm fine saying "transferring to new wallet" or whatever during reticketing.

src/lib/keys.js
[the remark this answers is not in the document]
Fang-: Hah. I'm fine cargo-culting that then, though I'm pretty sure we _can_ remove it. If anything, the "perhaps even malformed" should be removed, it's more spooky than needed.

src/store/lib/useOptimisticDetailsStore.js
Fang-: As per email: confirmed that's a-ok!

src/lib/invite.js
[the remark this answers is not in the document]
shrugs: I'd be down to have a more general `executeReticket` lib function that conveys progress with technical terms and then wrap that logic with `claimInvite` and `reticket` functions that translate that technical progress into our higher level abstractions that show up in the UI. same logic, different presentation.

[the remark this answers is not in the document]
shrugs: honestly that presentation logic should live in the component. I'd move the progress labels into the components and instead just return the progress float like `0.5` from the reticket lib function.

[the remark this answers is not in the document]
Fang-: That sounds very reasonable!

src/lib/constants.js
```diff
@@ -13,6 +18,10 @@ const GALAXY_ENTROPY_BITS = 384;
 
 const SEED_ENTROPY_BITS = 128;
 
+// the null result if a key is unset within azimuth
+const ZERO_KEY =
```
shrugs: is this the best place to put this constant?

src/lib/keys.js
```diff
@@ -64,68 +75,100 @@ const genKey = (networkSeed, point, revision) => {
...
+export const deriveNetworkSeedFromMnemonic = async (
+  wallet,
+  authMnemonic,
+  details,
+  revision = 1
+) => {
+  const isOwner = eqAddr(wallet.address, details.owner);
+  const isManagementProxy = eqAddr(wallet.address, details.managementProxy);
+
+  // the network seed is only derivable from mnemonic if the derived
```
shrugs: is this comment (and impl) true? the original code only allowed it if the mnemonic was the management proxy but it seems to make sense that you should also be able to do this if you're the owner.

src/lib/tank.js
```diff
@@ -1,6 +1,6 @@
 // /lib/tank: functions for funding transactions
 
-import * as retry from 'async-retry';
+import retry from 'async-retry';
```
shrugs: did this import style not work initially?

src/lib/walletgen.js
```diff
@@ -55,12 +55,12 @@ export const generateWallet = async (point, ticket, boot) => {
     boot: boot,
   };
 
-  const wallet = await kg.generateWallet(config);
-
   // This is here to notify anyone who opens console because the thread
   // hangs, blocking UI updates so this cannot be done in the UI
   console.log('Generating Wallet for point address: ', point);
 
+  const wallet = await kg.generateWallet(config);
```
shrugs: (just changing the order so the console.log appears _before_ the ui hangs)

src/store/lib/useRekeyDateStore.js
```diff
@@ -37,26 +39,30 @@ export default function useRekeyDateStore() {
...
+          [point]: Just(Result.Error('No rekey date available.')),
         });
       } else {
         // last log is most recent
         const blockNumber = logs[logs.length - 1].blockNumber;
         const block = await _web3.eth.getBlock(blockNumber);
         addToRekeyDateCache({
-          //TODO date format
-          [point]: Just(new Date(block.timestamp * 1000).toString()),
+          [point]: Just(Result.Ok(new Date(block.timestamp * 1000))),
```
shrugs: how do we feel about the `Maybe<Result<>>` pattern here? lets us handle the 'loading' state by checking for `Maybe.Nothing` and handle the failure state by checking `Maybe.Just(Result.Error)` and success by `Maybe.Just(Result.Ok)`

src/style/bridge.scss
```diff
@@ -26,3 +26,57 @@ body {
...
+    width: 24px;
+    height: 100%;
+    background-color: #000;
+    border: none;
+    border-radius: 0;
+    cursor: ew-resize;
+  }
+
+  /* IE */
+  &::-ms-fill-lower {
```
shrugs: have not actually tested this on IE — copied this from the internet lol

src/views/Login.js
```diff
@@ -71,26 +93,22 @@ export default function Login() {
...
+    // if no point cursor set by login logic, try to deduce it
     if (Nothing.hasInstance(deduced)) {
       const owned = await azimuth.azimuth.getOwnedPoints(
         _contracts,
         _wallet.address
       );
       if (owned.length === 1) {
         deduced = Just(owned[0]);
-      } else if (owned.length === 0) {
-        const canOwn = await azimuth.azimuth.getTransferringFor(
```
shrugs: removed because acceptance will be on the 'points' page, so we want those users to end up at points

src/views/Login.js
```diff
@@ -71,26 +93,22 @@ export default function Login() {
...
-          _contracts,
-          _wallet.address
-        );
-        if (canOwn.length === 1) {
-          deduced = Just(canOwn[0]);
-        }
       }
     }
 
+    setDeducing(false);
```
shrugs: i had some of these calls take a long enough time on local that I decided to add a loader

src/indigo-react/lib/useForm.js
```diff
@@ -192,11 +192,16 @@ export default function useForm(inputConfigs = []) {
...
+              ? datas[i]
+              : initialValue !== undefined
+              ? initialValue
+              : '';
           return {
             // Input props
             name,
-            data: datas[i],
+            value: value,
+            data: passes[i] ? datas[i] : undefined,
```
shrugs: now `value` is the raw value and `data` is either `value` or `undefined` depending on pass state

this lets us use `data` in callers and assume that if it's not undefined it's valid

shrugs:
todo:

- [ ]  manual network seed
- [ ]  network seed discontinuity
- [ ]  allow unset of transfer using same button in admin/transfer

src/views/Party/PartySetPoolSize.js
```diff
@@ -0,0 +1,145 @@
...
+        <Grid.Item full as={MiniBackButton} onClick={() => pop()} />
+        <Grid.Item full as={ViewHeader}>
+          Set Pool Size
+        </Grid.Item>
+        {!spawnIsDelegatedSending && (
+          <Grid.Item full as={WarningBox} className="mb4">
+            The spawn proxy must be set to{' '}
+            <code className="mono f6">
+              {_contracts.delegatedSending.address}
+            </code>{' '}
```
Fang-: Eww. At this point, should we maybe just `{''}` the entire lines surrounding it, instead of putting these in?

src/lib/useEthereumTransaction.js
```diff
@@ -47,13 +47,15 @@ export default function useEthereumTransaction(
   const [chainId, setChainId] = useState();
   const [nonce, setNonce] = useState();
   const [gasPrice, setGasPrice] = useState(initialGasPrice); // gwei
+  const [averageGasPrice, setAverageGasPrice] = useState(gasPrice); // gwei
```
Fang-: It was initially unclear to me what this was for, should maybe be renamed `recommendedGasPrice`.

src/views/Admin/AdminNetworkingKeys.js
```diff
@@ -65,54 +76,56 @@ function useSetKeys() {
...
+      }
+      const pair = deriveNetworkKeys(_seed);
+
+      const txn = azimuth.ecliptic.configureKeys(
+        _contracts,
+        _point,
+        addHexPrefix(pair.crypt.public),
+        addHexPrefix(pair.auth.public),
+        CRYPTO_SUITE_VERSION,
+        isDiscontinuity ? 1 : 0
```
Fang-: `true : false` (and probably also just `isDiscontinuity`) should work here, right? Why is this written like this?

src/views/Admin/AdminNetworkingKeys.js
```diff
@@ -134,19 +147,62 @@ export default function AdminNetworkingKeys() {
...
     bind,
   } = useSetKeys();
 
-  const setDidRequestSetKeys = useCallback(() => {
-    setKeys();
-    _setDidRequestSetKeys(true);
-  }, [setKeys, _setDidRequestSetKeys]);
+  useEffect(() => {
+    if (
+      (!showNetworkSeed || (showNetworkSeed && validNetworkSeed)) &&
```
Fang-: Can we get a comment about this clause?

shrugs:
todo

- [ ] ensureFunds in `useEthereumTransaction` / `InlineEthereumTransaction`
- [ ] replace 'Create Galaxy' and 'Issue Child' views and then nuke every old view related to admin actions

src/indigo-react/lib/useForm.js
shrugs: right yeah — lets us remove the onValue property within Input which was pretty hacky

src/indigo-react/components/IconButton.js
```diff
@@ -37,7 +41,6 @@ export default function IconButton({
         height: '1rem',
         fontSize: '1rem',
         fontWeight: 'bold',
-        // TODO: ^ replace with 1:1 svgs and remove explicit sizing
```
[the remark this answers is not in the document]
shrugs: changed our mind, since the icons are actually built into the Inter font — `1x1rem` is actually correct now

src/indigo-react/components/IconButton.js
```diff
@@ -10,6 +10,7 @@ export default function IconButton({
   disabled = false,
   secondary = false,
   solid = false,
+  flush = false, // whether or not the icon should include horizontal padding
```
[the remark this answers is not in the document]
shrugs: correct — flush = no padding. i feel like 'with padding' should be the default, so I went for 
'flush' as the optionally more specific version of the button. could change it to 'inline' or something?

src/lib/useEthereumTransaction.js
```diff
@@ -0,0 +1,179 @@
...
+  const [txHash, setTxHash] = useState();
+
+  const initializing = nonce === undefined || chainId === undefined;
+  const constructed = !!unsignedTransaction;
+  const signed = state === STATE.SIGNED;
+  const broadcasted = state === STATE.BROADCASTED;
+  const confirmed = state === STATE.CONFIRMED;
+
+  // disable the inputs when:
+  const inputsLocked = signed || broadcasted || confirmed;
```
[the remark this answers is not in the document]
shrugs: I'd like us to use a real state machine/enum library, which could give us operators like `const locked = state <= STATE.CONFIRMED` or `state.before(STATE.CONFIRMED)`, but didn't think it was worth the addition just yet.

In this case we're ok if we're diligent about including an exhaustive list of states in these conditions, but it's obviously prone to error.

should we look at a simple state machine lib?

src/views/Admin/AdminNetworkingKeys.js
```diff
@@ -0,0 +1,324 @@
...
+            Unset
+          </Grid.Item>
+        ),
+        Just: ({ value: key }) => (
+          <>
+            <Grid.Item full as="code" className="f5 wrap">
+              0x
+            </Grid.Item>
+            <Grid.Item full as="code" className="f5 wrap">
+              {renderNetworkKey(key)}
```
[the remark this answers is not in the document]
shrugs: I believe so. at the very least a 'copy' button would solve that problem. we should have copy-able text across the app (addresses, keys, hex codes, etc etc) — worth doing a pass as a separate PR.

https://github.com/urbit/bridge/issues/196

src/views/Admin/AdminSetProxy.js
```diff
@@ -0,0 +1,199 @@
...
+          className={cn('mono', {
+            black: !confirmed && isProxySet,
+            gray4: !confirmed && !isProxySet,
+            green3: confirmed,
+          })}>
+          {isProxySet ? proxyAddress : 'Unset'}
+        </Flex.Item>
+        {!confirmed && isProxySet && (
+          <Flex.Item as={LinkButton} onClick={unset} disabled={inputsLocked}>
+            Unset
```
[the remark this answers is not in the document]
shrugs: updated!

src/views/Invite/InviteEmail.js
```diff
@@ -121,7 +123,7 @@ export default function InviteEmail() {
   const [hovered, setHovered] = useSetState();
   const [invites, addInvite, clearInvites] = useSetState();
   const [receipts, addReceipt, clearReceipts] = useSetState();
-  const [errors, addError, clearErrors] = useSetState();
```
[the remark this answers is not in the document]
shrugs: correct, this is still unimplemented. I removed it to get rid of the eslint warning.

src/views/Party/PartySetPoolSize.js
shrugs: honestly I thought the same as you here, but I've learned to live with them as a gimmick of jsx. my editor automatically inserts them when necessary and everything, so it seems to be the 'right' way to do this. constructing larger strings within `{' ... '}` seems even hackier, imo

wdyt?

src/views/Admin/AdminNetworkingKeys.js
shrugs: ah, thought it had to be a number — sgtm

src/indigo-react/components/IconButton.js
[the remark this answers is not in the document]
Fang-: Nah the name is fine, just thinking the comment should be more like "if false, we add padding" or similar.

src/lib/useEthereumTransaction.js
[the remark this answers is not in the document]
Fang-: No thanks for now. If we run into this/consider it painful again then we might want to take a look, but we have plenty better places to spend our time.

src/views/Party/PartySetPoolSize.js
Fang-: I think editors are not a valid argument, and still prefer the larger `{' ... '}`, but don't feel super strongly.

Fang-:
On local testnet, after I set a management proxy, the following was displayed:

image

Hitting the back button hadn't loaded in the new management proxy yet either, but relogging got me the new value.

Fang-:
When configuring keys, after sending the transaction, it first briefly displays the "cannot download non-deterministic keys" prompt, but a couple seconds later updates to give me a functional "download keyfile" button. We should just wait in the "sending" state until the latter is achieved, instead of showing the user a soon-to-be-outdated prompt.

Fang-:
Changes so far seem good otherwise. (:

Fang-:
image

It shouldn't tell me this if I never configured keys in the first place.

Fang-:
It also displays that message even when I generated a nondeterministic seed in the current session. It should be able to still get that from memory, right?

(Sorry for piling these on like this, just filing what I run into as I work on other things, don't want this to get lost. (: )

shrugs:
> On local testnet, after I set a management proxy, the following was displayed:
> 
> image
> 
> Hitting the back button hadn't loaded in the new management proxy yet either, but relogging got me the new value.

I think what I'll do is convince the inline transaction component to consider itself 'loading' while we're re-fetching things like details, etc. that way there's no state where 'the tx is confirmed but we haven't yet refetched on-chain info'

> When configuring keys, after sending the transaction, it first briefly displays the "cannot download non-deterministic keys" prompt, but a couple seconds later updates to give me a functional "download keyfile" button. We should just wait in the "sending" state until the latter is achieved, instead of showing the user a soon-to-be-outdated prompt.

this should be fixed in the latest—I attempted the pattern mentioned above. lmk how it feels

> It shouldn't tell me this if I never configured keys in the first place.

fixed

> It also displays that message even when I generated a nondeterministic seed in the current session. It should be able to still get that from memory, right?

right now I only store nondeterministic/custom seeds in the state of the `AdminNetworkingKeys` component, so the keyfile button in there can download manual keys but the keyfile button in `AdminHome` doesn't have access to that state. we can move that state higher in the tree if necessary, but imo this makes sense if the admin button is more of a 'redownload' and users therefore won't expect to rely on it

shrugs:
@Fang- can you 👀 my changes in https://github.com/urbit/bridge/pull/187/commits/8282b799eaaf327b15ecf61e8b6f6ce690a88213 and let me know if they're reasonable? I think the network keys behavior needs to be looked at, since I'm not actually 100% clear on the spec. lmk if it's behaving as you expect. 

sometimes my revision isn't incremented even after the transaction is successful, so I'm not surw what's up

Fang-:
> we can move that state higher in the tree if necessary, but imo this makes sense if the admin button is more of a 'redownload' and users therefore won't expect to rely on it

Having it higher and available for the duration of the session was the behavior of old Bridge, though it was less fancy about the flow in general. If it's easy/clean to just chuck it in there, maybe with a "download keys (tmp)" notification if you think the reliance would become an issue, then go for it. If it's non-trivial work, just let it be.

re: keys stuff, I'll take a look. (:

Fang-:
The code looks correct, and it gives identical results to old Bridge, so that's approved. I think I messed up the thing you fixed during my PR, sorry!

Can't reproduce your "revision not incremented error", though it also doesn't prevent me from clicking the back button until we have that data. So I can "back" to see the "permissions" list with revision 1, and then see it update to 2 before my very eyes. Not that big a deal.

Alternative solution to the "not storing ndKeys in session" problem would be to disable the "return" button until the user downloads the keyfile in such a case. Also not the biggest deal.

shrugs:
> I think what I'll do is convince the inline transaction component to consider itself 'loading' while we're re-fetching things like details, etc. that way there's no state where 'the tx is confirmed but we haven't yet refetched on-chain info'

implemented in the latest commit

shrugs:
> though it also doesn't prevent me from clicking the back button until we have that data. So I can "back" to see the "permissions" list with revision 1, and then see it update to 2 before my very eyes. Not that big a deal.

ahhh whoops, ok I'm going to disable the back button across all of those views... although we'd also have to disable our router's pop functionality, which gets into 'takes longer than 2 minutes' territory, so I'm going to postpone this for later

shrugs:
sneaking a merge in while you're asleep 😉 

(but really, I'm assuming that the new IssueChild isn't a surprise and that all previous code is a-ok)

shrugs:
image
honestly kind of sad knowing that there's no big drop in line count for hooks code. I was sold line-count reductions 😢

PR #188, opened 2019-07-09
Email invite fixes

Fang-:
Also an unrelated indigo-react export addition that was missing. I _think_ I'm fine just adding that in, but feel free to yell at me @shrugs.

Draft because I'm still trying to fix an issue in this code:

https://github.com/urbit/bridge/blob/a531c3055d3d83bc7057a613209e9ee5361c99b3/src/views/Invite/InviteEmail.js#L141-L156

Note how line 145 checks if `config.name` has received an invite yet, instead of using the actual value from the input. `config` doesn't contain that. `inputs`, a couple lines below, does, but it's out of scope for the logic that needs changing. @shrugs I've thought about doing just `.map` over `dynamicConfigs` _after_ `inputs` is defined, but I get the feeling that would mess this up. Can you recommend an approach?

### DIFF
--- a/src/components/old/StatelessTransaction.js
+++ b/src/components/old/StatelessTransaction.js
@@ -8,6 +8,7 @@ import { Button } from './Base';
 import { CheckboxButton, Input, InnerLabel, InnerLabelDropdown } from './Base';
 import { Warning } from './Base';
 
+import { GAS_LIMITS } from '../../lib/constants';
 import { BRIDGE_ERROR } from '../../lib/error';
 import { ROUTE_NAMES } from '../../lib/routeNames';
 import { compose } from '../../lib/lib';
@@ -40,7 +41,7 @@ class StatelessTransaction extends React.Component {
 
     this.state = {
       gasPrice: '5',
-      gasLimit: '600000',
+      gasLimit: '' + GAS_LIMITS.DEFAULT,
       showGasDetails: false,
       userApproval: false,
       chainId: '1',
--- a/src/components/old/StatelessTransaction.js
+++ b/src/components/old/StatelessTransaction.js
@@ -41,7 +41,7 @@ class StatelessTransaction extends React.Component {
 
     this.state = {
       gasPrice: '5',
-      gasLimit: '' + GAS_LIMITS.DEFAULT,
+      gasLimit: GAS_LIMITS.DEFAULT.toString(),
       showGasDetails: false,
       userApproval: false,
       chainId: '1',
--- a/src/lib/useMailer.js
+++ b/src/lib/useMailer.js
@@ -43,14 +43,10 @@ export default function useMailer(emails) {
   const _sendMail = useCallback(async (email, ticket, rawTx) => {
     if (STUB_MAILER) {
       console.log(`${email} - ${ticket}`);
-      return;
+      return true;
     }
 
-    const mailSuccess = await sendMail(email, ticket, rawTx);
-
-    if (!mailSuccess) {
-      throw new Error(`Internal mailing error when mailing ${email}`);
-    }
+    return await sendMail(email, ticket, rawTx);
   }, []);
 
   return { ...hasReceivedCache, sendMail: _sendMail };
--- a/src/lib/useMailer.js
+++ b/src/lib/useMailer.js
@@ -9,14 +9,14 @@ const STUB_MAILER = process.env.REACT_APP_STUB_MAILER === 'true';
 function useHasReceivedCache() {
   const [cache, addToCache] = useSetState();
 
-  const getHasRecieved = useCallback(
+  const getHasReceived = useCallback(
     email => cache[email] || Nothing(), //
     [cache]
   );
 
   const syncHasReceivedForEmail = useCallback(
     async email => {
-      if (Just.hasInstance(getHasRecieved(email))) {
+      if (Just.hasInstance(getHasReceived(email))) {
         // never update the cache after we know about it
         return;
       }
@@ -29,10 +29,10 @@ function useHasReceivedCache() {
       const _hasReceived = await hasReceived(email);
       addToCache({ [email]: Just(_hasReceived) });
     },
-    [getHasRecieved, addToCache]
+    [getHasReceived, addToCache]
   );
 
-  return { getHasRecieved, syncHasReceivedForEmail };
+  return { getHasReceived, syncHasReceivedForEmail };
 }
 
 export default function useMailer(emails) {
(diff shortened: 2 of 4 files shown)


src/lib/useMailer.js
```diff
@@ -9,14 +9,14 @@ const STUB_MAILER = process.env.REACT_APP_STUB_MAILER === 'true';
 function useHasReceivedCache() {
   const [cache, addToCache] = useSetState();
 
-  const getHasRecieved = useCallback(
+  const getHasReceived = useCallback(
```
shrugs: great catch on that spelling 👍

shrugs:
ah yeah you're right. this is something that stumps me as well, given the mutual dependencies between the useForm logic and any additional logic we want to use that depends on the current values.

i feel like we should probably provide a `async validate()` argument to useForm that accepts arguments like `await validate(config.name, values[i], configs, values)` or something like that where we can inject our custom validation logic. then in useForm we'd have a `const [validating, setValidating] = useSetState({})` line where we keep track of which fields are currently being async-validated (and then pass that info to the `*Input` components that can show a blinky/spinner accessory or some other loading state).

then our custom validate function for this page might look like

```js

const validate = useCallback(async (name, value) => {
  const hasReceived = await getHasReceived(value);
  if (!hasReceived) {
    return 'This email already received an invite.'
  }
}, [...])

const [...] = useForm(dynamicConfigs, { validate })

```

I'm gonna take a crack at this real quick and let you know if I run into any blockers.

shrugs:
yeah.... this is annoying as hell. what the actual fuck, how are we supposed to architect this? doesn't feel natural at all.

anyway, I got it working pretty well on this branch https://github.com/urbit/bridge/tree/shrugs/email-fixes feel free to pull it and test it out. might be worth architecting your own solution before you see mine, to avoid influence.

Fang-:
I tried out a bunch of different things that were "dumb enough that they might just work". But of course they all broke in different ways, mostly just state being eaten, because sufficiently advanced JavaScript forces you to over-engineer rather than slap on a simple solution.

> what the actual fuck, how are we supposed to architect this?

The _real_ answer would be to Use Less JavaScript. But that would mean reverting this back into stateful React. I think it's too late to backpedal on that, at least for now. Your change, ~200 new/changed LoC spread over 8 files feels absolutely insane for a dumb little thing like this, but I don't have any better alternatives to offer...

Perhaps there's a way to consolidate the `inputConfigs` and `inputs` into a single object? Like, what I had dreamed of having was stateful input components that could run their own on-change logic, display their own errors, etc.

No man should have to witness these crimes, let alone take part in them. But here we are, blood on our hands, wondering how it ever got this far...

shrugs:
@Fang- can you expand on the "consolidate" idea and what your dream API would look like?

> it's just us, standing over the dead body of `useForm.js` as the sirens approach, ever closer
> we look to each other, resigned, locking eyes as we both hit the 'merge' button and seal our fate

Fang-:
I'm just a little bit confused about why we have `inputConfigs` _and_ `dynamicConfigs` _and_ `inputs`. I mean, I "get" it, but everything about this just seems overengineered...

I'm imagining something like:

```js
buildInputConfig({ label: 'Email Address', autoFocus: true, onChange: (value, input) => {
  input.value = value;
  if (getHasReceived(value)) {
    input.error = "Already received an invite.";
  }
} })
```

shrugs:
that behavior is basically what we've encapsulated with the (sync) validators. making them dynamic is basically the same as providing that synchronous callback for validation. is it the verbosity around dynamicInputs that's the problem? we can always inline the `useMemo` hook and avoid the extra variable. the memo is important because it stops the validators from getting a new identity (and re-triggering validation) when not necessary. the same thing would be necessary for that `onChange` handler, which would have to be `useCallback`'d to avoid constant re-computes.

I've decided that I don't like hooks anymore. I'll change my mind when someone writes a form library that doesn't suck (the existing options are either contrived or don't do what we want)

shrugs:
https://github.com/iamkevinwolf/formal actually looks pretty decent

Fang-:
We got ourselves into this mess by adding to our abstractions stack, we can get ourselves out of it by adding to our abstractions stack!

Half-joking aside, I'm not a great judge of these things. This seems really young (v0!) and the examples it offers seem a bit shallow, but if this is stable and a good fit, then maybe it's fine? Your call, @shrugs.

shrugs:
haven't actually looked at it yet - will take a keen eye on monday

Fang-:
Alright I hacked something in!

The code makes it seem like `addError` would actually get displayed for the component, but it doesn't. More `dynamicConfigs` jank? ):

src/views/Invite/InviteEmail.js
```diff
@@ -187,6 +187,27 @@ export default function InviteEmail() {
       throw new Error('Internal Error: Missing Contracts/Web3/Wallet');
     }
 
+    //TODO want to do this on-input, but that gets weird. see #188
+    let errorCount = 0;
+    const mailsVerified = inputs.map(async input => {
```
shrugs: nothing is returned from this map — is mailsVerified used? we probbaly want either a `reduce` to collect the number of errors or just a normal for loop for iterating the inputs

src/components/old/StatelessTransaction.js
```diff
@@ -40,7 +41,7 @@ class StatelessTransaction extends React.Component {
 
     this.state = {
       gasPrice: '5',
-      gasLimit: '600000',
+      gasLimit: '' + GAS_LIMITS.DEFAULT,
```
shrugs: `GAS_LIMITS.DEFAULT.toString()`?

shrugs:
addError should be included in the dynamicInputs mapping, setting `input.error = errors[input.name]` or something?

src/views/Invite/InviteEmail.js
Fang-: Is this not also an ES6-acceptable way of "do x for every item in array"? `mailsVerified` holds the promises, right?

Fang-:
> addError should be included in the dynamicInputs mapping, setting input.error = errors[input.name] or something?

Well it already does that in `dynamicConfigs`, which _should_ carry over into the `inputs = useForm(dynamicConfigs)`, right?

https://github.com/urbit/bridge/blob/f9dc008e45fd98312853007ebbe5701fa1fcff96/src/views/Invite/InviteEmail.js#L149

Fang-:
Gonna assume this is all good now. (:

PR #189, opened 2019-07-12
indigo-admin helper

Fang-:
@shrugs Went ahead and did some small bits of work here, mostly dealing with making UI match chain logic. Hope this doesn't overlap with anything un-pushed!

(A little bit unsure about my `useMemo` usage. Had `availableInvites` fetching in there as well at first, (we really only want to do that in the planet case), but hooks in callbacks are invalid. Maybe this should just be an anonymous function call now instead.)

### DIFF
--- a/src/views/Admin/AdminEditPermissions.js
+++ b/src/views/Admin/AdminEditPermissions.js
@@ -45,6 +45,7 @@ export default function AdminEditPermissions() {
   const { pointCursor } = usePointCursor();
   const { getDetails } = usePointCache();
 
+  const _web3 = need.web3(web3);
   const _contracts = need.contracts(contracts);
 
   const [rekeyDate, setRekeyDate] = useState(Nothing());
@@ -59,7 +60,7 @@ export default function AdminEditPermissions() {
   const { canManage, isOwner } = usePermissionsForPoint(userAddress, point);
 
   useLifecycle(() => {
-    getRekeyDate(need.web3(web3), _contracts, point).then(res => {
+    getRekeyDate(_web3, _contracts, point).then(res => {
       if (res === null) setRekeyDate(Just(new Date()));
       else setRekeyDate(Just(res));
     });
--- a/src/views/Admin/AdminEditPermissions.js
+++ b/src/views/Admin/AdminEditPermissions.js
@@ -74,7 +74,7 @@ export default function AdminEditPermissions() {
 
   const proxyAction = (proxyType, address, enabled) => {
     if (eqAddr(address, ETH_ZERO_ADDR)) {
-      address = 'Not yet set';
+      address = 'Not set';
     } else if (eqAddr(address, _contracts.delegatedSending.address)) {
       address = address + ' (invites contract)';
     }
--- a/src/views/Admin/AdminEditPermissions.js
+++ b/src/views/Admin/AdminEditPermissions.js
@@ -2,6 +2,7 @@ import React, { useState, useCallback } from 'react';
 import { Just, Nothing } from 'folktale/maybe';
 import * as need from 'lib/need';
 import { Grid } from 'indigo-react';
+import { azimuth } from 'azimuth-js';
 
 import { ForwardButton } from 'components/Buttons';
 import { matchBlinky, matchBlinkyDate } from 'components/Blinky';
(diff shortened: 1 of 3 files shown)


src/views/Admin/AdminEditPermissions.js
```diff
@@ -43,24 +45,24 @@ export default function AdminEditPermissions() {
...
+  const { canManage, isOwner } = usePermissionsForPoint(userAddress, point);
 
   useLifecycle(() => {
-    getRekeyDate(need.web3(web3), need.contracts(contracts), point).then(
-      res => {
-        if (res === null) setRekeyDate(Just(new Date()));
-        else setRekeyDate(Just(res));
-      }
-    );
+    getRekeyDate(need.web3(web3), _contracts, point).then(res => {
```
shrugs: since we're `need`ing contracts and point we might as well need.web3 outside of this block as well

src/views/Admin/AdminEditPermissions.js
```diff
@@ -71,8 +73,10 @@ export default function AdminEditPermissions() {
   const goSetKeys = useCallback(() => push(names.SET_KEYS), [push, names]);
 
   const proxyAction = (proxyType, address, enabled) => {
-    if (address === ETH_ZERO_ADDR) {
-      address = 'Not yet set';
+    if (eqAddr(address, ETH_ZERO_ADDR)) {
+      address = 'Not set';
+    } else if (eqAddr(address, _contracts.delegatedSending.address)) {
+      address = address + ' (invites contract)';
```
shrugs: 👍nice

src/views/Admin/AdminEditPermissions.js
```diff
@@ -114,17 +121,11 @@ export default function AdminEditPermissions() {
       </Grid.Item>
       <Grid.Divider />
 
-      {proxyAction(PROXY_TYPE.MANAGEMENT, details.managementProxy, canManage)}
-      <Grid.Divider />
+      {proxyAction(PROXY_TYPE.MANAGEMENT, details.managementProxy, isOwner)}
```
shrugs: are these really only available to owners? a transfer proxy can't change transfer proxy? not even a management key can change proxies?

src/views/Point.js
```diff
@@ -48,6 +51,30 @@ export default function Point() {
     history,
   ]);
 
+  const inviteButton = useMemo(() => {
```
shrugs: you'll want to raise all of those callbacks out of this function, and there's no real need to useMemo here.

```jsx
const goInvite = useCallback(() => history.push(ROUTE_NAMES.INVITE), [push, names]);
const goParties = useCallback(() => history.push(ROUTE_NAMES.INVITES_MANAGE), [push, names]);

const inviteButton = (() => {
    switch (azimuth.getPointSize(point)) {
      case azimuth.PointSize.Planet:
        return (
          <FooterButton disabled={!isOwner} onClick={goInvite}>
            Invite <sup>{matchBlinky(availableInvites)} available</sup>
          </FooterButton>
        );
      case azimuth.PointSize.Star:
        return (
          <FooterButton disabled={!isOwner} onClick={goParties}>
            Manage parties
          </FooterButton>
        );
      default:
        return null;
    }
})()
```

also please use the `names` object returned from `useHistory` so we can avoid pulling in `routeNames` everywhere.

src/views/Admin/AdminEditPermissions.js
Fang-: If a management key could change the transfer proxy then we'd have a problem. (^:

Hindsight does tell us that proxies should probably be able to change themselves, but that hasn't been officially proposed yet, let alone implemented/voted on.

Fang-:
Fixed!

Also made sure that we're following contract logic wrt only _active_ points being able to do invite things. This means that if a gifted planet is still temporarily held by its star('s owner), that star can't use that planet to send invites. (Trying to do so would fail on-chain.)

src/views/Point.js
```diff
@@ -45,35 +45,36 @@ export default function Point() {
 
   const showActions = Just.hasInstance(wallet);
 
-  const goAdmin = useCallback(() => history.push(ROUTE_NAMES.ADMIN), [history]);
+  const goAdmin = useCallback(() => push(names.ADMIN), [history, names]);
```
shrugs: the dependencies for these should be `[push, names]` — as far as I can tell `history` would be undefined here?

Fang-: Oh whoops, bit too hasty here.

_curses at less-loud compile errors_

PR #190, opened 2019-07-12
Implied ticket & navigation

Fang-:
Deduces the ticket from the url#hash so that we can pre-fill the field in the invite activation flow.* Also auto-navigates to login or activation if a point or invite ticket are implied respectively.

\* We may or may not want to do slightly more sophisticated ticket-in-url formatting at some point.

### DIFF
--- a/src/lib/useImpliedTicket.js
+++ b/src/lib/useImpliedTicket.js
@@ -7,8 +7,8 @@ import { prependSig } from './transformers';
  */
 export default function useImpliedTicket() {
   const hash = window.location.hash.slice(1);
-  const tick = prependSig(hash);
-  const isValid = ob.isValidPatq(tick);
+  const ticket = prependSig(hash);
+  const isValid = ob.isValidPatq(ticket);
 
-  return isValid ? tick : null;
+  return isValid ? ticket : null;
 }
--- a/src/lib/useImpliedTicket.js
+++ b/src/lib/useImpliedTicket.js
@@ -1,0 +1,14 @@
+import * as ob from 'urbit-ob';
+
+import { prependSig } from './transformers';
+
+/**
+ * pull the suggested ticket from the #hash in the url
+ */
+export default function useImpliedTicket() {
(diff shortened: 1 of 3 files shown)


src/lib/useImpliedTicket.js
```diff
@@ -0,0 +1,14 @@
+import * as ob from 'urbit-ob';
+
+import { prependSig } from './transformers';
+
+/**
+ * pull the suggested ticket from the #hash in the url
+ */
+export default function useImpliedTicket() {
```
shrugs: did we delete the version of this hook that I'd written at some point?

src/lib/useImpliedTicket.js
```diff
@@ -0,0 +1,14 @@
...
+  const hash = window.location.hash.slice(1);
+  const tick = prependSig(hash);
```
shrugs: `ticket`, to be clear?

src/lib/useImpliedTicket.js
Fang-: I found and used an existing one for the implied point from the subdomain, if that's what you're thinking of?

src/lib/useImpliedTicket.js
Fang-: You are clearly not a hoon programmer. (^:

src/lib/useImpliedTicket.js
shrugs: I thought I deleted that comment — sorry I'm slow today — I was just confused because of the similarities but I realized that they were different (and then didn't delete my confusing comment) — please ignore lol

src/lib/useImpliedTicket.js
shrugs: 😂

src/lib/useImpliedTicket.js
Fang-: Pretty sure today's not a bridge day for you anyway, so you are more than forgiven. (;

PR #191, opened 2019-07-12
Display disclaimer only on first run

Fang-:
Uses super dumb cookie logic to make that happen. If we ever do anything more with cookies (unlikely imo), we'll want a `/lib/cookies` or whatever.

Yes, this works for localhost. No, this doesn't work across hosts.

### DIFF
--- a/src/views/Landing.js
+++ b/src/views/Landing.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import { H1, P } from 'indigo-react';
 
-import { ROUTE_NAMES } from '../lib/routeNames';
 import { useHistory } from '../store/history';
 import useImpliedPoint from 'lib/useImpliedPoint';
 import useImpliedTicket from 'lib/useImpliedTicket';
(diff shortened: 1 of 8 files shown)


Fang-:
Note that this moves the disclaimer to always be the first thing you see, instead of seeing it after entering an invite code. Pretty sure this is the most desired flow. cc @alexmatz @eriknewton

src/views/Landing.js
```diff
@@ -12,9 +11,11 @@ import { ForwardButton } from 'components/Buttons';
 function Landing() {
   const { push, names } = useHistory();
 
-  if (useImpliedPoint()) {
+  const impliedPoint = useImpliedPoint();
+  const impliedTicket = useImpliedTicket();
+  if (impliedPoint) {
     push(names.LOGIN);
-  } else if (useImpliedTicket()) {
+  } else if (impliedTicket) {
```
shrugs: good catch

PR #192, opened 2019-07-15
Small unhappy-path fixes

Fang-:
Couple small catches.

### DIFF
--- a/src/lib/txn.js
+++ b/src/lib/txn.js
@@ -166,8 +166,11 @@ const sendSignedTransaction = (web3, stx, doubtNonceError, confirmationCb) => {
         // we just wait for first confirmation here.
         console.error(err);
         if (
-          doubtNonceError &&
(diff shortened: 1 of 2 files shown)


src/lib/txn.js
```diff
@@ -166,8 +166,11 @@ const sendSignedTransaction = (web3, stx, doubtNonceError, confirmationCb) => {
         // we just wait for first confirmation here.
         console.error(err);
         if (
-          doubtNonceError &&
-          (err.message || '').includes("the tx doesn't have the correct nonce.")
+          (err.message || '').includes('known transaction: ') ||
```
shrugs: can we move these into a variable like

```
const isKnownTransaction = (err.message || '').includes(...)
const isNonceError = doubtNonceError && (err.message || '').includes(...)
if (isKnownTransaction || isNonceError) {...}
```

PR #193, opened 2019-07-16
/lib/txn improvements

Fang-:
As discussed yesterday with @shrugs.

Async-retry seems unreliable when using `Infinity` retries, so we settle for `99999` instead.

`signTransaction` still needs its interface cleaned up. @shrugs I'm thinking the `txn` unsigned transaction object should just contain all of `nonce`, `gas`, etc. instead of passing those in as separate arguments.

We might still care about doing the `toHex` on those arguments though, since we're using `ethereumjs-tx`? Related, we might be able to get rid of the EIP-155 logic here if we [use ethereumjs-tx correctly](https://github.com/ethereumjs/ethereumjs-tx/#chain-and-hardfork-support).

### DIFF
--- a/src/lib/txn.js
+++ b/src/lib/txn.js
@@ -172,14 +172,12 @@ const waitForTransactionConfirm = (web3, txHash) => {
   }, RETRY_OPTIONS);
 };
 
-const sendAndAwaitAll = async (web3, stxs, doubtNonceError) => {
-  await Promise.all(
-    stxs.map(tx => {
-      return new Promise(async (resolve, reject) => {
-        const txHash = await sendSignedTransaction(web3, tx, doubtNonceError);
-        await waitForTransactionConfirm(web3, txHash);
-        resolve();
-      });
+// returns a Promise that resolves when all stxs have been sent & confirmed
+const sendAndAwaitAll = (web3, stxs, doubtNonceError) => {
+  return Promise.all(
+    stxs.map(async tx => {
+      const txHash = await sendSignedTransaction(web3, tx, doubtNonceError);
+      await waitForTransactionConfirm(web3, txHash);
     })
   );
 };
--- a/src/lib/txn.js
+++ b/src/lib/txn.js
@@ -63,6 +63,7 @@ const signTransaction = async config => {
     gasLimit,
   } = config;
 
+  //TODO require these in txn object
   nonce = toHex(nonce);
   chainId = toHex(chainId);
   gasPrice = toHex(toWei(gasPrice, 'gwei'));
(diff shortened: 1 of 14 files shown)


shrugs:
I'd love to remove the eip-155 logic if it'll 'just work' on an updated ethereum-tx. did you test it with your ledger?

src/lib/txn.js
```diff
@@ -195,23 +164,22 @@ const sendSignedTransaction = (web3, stx, doubtNonceError, confirmationCb) => {
...
-      console.log('tried, got', receipt);
-      let confirmed = receipt !== null;
-      if (confirmed) resolve(receipt.status === true);
-      else setTimeout(checkForConfirm, CHECK_BLOCK_EVERY_MS);
-    };
-    checkForConfirm();
-  });
+  return retry(async (bail, n) => {
+    const receipt = await web3.eth.getTransactionReceipt(txHash);
+    let confirmed = receipt !== null;
```
shrugs: nit: `const confirmed`

shrugs: also, 'confirmed' isn't technically correct here, right? this is more of a 'was the transaction processed' or `hasReceipt` boolean, since whether or not the tx is confirmed is down to the `receipt.status` state.

shrugs:
agreed, let's give the txn object all of its properties. if we remove eip-155 I don't think there's a reason for those to be separate anymore

src/lib/txn.js
Fang-: Pretty sure we still call that confirmed, even if status is false? A failed transaction still has an impact (gas cost), gets included in the chain, and thus gets confirmed with every block it's included in.

Fang-:
Did some pretty thorough testing with Ledger against Ropsten, but Ethereumjs-tx can't Just Do It for us...

We should still expect those arguments to be included in the txn object, we can just read out what we need from there. But I'm punting on that change because updating current old StatelessTransaction to do that reliably seems like a pain.

We probably want the caller to take care of _all_ txn object properties anyway? I'm imagining a generic `completeTxnParams()` that looks at `network` and `wallet` and figures out what needs to be added to the txn object that isn't already there. This way, the individual action pages/callers only have to worry about constructing a bare-bones tx and setting its gas limit. Sane?

Fang-:
Wait, no, don't believe anything I just said. Turns out I was still running with a way old version of Ethereumjs-tx. Will report back again after testing latest!

shrugs:
> We probably want the caller to take care of _all_ txn object properties anyway? I'm imagining a generic `completeTxnParams()` that looks at `network` and `wallet` and figures out what needs to be added to the txn object that isn't already there. This way, the individual action pages/callers only have to worry about constructing a bare-bones tx and setting its gas limit. Sane?

that's what my `useEthereumTransaction` hook will do, yeah — and then I can shove the unsigned tx object into `signTransaction` as planned 👍

Fang-:
Hmm, upgrading to v2 ethereumjs-tx causes the gas tank (which is still running v1) to trip on transaction decoding, getting completely incorrect values. Tried using v2 there also, but ran into ethereumjs/ethereumjs-tx#169, which makes this seem _way_ more trouble than it's worth. I'm perfectly fine keeping the manual EIP-155 support (and ethereumjs-tx v1) if it keeps us from having to deal with this close to launch.

@shrugs, if you agree, I'll go ahead and merge.

shrugs:
sgtm, ship it 👍

PR #194, opened 2019-07-16
Pass number instead of string to azimuth-js

Fang-:
This worked at some point. Pretty sure we're flip-flopping types all over the place.


shrugs:
is `issuingPoint` a string like '123'? if so, let's `parseInt(issuingPoint, 10)` instead of casting it to number. just a bit more resilient/explicit/standard. This is stuff that BN.js will help us with. lgtm otherwise

PR #197, opened 2019-07-18
Fix issues around networking key derivation

Fang-:
Noticed that we didn't account for the passphrase during key derivation. The Urbit wallet already contains the passphrase in its metadata, but for the mnemonic wallet case we add an extra field to store it.

### DIFF
--- a/src/lib/reticket.js
+++ b/src/lib/reticket.js
@@ -33,7 +33,7 @@ export async function reticketPointBetweenWallets({
   contracts,
   onUpdate,
   transferEth = false,
-  nextNetworkKeyRevision = INITIAL_NETWORK_KEY_REVISION,
+  nextRevision = INITIAL_NETWORK_KEY_REVISION,
 }) {
   const askForFunding = (address, minBalance, balance) =>
     onUpdate({
@@ -66,10 +66,7 @@ export async function reticketPointBetweenWallets({
   transferTmpTx.gas = GAS_LIMITS.TRANSFER;
 
   // configure networking public keys
-  const seed = await deriveNetworkSeedFromUrbitWallet(
-    toWallet,
-    nextNetworkKeyRevision
-  );
+  const seed = await deriveNetworkSeedFromUrbitWallet(toWallet, nextRevision);
 
   const networkSeed = seed.matchWith({
     Nothing: () => {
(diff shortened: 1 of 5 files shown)


Fang-:
It still gives me "keys dont match" when I log in with ~palfun-foslup's management key on the mainnet though, so something else about this is incorrect too... Investigating!

Fang-:
Okay, couple things that are messed up here.

First, we don't match old Bridge logic for deriving networking keys. It would use the `revision - 1` for the current keys, and `revision` for next. Changing this in Mino Bridge makes it successfully derive ~palfun-foslup's keyfile.

https://github.com/urbit/bridge/blob/cb2d061f6d123103cc1bd413287f6eba6d5fe12b/src/bridge/lib/keys.js#L63-L64
https://github.com/urbit/bridge/blob/cb2d061f6d123103cc1bd413287f6eba6d5fe12b/src/bridge/lib/keys.js#L92-L95

Either we match that, or we stick to our guns but fall back to deriving at `revision-1` if that doesn't work. I'd prefer the former, since it's less cruft to carry around into the future. Thinking about "revision" being the n in "nth key in zero-indexed array of keys" seems not all that painful. Though I can be persuaded to think otherwise.

Secondly, `/lib/reticket` always derives keys at `INITIAL_NETWORK_KEY_REVISION`. But in the reticketing-existing-point case, it's possible that our revision is non-zero, in which we want to derive at a different revision, so that Bridge can re-derive the keys after reticketing.

I'll commit proposed fixes for these in a little bit. @shrugs please verify that they're sane both in implementation and how they relate to existing usage, and feel free to propose alternatives.

src/lib/reticket.js
```diff
@@ -33,6 +33,7 @@ export async function reticketPointBetweenWallets({
   contracts,
   onUpdate,
   transferEth = false,
+  nextNetworkKeyRevision = INITIAL_NETWORK_KEY_REVISION,
```
shrugs: another function somewhere takes `revision` as the argument name — can we standardize on `networkKeyRevision` or `revision` for arguments?

Fang-:
>tbh I have no idea what revision is supposed to be, data-mode-wise

It's just a part of the "derivation path" that leads to your networking keys, so can technically be any value. We try to be consistent about it in a somewhat-logical fashion so that we don't have to randomly guess until we get it right.

PR #199, opened 2019-07-18
Use latest urbit-sigil-js

Fang-:
Just PRing against indigo-admin at this point, it's kinda the de-facto dev branch now. If this goes on much longer I'll have to make a "how many dev branches are you on" meme.

We flip the colorways because apparently new sigil-js flips how it interprets them. Removed the try-catch because this shouldn't throw in our usage. If it does, that's a red flag and _should_ be loud.


shrugs:
1. let's rebase this commit off of mino once I've merged admin
2. the new renderer correctly sets viewBox but still manually sets width and height on the inline styles, which means I can't easily override them and we get non-responsive sigils

responsive:
image
not responsive:
image

shrugs:
also, we can, I think, remove the ReactSVGComponents and the other occurrences of `sigil-js`, since we safely never touch any of those routes anymore (and I think most usages have been nuked as of the admin merge)

Fang-:
> also, we can, I think, remove the ReactSVGComponents and the other occurrences of `sigil-js`

I'll make a separate PR for nuking all remaining old code.

> the new renderer correctly sets viewBox but still manually sets width and height on the inline styles, which means I can't easily override them and we get non-responsive sigils

This seems like it might be an issue for the sigil-js repo? We do seem to consistently pass a `size` of `64`. Maybe we want to do something different there? Though, it does also seem to render sigils smaller within a same-size canvas by default.

shrugs:
yeah, sigil-js should not be setting inline styling. honestly it may not even need a size parameter for the svg renderers, since they're vector by default?

g-a-v-i-n:
feel free to @ me we questions about the library.

Fang-:
@g-a-v-i-n do new sigils render as smaller within their canvas, or is that just me?

g-a-v-i-n:
They do, but if you need a specific size you should override the margin with {margin:0} and use css for padding/margin.

shrugs:
are the inline styles applied by the library? the I'd modified the
ReactSVGRendered in bridge to set the viewbox to the correct dimensions and
the actual inline styles to height:100%; width:100% to have it responsively
expand to its container. at first glance that doesn't seem to be (easily)
possible with the new lib since the inline styles are set and we'd have to
use !important to get an override (afaict).

-matt

g-a-v-i-n:
@shrugs Can you tell me more about your usecase? Feel free to make an issue on https://github.com/urbit/sigil-js

shrugs:
@g-a-v-i-n I've made an issue documenting exactly what I mean: https://github.com/urbit/sigil-js/issues/37

shrugs:
also cleaned up some components and nuked the walletgen components since

1. they're totally unused
2. we won't want to bother refactoring them to use the new sigil-js
3. if we want to reference the old code we can rewind in git to see it, no use keeping it around

Fang-:
image
Ah, that's the stuff. (^:

Thanks for the changes! Good to merge once the update's in afaic.

Fang-:
Alright, this matches the size we were at previously +/- a pixel.

image

@g-a-v-i-n since not passing in a `style: { height, width }` results in those values being defaulted to the `size`, and we don't want to actually set those values here (I don't think), I resorted to passing empty strings for both. Is this the way to do it, or is there a better/cleaner way? Please confirm/advise.

shrugs:
@Fang- you'll want to pass `100%` for each

PR #202, opened 2019-07-19
Remove old components and views

Fang-:
I think these are the 2.2k deletions you were looking for, @shrugs. (;

Points overview page was still using an old component. Seeing as how you'll be tackling that soon, I just copy-pasted a simple version of the old thing in for now.


shrugs:
satisfying. haven't run the code yet, but if it builds that's good enough for me, LGTM

Fang-:
It builds and runs without issue on my end. (:

PR #204, opened 2019-07-22
Feat/multi home

shrugs:
implements multi-home and incoming/outgoing transfer accept/reject/cancel views

TODO: what exactly is the behavior we want for rejecting transfers? hiding locally or unsetting transfer proxy? I get a revert when attempting to set the transfer proxy to 0 (as the current transfer proxy), so seems to be local hide?

otherwise this is good for review


src/views/RejectTransfer.js
```diff
@@ -0,0 +1,93 @@
...
+  const { pointCursor } = usePointCursor();
+  const { syncDetails, syncControlledPoints } = usePointCache();
+
+  const _contracts = need.contracts(contracts);
+  const _point = need.point(pointCursor);
+
+  const { construct, ...rest } = useEthereumTransaction(
+    useCallback(
+      () =>
+        azimuth.ecliptic.setTransferProxy(_contracts, _point, ETH_ZERO_ADDR),
```
Fang-: > I get a revert when attempting to set the transfer proxy to 0 (as the current transfer proxy)

Right. Azimuth will eventually want to support this, for this and similar use cases, but doesn't allow proxies setting proxies right now. Locally hiding is fine, looks like it's time to build that better `/lib/cookie`...

Alternatively, we could transfer it to the current owner, with the `reset` arg as `true`, but I'm not sure if that's more desirable.

src/components/Passport.js
```diff
@@ -94,3 +109,71 @@ export default function Passport({
     </Grid>
   );
 }
+
+// TODO: Passport.Mini should probably accept Maybe<> for point
+// in order to display a pretty skeleton loading view
+// but that requires making usePermissionsForPoint accept a Maybe<point>
```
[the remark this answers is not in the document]
shrugs: the usePermissionsForPoint assumes that point always exists, specifically, and I didn't want to refactor it

src/views/RejectTransfer.js
shrugs: @Fang- can you ask Alex/Jimmy in today's meeting? I won't make it. I'll either import or write a simple local storage hook and that'll work just fine.

src/components/Passport.js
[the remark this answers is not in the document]
Fang-: I get that. I meant something like:

```js
point.matchWith({
  Nothing: () => 'Loading',
  Just: p => buildKeyType(usePermissionsForPoint(address, p.value)),
})
```

But maybe there's complications here I'm not seeing?

shrugs:
ah, that's a conditional hook, which isn't allowed by react

shrugs:
added our local ignore list using a simplified useLocalStorage hook from two other libraries (avoided dependency and simplified some logic we wouldn't be using

src/lib/useEthereumTransaction.js
```diff
@@ -174,7 +173,6 @@ export default function useEthereumTransaction(
   ]);
 
   const reset = useCallback(() => {
-    setUnsignedTransaction(undefined);
```
Fang-: Why are we no longer clearing this on-reset?

Fang-: And should we maybe clear `nonce` here as well, so we fetch a fresh one? Very edge case, but can't hurt.

src/lib/useLocalStorageState.js
```diff
@@ -0,0 +1,38 @@
+import { useState, useEffect } from 'react';
+
+/**
+ * useLocalStorageState is useState but it persists to localStorage
+ *
+ * inspired by:
+ * https://github.com/streamich/react-use/blob/master/src/useLocalStorage.ts
+ * https://usehooks.com/useLocalStorage/
+ */
+export default function useLocalStorageState(key, initialValue) {
```
Fang-: Nice! Do you think it's worth updating the Disclaimer / hasDisclaimed to use this also, instead of dumb cookie usage?

src/views/RejectTransfer.js
Fang-: Oh snap, I forgot to ask for input on this. Sorry! But local storage seems sane and good.

src/lib/useEthereumTransaction.js
shrugs: good call on nonce

clearing the unsigned transaction put us in a weird state (especially for components that did something like `useLifecycle(() => construct(...))` because after a reset there was no unsigned transaction and lifecycle wouldn't run again.

src/lib/useLocalStorageState.js
shrugs: yes, will do that

src/views/Disclaimer.js
```diff
@@ -14,15 +14,16 @@ const TEXT_STYLE = 'f5';
 
 export default function ActivateDisclaimer() {
   const { pop } = useHistory();
+  const [, setHasDisclaimed] = useHasDisclaimed();
```
Fang-: wow

shrugs: javascript is a good language

PR #207, opened 2019-07-23
Remove the landing page

Fang-:
Displays login or activate directly, depending on context.

Simplifies the Disclaimer routing logic so it doesn't need to know what page to send the user to.


src/Bridge.js
```diff
@@ -23,16 +24,18 @@ const INITIAL_NETWORK_TYPE = isDevelopment
 // NB(shrugs): modify these variables to change the default local state.
 const SHOULD_STUB_LOCAL = process.env.REACT_APP_STUB_LOCAL === 'true';
 const IS_STUBBED = isDevelopment && SHOULD_STUB_LOCAL;
+const LANDING_PAGE = !!useImpliedTicket()
```
shrugs: it's kind of lucky that this 'just works' — we're lucky that this specific hook is also just a normal function with no other hook dependencies. Can we move the logic from the `useImplied{Ticket,Point}` hooks into a lib or something that we can use in both places without confusion?

shrugs: alternatively (and easier), this hook and the decision on routing can go into the render method of Bridge where it sets the initialRoutes

Fang-: Ah, your concern is that hooks may or may not be in a sane state outside of the render function?

I'll move routing into the function here.

Fang-: Or, I think more likely, that they won't change on re-render even if the underlying values changed?

shrugs: it will be in a sane state (because in this specific case our hook is _also_ a plain old javascript function) but we're still crossing patterns here, since hooks by design are only called in the render method of your component. so to make that distinction explicit we should either use a _specifically_ plain old js function like

```
const getImpliedPoint = () => {...}

...

function useImpliedPoint() {
  return getImpliedPoint();
}

...

const INITIAL_ROUTES = !!getImpliedPoint() : ... ? ...
```

or just use the hook in the render function. this is more for code readers than react or javascript, since (in this specific case only) there's no difference.

Issue #208, opened 2019-07-23
has disclaimed fix


shrugs:
the has disclaimed update isn't actually functional. with the following code

```js
  const goBack = useCallback(() => {
    setHasDisclaimed(true);
    pop();
  }, [pop, setHasDisclaimed]);
```

what actually happens is:

1. `setState` in `useHasDisclaimed` fires synchronously (great 👍 )
2. `pop()` unmounts the current component (great 👍 )
3. react cancels the `useEffect` from `useHasDisclaimed` before the new value is written to local storage (bad 🙅‍♂)

solution 1:

```js 
import timeout from 'lib/timeout'

const tick = () => timeout(0)

...

const goBack = useCallback(async () => {
    setHasDisclaimed(true);
    await tick();
    pop();
  }, [pop, setHasDisclaimed]);
```

which lets the effects be flushed before the pop unmounts them

solution 2:
a synchronous setState that wraps the normal `setState`

---

after writing all this out, I'm going to do option 2 cause it's the least leaky and we're talking nanoseconds at worse, afaik

Fang-:
Good god, what a minefield. ):

Nice catch, agreed that solution 2 seems better!

shrugs:
closed in #209

PR #209, opened 2019-07-23
feat: design pass updates

shrugs:
fixes https://github.com/urbit/bridge/issues/208


src/views/Admin/AdminNetworkingKeys.js
```diff
@@ -330,7 +330,8 @@ export default function AdminNetworkingKeys() {
         {completed && (
           <>
             <Grid.Item full as={NoticeBox} className="mb3">
-              You need this keyfile to authenticate with Arvo.
+              You need this keyfile to authenticate with Arvo. If you have
+              already booted Arvo, reboot with this new keyfile.
```
Fang-: This is only true in the `discontinuous` case.

To use new keys while remaining continuous, you use the `|rekey` command, pasting in the contents of your keyfile as a string. ie `|rekey '0w123.whatever`

That has its own slew of issues (it's kinda strict about when/how you can change keys), but certainly better than telling people to reboot their ships.

Proposed copy: "If you have already booted Arvo, run `|rekey '0wkeyfile.contents'`

I'd include a link to the docs, but that doesn't include this at all. cc @rmariani @baudtack

shrugs: I'm going to update to that copy but this needs more thought around the whole set of flows and documentation etc to be really handled.

src/lib/useInputs.js
```diff
@@ -22,7 +22,12 @@ import { prependSig, convertToNumber } from 'lib/transformers';
 
 import InputSigil from 'components/InputSigil';
 
-const EXAMPLE_PRIVATE_KEY =
+const PLACEHOLDER_POINT = '~zod';
+const PLACEHOLDER_HD_PATH = DEFAULT_HD_PATH;
+const PLACEHOLDER_MNEMONIC = bip39.generateMnemonic();
+const PLACEHOLDER_TICKET = '~ragtyd-modwen-bostec-hinsep';
+const PLACEHOLDER_ADDRESS = '0x6DEfFb0caFDB11D175F123F6891AA64F01c24F7d';
+const PLACEHOLDER_PRIVATE_KEY =
```
Fang-: Proposed sample values:

```js
const PLACEHOLDER_POINT = '~sampel-ponnym';
const PLACEHOLDER_HD_PATH = DEFAULT_HD_PATH;
const PLACEHOLDER_MNEMONIC =
  'example crew supreme gesture quantum web media hazard theory mercy wing kitten';
const PLACEHOLDER_TICKET = '~sampel-ticlyt-migfun-falmel;
const PLACEHOLDER_ADDRESS = '0x12345abcdeDB11D175F123F6891AA64F01c24F7d';
const PLACEHOLDER_PRIVATE_KEY =
  '0x12345abcdee6beb2f323fab48b432925c9785808d33a6ca6d7ba00b45e9370c3';
```

shrugs:
@Fang- not at my computer, can you make those changes and merge? lgtm

PR #210, opened 2019-07-24
Mino more fixes

Fang-:
Bunch of small QoL ones again. See commit messages for the full list.

I went and took the liberty of expanding `0xab...1234` into `0xabcd...1234`. This isn't _exactly_ according to design, but seems much more useful.

@shrugs can you take a look at that last, WIP commit? The result that's returned with `return validateInSet(` is correct and sane, but it's not getting reflected in the UI anymore. What am I missing?

(Targetting your dev branch because that has slightly overlapping work. Happy to hold off on merging until that gets in, or whatever's appropriate.)


src/views/IssueChild.js
```diff
@@ -54,7 +54,7 @@ export default function IssueChild() {
   const { pointCursor } = usePointCursor();
 
   const _contracts = need.contracts(contracts);
-  const _point = need.point(pointCursor);
+  const _point = parseInt(need.point(pointCursor), 10);
```
shrugs: pointCursor should _always_ be `Maybe<number>` — where is it being set as a string?

src/views/IssueChild.js
```diff
@@ -75,12 +75,19 @@ export default function IssueChild() {
 
   const validators = useMemo(
     () => [
-      validateInSet(
-        availablePointNames.getOrElse(new Set()),
-        'This point cannot be spawned.'
-      ),
+      m =>
+        simpleValidatorWrapper({
```
shrugs: I would move this into `validators.js` and probably not try and chain the validators like this — I find it hard to confirm that this does what we're expecting it to. I expect that when you move it to validators and pattern-match the existing logic the bug will resolve itself.

src/views/IssueChild.js
Fang-: We weren't consistently using `lib/patp2dec` over `ob.patp2dec`. I've replaced all instances of the latter with the former, for consistency's sake.

src/components/Passport.js
```diff
@@ -142,7 +142,7 @@ Passport.Mini = function MiniPassport({ point, className, inverted, ...rest }) {
           patp={name}
           size={64}
           margin={0}
-          colorway={inverted ? INVERTED_COLORWAY : undefined}
+          colors={inverted ? INVERTED_COLORWAY : undefined}
```
shrugs: was this not fixed in the sigil branch?

Fang-: The branch this PR is coming from contains mino, which contains the sigil branch, so no. I overlooked this.

PR #213, opened 2019-07-29
InviteEmail hacks

Fang-:
Still haven't managed to fix the "live async email checking" thing. It's weird. When I include `getHasReceived` as a hook dependency, the hook fires when that changes, but doesn't include the new value.

This PR contains two smaller fixes to the user flow for InviteEmail that should make the lack of the above more bearable.

### DIFF
--- a/src/views/Invite/InviteEmail.js
+++ b/src/views/Invite/InviteEmail.js
@@ -413,6 +413,7 @@ export default function InviteEmail() {
       setGeneralError(new Error(`Duplicate email.`));
     } else {
       setGeneralError(null);
+      setStatus(STATUS.INPUT);
     }
   }, [inputs]);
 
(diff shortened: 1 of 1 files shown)


src/views/Invite/InviteEmail.js
```diff
@@ -189,24 +189,30 @@ export default function InviteEmail() {
...
     //TODO want to do this on-input, but that gets weird. see #188
-    let errorCount = 0;
+    let knowAll = true;
+    let alreadyReceived = [];
     await Promise.all(
       inputs.map(async input => {
         const email = input.data;
         const hasReceived = getHasReceived(email).matchWith({
-          Nothing: () => false, // loading
+          Nothing: () => 'unknown', // loading
```
shrugs: I would just do something like

```
const knowAll = every(inputs.map(input => getHasReceived(input.data).getOrElse(false))
const alreadyReceived = inputs.map(input => getHasReceived(input.data).getOrElse(false) ? input.data : null).filter(x => !!x);

...
```

Fang-: We discussed (out of band) doing this with `Result` instead, but both that and the current solution just end up taking needless extra steps. Pushed a version that gets to the point more directly.

PR #216, opened 2019-07-30
Design Pass 2: Design Passer

shrugs:
see commits + asana 'Fixed' column for deets


shrugs:
@Fang- I'm going to do all of the form stuff in a separate PR, so this is ready for review

src/lib/useRouter.js
```diff
@@ -52,15 +50,27 @@ export default function useRouter({
...
-        return setRoutes(routes => [...routes.slice(0, routes.length - count)]);
-      }
+      if (size <= 1) {
+        // if we are at the root, pass this event to our parent
+        if (oldPopState.current) {
+          console.log('attempting to pass to parent');
+          window.history.back();
+        }
+
+        console.log('at root');
```
Fang-: These `console.log`s should probably go?

src/views/ViewPoint.js
```diff
@@ -1,19 +1,19 @@
...
 
-import { ROUTE_NAMES } from 'lib/routeNames';
 import patp2dec from 'lib/patp2dec';
 import { usePointInput } from 'lib/useInputs';
 
 import View from 'components/View';
 import { ForwardButton } from 'components/Buttons';
+import ViewHeader from 'components/ViewHeader';
 
 export default function ViewPoint() {
```
Fang-: This isn't routed to from anywhere anymore, I think we can rip this out? (Missed this during my teardown.)

PR #217, opened 2019-07-30
add old PCR fonts

g-a-v-i-n:
- add work sans
- add SF Pro


g-a-v-i-n:
I have resolved the build issues in PCR and updated the fonts to the new fonts, this is no longer needed.

Fang-:
Ah, what timing. Was about to tell you the following, but I think that's obsolete then.

image

In both Firefox and Chrome. I cleared out the `node_modules` and did a fresh `npm install` beforehand.

matildepark:
Times New Roman! It's so classy now!

alexmatz:
this PR is dead. @Fang- can you update paper collateral renderer on Bridge and see if that fixes the font issue on 1.1.2?

Fang-:
Continuing comms over email now for slightly less overhead. For the public record: fonts still broken in 1.1.2.

PR #218, opened 2019-07-30
Feat/ga fonts

g-a-v-i-n:
If the font had been loaded and it is cached, there is no font load time so the font will be available immediately when PaperCollateralRenderer runs. Since bold Inter is not used in bridge, it is never loaded and therefore needs to be loaded. During the load time, the page renders without the font as it does normally in most webpages in the world. During this time, the font is not available, so Firefox does not seem to render it. This can happen to any font that has not been loaded.

I have base64 encoded the fonts and inserted them into a stylesheet called font64.scss. This should mean the fonts are loaded with the HTML immediately, before anything happens with React.

In order to test this, you must open font book and disable Inter, Inter UI and Source Code Pro.


g-a-v-i-n:
In 04167ab, I added fonts but then removed them in a8fe151

g-a-v-i-n:
Try clearing browser caches too @Fang-

Fang-:
I always cache-refresh before testing! That's bitten me too many times in the past.

PR #219, opened 2019-07-31
Replace mentions of Multipass

Fang-:
@alexmatz pretty sure we're sticking with Bridge for now, right?


Fang-:
Confirmed as a-ok by @alexmatz out of band.

PR #220, opened 2019-07-31
Don't display disclaimer prior to activation code use

Fang-:
Moves it to before the "download passport" step.
Still displays it prior to login.

### DIFF
--- a/src/Bridge.js
+++ b/src/Bridge.js
@@ -48,13 +48,12 @@ function useInitialRoutes() {
     ];
   }
 
-  const landingPage = hasImpliedTicket
-    ? ROUTE_NAMES.ACTIVATE
-    : ROUTE_NAMES.LOGIN;
-
-  const extraRoutes = !hasDisclaimed ? [{ key: ROUTE_NAMES.DISCLAIMER }] : [];
-
-  return [{ key: landingPage }, ...extraRoutes];
+  return hasImpliedTicket
+    ? [{ key: ROUTE_NAMES.ACTIVATE }]
+    : [
+        { key: ROUTE_NAMES.LOGIN },
+        ...(!hasDisclaimed ? [{ key: ROUTE_NAMES.DISCLAIMER }] : []),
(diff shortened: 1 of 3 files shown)


src/Bridge.js
```diff
@@ -48,13 +48,12 @@ function useInitialRoutes() {
...
-    : ROUTE_NAMES.LOGIN;
-
-  const extraRoutes = !hasDisclaimed ? [{ key: ROUTE_NAMES.DISCLAIMER }] : [];
-
-  return [{ key: landingPage }, ...extraRoutes];
+  return hasImpliedTicket
+    ? [{ key: ROUTE_NAMES.ACTIVATE }]
+    : [
+        { key: ROUTE_NAMES.LOGIN },
+        ...(!hasDisclaimed ? [{ key: ROUTE_NAMES.DISCLAIMER }] : []),
```
shrugs: was this reverted or something? left version is much more readable

src/views/Activate/ActivateCode.js
```diff
@@ -48,7 +50,10 @@ export default function ActivateCode() {
   const goToLogin = useCallback(() => history.popAndPush(ROUTE_NAMES.LOGIN), [
     history,
   ]);
-  const goToPassport = useCallback(() => push(names.PASSPORT), [names, push]);
+  const goToPassport = useCallback(() => {
+    push(names.PASSPORT);
+    if (!hasDisclaimed) push(names.DISCLAIMER);
```
shrugs: nit: curly braces for simple branches are used basically everywhere else

src/Bridge.js
Fang-: This is doing something different though. It's only adding the disclaimer to the route for the login case, not the activate case. The previous version was adding it for both.

The js too fancy here?

shrugs: ah, didn't notice the new logic. it's fancy, but I'd say it's a-ok

-matt

PR #221, opened 2019-07-31
The Final Form PR

shrugs:
Thoughts, @Fang-? should support our InviteEmail use-case along with hasReceived, etc as a async validator. had some problems with final-form (re: email and some un-told), so any weird architecture you see might be because of those. lmk what you think. definitely 'feels' better to me.


src/views/Activate/ActivateCode.js
```diff
@@ -1,150 +1,166 @@
...
+  // validate should be a pure function but we don't want to have to recompute
+  // all of this information on submit, so cache the invite wallet and avoid
+  // re-renders that may trigger re-validations (causing infinite loop)
+  const validateForm = useCallback(
+    async ({ ticket }, errors) => {
+      if (hasErrors(errors)) {
+        return errors;
+      }
+
+      console.log('FUCK — deriving seed');
```
Fang-: lol

src/views/Activate/ActivateCode.js
```diff
@@ -1,150 +1,166 @@
...
+
+      await timeout(100); // allow the ui changes to flush before we lag it out
 
-  useEffect(() => {
-    if (validTicket) {
       const _contracts = need.contracts(contracts);
+      const { seed } = await generateTemporaryOwnershipWallet(ticket);
+
+      // TODO(fang): isn't all this accessible in the ownership object?
```
Fang-: I thought I touched this up already? We can just use `generateResult.keys.address` etc. But Maybe that structure doesn't quite match what's currently in use.

src/views/Login/Ticket.js
```diff
@@ -1,187 +1,209 @@
...
+        // shards
+        try {
+          const ticket = kg.combine([
+            values.shard1,
+            values.shard2,
+            values.shard3,
+          ]);
+          const point = patp2dec(values.point);
+          cachedUrbitWallet.current = await urbitWalletFromTicket(
+            ticket,
```
Fang-: I think we can deduplicate this logic with the ticket flow? Looks like the ticket flow does `isOwner, isTransferProxy` checks also, which this doesn't do for some reason.

src/views/Activate/ActivateCode.js
shrugs: this is just a modification of the comment that was already there — cargo-culted it — will remove

shrugs:
I don't think there's anything wrong with not submitting the forms that we're creating, but we're technically not fulfilling the form API as part of the onValues type interactions (namely when also using InlineEthereumTransaction). I expect there's a nice clean API that can be use when creating a form that will also be submitted as an Ethereum transaction, but imo that's out of scope here since this is big enough as-is.

shrugs:
(i.e. returning a promise from `onSubmit` that resolves when the ethereum tx is confirmed and then using the state from final-form to handle some logic for presentation instead of using `useEthereumTransaction#completed`. idk, just spitballing at 10pm

Fang-:
> I don't think there's anything wrong with not submitting the forms that we're creating, but we're technically not fulfilling the form API as part of the onValues type interactions

I don't think I understand what this means. I might be missing some context here on what we _should_ be doing.

shrugs:
I don't actually know if we 'should' be submitting the forms or not, only that for our forms that use the live 'onValues' type api, they're never submitted. this probably isn't a problem, but worth noting that the form library may expect that the form is submitted at some point (but I don't think we use any of its features around submission, which is why this isn't a big deal—more of a 'heads up' to myself, really)

src/lib/useRouter.js
```diff
@@ -53,7 +53,7 @@ export default function useRouter({
       if (size <= 1) {
         // if we are at the root, pass this event to our parent
         if (oldPopState.current) {
-          window.history.back();
+          oldPopState.current();
```
shrugs: note: the reason for this change is that when we click the "return" button (which purposefully pops a route) we also inject a new `pushRoute` entry, which means the back button stops working. so instead of using the browser back button (which was a dumb shortcut, anyway), we ask the parent router to pop itself directly.

shrugs:
I'm not sure how to do warnings. The example react-final-form gives is https://codesandbox.io/s/m5qwxpr6o8 which makes a ton of sense, but requires access to the form api object, which we don't have in the validation functions (which is where the warnings would be generated).


one options is to just set local state and then use a hook that's like

```
function useWarnings(warnings) {
  const form = useForm();
  useEffect(() => {
    map(warnings, (warning, name) => form.mutators.)
  }, [form])
}
```

shrugs:
(accidentally closed instead of cancelling a comment I wrote lol)

src/views/Activate/ActivateCode.js
```diff
@@ -1,157 +1,152 @@
...
+      );
+      const transferring = await azimuth.azimuth.getTransferringFor(
+        _contracts,
+        _inviteWallet.address
+      );
+      const incoming = [...owned, ...transferring];
+
+      if (incoming.length > 0) {
+        if (incoming.length > 1) {
+          // TODO: putting a warning here doesn't make sense since the user
```
shrugs: I moved the validation here to onSubmit rather than in realtime, so now this warning doesn't make sense to display — what do?

src/views/Admin/AdminNetworkingKeys.js
```diff
@@ -335,20 +321,48 @@ export default function AdminNetworkingKeys() {
...
+                  name="useNetworkSeed"
+                  label="Use Custom Network Seed"
+                  inverseLabel="Back to Derived Network Seed"
+                  disabled={inputsLocked}
+                />
+                <Condition when="useNetworkSeed" is={true}>
+                  <Grid.Item full as={NoticeBox} className="mb2">
+                    When using a custom network seed, you'll need to download
+                    your Arvo keyfile immediately after this transaction is
+                    completed—Multipass does not store your seed.
```
Fang-: s/Multipass/Bridge

src/form/Inputs.js
```diff
@@ -28,6 +30,7 @@ const PLACEHOLDER_TICKET = '~sampel-ticlyt-migfun-falmel';
 const PLACEHOLDER_ADDRESS = '0x12345abcdeDB11D175F123F6891AA64F01c24F7d';
 const PLACEHOLDER_PRIVATE_KEY =
   '0x12345abcdee6beb2f323fab48b432925c9785808d33a6ca6d7ba00b45e9370c3';
+const PLACEHOLDER_EMAIL = 'Email Address';
```
Fang-: Should probably be something like `email@example.com` to be more in line with the other examples.

src/form/Inputs.js
```diff
@@ -54,12 +57,11 @@ export const buildCheckboxValidator = mustBe =>
 export const buildPassphraseValidator = () => buildValidator([]);
```
Fang-: Maybe a bit off-topic, but, didn't we have a `lib/validator`? Shouldn't all of these be in there?

src/views/Admin/AdminTransfer.js
```diff
@@ -47,46 +53,64 @@ export default function AdminTransfer() {
...
+        {({ handleSubmit, values }) => (
+          <>
+            <Grid.Item
+              full
+              as={Text}
+              className={cn('f5', {
+                green3: completed,
+              })}>
+              {completed
+                ? `${values.address} is now the Transfer Proxy for ${name} and can accept the transfer by logging into Multipass themselves. Until they accept your transfer, you will still have ownership over ${name}.`
```
Fang-: s/Multipass/Bridge again. You're probably fine to pull in mino now that #219 is merged and find-replace the remaining instances.

src/form/Inputs.js
Fang-: (fixed before I even reviewed it, truly the fastest dev (^:)

src/components/InputSigil.js
```diff
@@ -1,19 +1,18 @@
 import React, { useEffect, useState } from 'react';
 import { Just } from 'folktale/maybe';
-import * as ob from 'urbit-ob';
 
 import MaybeSigil from './MaybeSigil';
 
-const selectColorway = (pass, fail, focused) => {
-  if (pass) {
+const selectColorway = (valid, error, active) => {
```
Fang-: +1 on this change! (:

src/form/Inputs.js
```diff
@@ -41,6 +41,9 @@ export function TicketInput({ name, ...rest }) {
       type="password"
       name={name}
       placeholder={PLACEHOLDER_TICKET}
+      autoCapitalize="none"
```
Fang-: Nice, +1 for accessibility development! But, we haven't paid much attention to that during development at all, and I wouldn't be surprised if there are things in Bridge that are completely messed up for screenreaders and what have you. ):

shrugs:
> Reviewing these is actually really hard.

I hear you on this, thanks for your patience so far — this is something I'm aware that I'm terrible at, but haven't put enough effort into becoming better at. I'll be more conscious about this going forward!

src/form/Inputs.js
shrugs: figma uses 'Email Address', but probably just isn't updated

cc @jyng for input

src/views/Admin/AdminTransfer.js
shrugs: merged mino & fixed 👍

shrugs:
> In ticket inputs, the sig only gets auto-prepended after the second keypress/character. Also, if you paste a ticket, it doesn't auto-prepend a sig if it isn't already there. It used to do both of these, might have gotten broken during the "allow deleting single ~" change?

definitely what broke it — I guess I'll just revert that? I changed that because my friend was concerned as to why he couldn't delete the last character in the password field (it's the tilde but just shows as the dot). any preference?

shrugs:
> I can't see to login anymore? It's stuck on the above, where previously it would resolve and let me press the button. No console output, nothing.

that looks like the button is stalled within the `validateForm` function, almost definitely on https://github.com/urbit/bridge/blob/feat/final-form/src/views/Login/Ticket.js#L80 — any idea why that wouldn't resolve quickly? sometimes my local chain can take up to like 10 seconds if it's been running for a while but that doesn't seem to be what you're seeing.

shrugs:
> This error doesn't clear if I change the input field's contents, so I can't correct a typo or re-paste the thing or whatever:

looks like this is because `submitErrors` are different from validation errors (I previously assumed these were identical) and are not cleared when validation passes. Seems like odd (intentional) behavior to me. here's the react-final-form issue about it: https://github.com/final-form/react-final-form/issues/89

shrugs:
> Is that brief showing deliberate?

sorta.... I just left the original logic for showing the derived point alone, but since we only know the point on submit now, it shows in the span between derivation and actual routing.

src/views/Login/Ledger.js
```diff
@@ -77,10 +77,10 @@ export default function Ledger({ className, goHome }) {
         const chainCode = Buffer.from(info.chainCode, 'hex');
         const pub = secp256k1.publicKeyConvert(publicKey, true);
         const hd = bip32.fromPublicKey(pub, chainCode);
+
         setWallet(Just(hd));
         setWalletHdPath(addHdPrefix(values.hdpath));
       } catch (error) {
-        setWallet(Nothing());
```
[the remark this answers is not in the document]
shrugs: because we now set the wallet on submit after all validation passes, there's no situation in which we `setWallet(Just(...))` and then need to clear it.

src/lib/txn.js
```diff
@@ -129,13 +129,19 @@ const sendSignedTransaction = (web3, stx, doubtNonceError) => {
...
     const confirmed = receipt !== null;
-    if (confirmed) return receipt.status === true;
-    else throw new Error('retrying');
+    if (!confirmed) {
+      throw new Error('Transaction not confirmed.');
+    }
+
+    const success = receipt.status === true;
+    if (!success) {
+      return bail(new Error('Transaction failed.'));
```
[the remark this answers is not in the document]
shrugs: agreed that it's very unclean 😢  I'll get better at this.

I was refactoring InviteEmails and we were checking the boolean response there, but in no other callsites (swallowing those errors) so I just made it do what we expected it to do.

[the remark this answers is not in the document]
shrugs: we probably expected it to throw because we're using it in a flow like

```
... construct transaction

... send transaction

... wait for transaction

return
```

all of which is (probably?) in a trycatch somewhere, and a failure at any point in that flow is treated as a failure of the flow as a whole. in every case where we'd check `if (success) {...}` we'd just throw a new error anyway

src/form/Inputs.js
shrugs: yeah our accessibility is absolute garbage right now — i wasn't paying any attention. you can use chrome's a11y inspector to see that things are broken, especially around form stuff

shrugs:
> Hmm good Q. Couple options, and we should probably just go with whatever's easiest to implement. This is a very edge-y case.

 > Display the message on this page, and require the user to hit continue a second time to proceed.

seems right to me, and definitely cleanest...

although the solution to the submit errors problem and this one is also just debounced validation, which is an open issue: https://github.com/final-form/react-final-form/issues/369

src/lib/txn.js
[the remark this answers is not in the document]
Fang-: > all of which is (probably?) in a trycatch somewhere,
> in every case where we'd check if (success) {...} we'd just throw a new error anyway

I suppose that's fair from a refactoring standpoint, yeah.

Fang-:
> I'll be more conscious about this going forward!

Thank you! (:

> definitely what broke it — I guess I'll just revert that?

I actually really liked the change, an undeletable dot is indeed pretty confusing! (I've watched some colleagues raise question marks about it.) I'd think that just doing "don't prepend on-update if the new value is empty" isn't that hard, but at this point that's just incredibly naive. (^:  
If you can find a way to fix instead of revert, that'd be great. If not, fine to revert, we eventually want the `~••••••-••••••`-style ticket input anyway. (Which we could maybe solve for using dynamically-added input fields for each new section, instead of one input with complex masking? idk)

> that looks like the button is stalled within the validateForm function, almost definitely on https://github.com/urbit/bridge/blob/feat/final-form/src/views/Login/Ticket.js#L80 — any idea why that wouldn't resolve quickly?

That's just a straight lookup, and should always resolve quickly (unlike event-based lookups). We've had those checks in forever and they've always worked fine, are we sure that's where they stall?

> Seems like odd (intentional) behavior to me. here's the react-final-form issue about it:

> the solution to the submit errors problem and this one is also just debounced validation, which is an open issue:

god damnit lol

shrugs:
update: async validation as a concept seems pretty broken in final-form: https://github.com/final-form/final-form/issues/147#issuecomment-519881470 — it means that our long-running validations will necessarily break when someone goes from a sync-valid value to an invalid one (the validation result from the old value will arrive later than the validation result for the new value, meaning final-form will report new values using old errors, breaking the spacetime continuum)

shrugs:
I'm now evaluating formik, since final-form seems to be rather foundationally broken for the things we want to do.

Formik is much cleaner, code-wise, but:

generally needs some things reimplemented
- format/parse behavior 

and has the following problems:

- no default `switchMap` behavior for validations
  - a hack solution is possible because throwing validation errors isn't a problem like it is in final-form. [my demo](https://codesandbox.io/s/adoring-boyd-zsfogwdqdwqd-ixugl)
- the `isValidating` prop isn't updated onChange/onBlur (??????) seemingly [because of render inefficiency](https://github.com/jaredpalmer/formik/issues/956#issuecomment-447173679) - https://github.com/jaredpalmer/formik/issues/1624 but this means that figuring out if the form is being validated is a non-standard process (wtf???????)
- no support for async onSubmit handlers, so we'd have to roll our own logic for InviteEmails because of its complex submission flow (but that's kinda ok cause that's what we were doing for the invite sending part of the final-form version anyway)

but also does support:

- onValues / ValuesHandler pattern, assuming isValidating is correctly updated [demo](https://codesandbox.io/s/adoring-boyd-zsfogwdqdwqd-ixugl)
- 

---

conclusion: everything is bad.

additional conclusion: I should have understood the problems we were facing more explicitly before starting this refactor (although tbh idk if I could have seen these problems until post-integration, but I definitely could have attempted to solve them before refactoring the entire app)

---

I have to step away from my desk — submitting this and planning on updating it further.

shrugs:
ok I've gone ahead and implemented that async validation-related feature on final-form and it's [in a PR](https://github.com/final-form/final-form/pull/265). will point the project to a local copy for now (which means you'll need to clone https://github.com/shrugs/final-form into a sibling folder and checkout the `feat/switch-map-validations` branch to get bridge to compile (once I've pushed that change)

Fang-:
> (which means you'll need to clone https://github.com/shrugs/final-form into a sibling folder

Can't we just point the package.json to your repo?

> conclusion: everything is terrible.

Would you agree with me that things would be less profoundly terrible if we just hand-rolled everything in plain Javascript? (^:

shrugs:
> Would you agree with me that things would be less profoundly terrible if we just hand-rolled everything in plain Javascript? (^:

i thought about it, but I'd still say probably not, since there is a lot of devilish complexity hidden in form behavior. but hard to tell lol

anyway, my PR was merged super quickly so I'll just point it back to the latest published version of final-form

Fang-:
> my PR was merged super quickly

Pretty valuable argument for sticking with final-form I guess!

shrugs:
ok, the last feature for this PR is the partial input masking, but I honestly can't find a damn thing on the internet about this, so I feel like I'm googling wrong, since sites definitely do this. The problem is that every "text masking" library you encounter is only for adding/removing specific characters according to your mask logic, not conditionally hiding them. So I'm not really sure  what to google to find what I'm looking for.

The approach seems to be to override `onChange` or `onInput` or `onKeyDown` and use that to both modify the "unmasked" value and send the change to the normal handlers as well , but the details feel difficult. Will take a crack at this post-sync, but lmk if you have any ideas in this direction, @Fang-

Fang-:
> since sites definitely do this

If you can find a site that does this sufficiently similarly to what we want, then we could look at their shipped source and see if we can clean any info from that? Long shot, but idk.

src/lib/isPromise.js
```diff
@@ -0,0 +1,6 @@
+// an object is a promise if it has a `.then` function
+// https://github.com/then/is-promise/blob/master/index.js
+export default obj =>
+  !!obj &&
+  (typeof obj === 'object' || typeof obj === 'function') &&
+  typeof obj.then === 'function';
```
Fang-: How can `obj` both be of type `function` _and_ have `obj.then`?

src/views/Login/Keystore.js
```diff
@@ -44,7 +44,7 @@ export default function Keystore({ className, goHome }) {
         const wallet = new EthereumWallet(privateKey);
         setWallet(Just(wallet));
       } catch (error) {
-        console.error(error);
+        // console.error(error);
```
Fang-: Maybe we should keep this in?

src/form/formatters.js
```diff
@@ -10,6 +11,35 @@ export const prependSig = s => {
   return s.charAt(0) === '~' ? s : `~${s}`;
 };
 
+/**
+ * inserts a - every 7 characters
+ * (must be preceeded by prependSig to work correctly)
```
Fang-: That seems a bit icky. Maybe want an `ensurePatQFormat` or something? Probably not all that bad to leave this as is though.

src/lib/isPromise.js
shrugs: classes are just functions and functions can have properties, love javascript. this is just is-promise, btw — not my doing.

src/form/formatters.js
shrugs: I was able to easily expand `ensurePatFormat` to cover the prependSig case as well

shrugs:
@Fang- which comments are still pending? my TODO looks like:

- [ ] figure out why your login was stalled (no leads yet)
- [ ] partial masked ticket input
- [ ] some sort of double-confirm behavior so that onSubmit warnings have time to be shown to the user.

shrugs:
I've added the double-confirm behavior where it will only double-confirm when there is a warning, which feels correct and honestly kinda nice. I've asked the dev for best practices around warnings, but this solution feels a-ok

ref: https://github.com/final-form/final-form-set-field-data/issues/3

shrugs:
an idea: we show plaintext on focus and asterisks on defocus (i.e., just changing the `type` from `text` to `password` based on focus events)

shrugs:
here's what that looks like 
Untitled

Fang-:
> which comments are still pending?

Hmm, I think we're good actually, with recent changes. Functional review incoming!

The on-focus unmasking seems nice, but I'd maybe vote against it still. [For reference.](http://www.passwordmasking.com/)

shrugs:
ah yeah, form inputs need to be disabled post submission. good catch.

the focus masking is on a separate branch, not on this one, so that's ok.
I'm very down to go with that website's result. only design question is
where the checkbox goes, so Jimmy would need to weigh in on that. can you
ping him in person and get some feedback? then we can wrap this up in
reasonable time.

I've experienced the hanging on mobile devices but not yet on a desktop.
you're using Firefox? how many emails were you generating, and how long did
you give it before you killed the tab? perhaps browsers slow down problem
threads, which may exacerbate the problem?

shrugs:
@Fang- actually, simply disabling the inputs after successful (but warned) submit isn't a real solution, since now the user can't go 'back' and edit their values.

I've changed the approach in the latest commit to use final-form's error feature instead of working around it. Now we have a custom `WARNING` error that we can return from a failed submit, which means final-form doesn't consider the form 'succeeded'. We keep track of whether or not we're double-submitting using the `didWarn` reference, so the second submission from a pristine (i.e. `!dirtySinceLastSubmit`) form will pass through correctly.

This does have the downside that our submit function is called twice, which means generation happens twice in the case that you want to auth with 'invalid' credentials or if your invite code has multiple points to be claimed, both of which are highly unlikely scenarios.

We also correctly update the state after you edit, so the warning will go away until the submission fails again.

I also added the `submitSucceeded` check on all of the inputs cause it's almost definitely the right thing to do anyway, since we never (?) want anyone editing the form after we've successfully submitted (we always either hide the inputs or redirect away anyway).

shrugs:
give it an  👀  and lmk what jimmy says about the masking toggle so I can implement that monday morning.

Fang-:
re: masking toggle: we should just do a "check to reveal" checkbox underneath the ticket input. This should be consistent on both login and ticket verification during activation & reticketing.

That is separate work for a separate PR though, this one just needs to be merged after this long. Same goes for your fix for #229. Smaller PRs is more better!

shrugs:
@Fang- I'm worried that the checkbox could be confusing on login, since we also have additional checkboxes below that. probably needs some sort of visual hierarchy/separator.

shrugs:
we can merge this now and I can make masking part of a separate PR as well. up to you

Fang-:
@jyng do you have thoughts/mockup for us?

This is not fine to be merged, since invite email sending is still broken for me. It starts hanging before it even gets to doing derivation, and it didn't do this before.

Behavior is much more functional if mailer service is online, though it doesn't display the "email has already received" message until you un-focus the field, which you might not make if the "generate invites" button is unclickable. That's a bit confusing.

jyng:
@shrugs @Fang- What are the original fixes that this PR is addressing? I'm a sort of losing the trail of conversation here.

jyng:
Screen Shot 2019-08-14 at 1 40 20 PM
Nice

shrugs:
> This is not fine to be merged, since invite email sending is still broken for me. It starts hanging before it even gets to doing derivation, and it didn't do this before.

oops, skipped my mind. do you have those details about what browser you're using & how many emails you're generating etc? fwiw I'm able to go through the happy path for invite sending with one email, but i realize 'works on my device' is unhelpful. will test this with a bunch of invites and try to debug.

@jyng 

1) replaces our home-grown form logic with a more popular/standard lib, though not without its own quirks. this results in a lot cleaner and less confusing codebase, this PR excluded
2) lets us do async validation/submission, the lack of which was causing a bunch of inconsistency errors in the mino version (i.e. generating keys, checking whether or not an email has already received an invite, seeing if a star is able to be spawned, etc)
3) the ticket masking problem (partial input masking is a no-go, so next best is definitely the checkmark, see Mark's http://passwordmasking.com/ link for deets
  - I'm waiting for your opinion on what that should look like, since, imo, chucking another checkbox into the list below the input on login could be confusing due to the lack of hierarchy, so it needs some distinction that it's attached to the input and is not a separate option on its own.

jyng:
@shrugs Ok, I think 3 should be in a different PR, as its not blocking us from launching. Ping me and we'll continue the discussion for that once that new PR has been made. Let's try focus on fixing 2 and get it merge ready asap.

Fang-:
> do you have those details about what browser you're using & how many emails you're generating etc?

Firefox, and just one email. As I said, it doesn't get into doing generation, so the amount of emails shouldn't matter here. (And it didn't take >10 seconds before, and it just hangs forever (>minute) rn.)

> fwiw I'm able to go through the happy path for invite sending with one email

Is it connecting to an email service successfully? If you stub that out to localhost (I think it already is) and don't actually have that running, what happens? In my case, it hangs.

Fang-:
fwiw I can reproduce in chrome

shrugs:
I've never had the email service running, unfortunately. I remember receiving the code for the gas tank, which I got running, but I'm not sure where the email code lives.

Regardless, the invite email page hadn't been updated to the new format of buildValidator, which is why things were being wonky. The stubbing definitely caused me to not run into that problem when manually testing.

I've coded up a fix but it's late so I'm going to review it tomorrow morning and then submit it here.

shrugs:
ty for patience on that one

shrugs:
here's a dupe of that last commit message to describe why I made some of the changes. give it a run through and lmk.

I wasn't able to reproduce the 'this page is taking a while' error, but I expect it's solved due to correctly handling the validation promise now

---

feat: tidy up InviteEmails.js and cover a few error edge cases

NOTE: the render function diff will be large because of the indentation
change within the `FieldArray` component.

+ update the buildValidator usage to allow additional (optionally async)
  validators per-field and in buildArrayValidator
+ fix: FormError was listening to a wrong error value in a subscription
+ commented more about FormError logic for clarity
+ pull Input.js error logic forward
+ fix FF-specific box-shadow css issue on required & invalid inputs
+ for plurality, we want to use the plural form for 0 counts as well
  (i.e. '0 errors' vs '0 error')
+ simulate a delay when stubbing useMailer
+ switch back to using `name` as the key for invites, receipts,
  and errors in InviteEmails.js, which fixes the accessory bug
  (accessories were not updating when they should have)
+ use the new WARNING pattern for messaging about errors during
  submission while still allowing submission anyway
+ add copy clarifying that the form can still be submitted
+ we now only send the valid invites, avoiding a `throw` case
  + we only tell the user about valid invites that were sent
+ if there was a warning generating invites, we allow the user to send
  the invites anyway

known issues that could/should be handled in a future PR:

- InviteEmails error handling still needs some 👀
- InviteEmails Inputs don't respond to the `error` state tracked within
  InviteEmails, only the error state within final-form's data-model
- you can "Send Invites" with 0 valid invites. Nothing bad happens
  (nothing happens at all, really — you get a success state with
  '0 invites were successfully sent')

Fang-:
This still gives me the hanging behavior for clean, _not stubbed out_ mailer logic pointing to a mailer URL that doesn't actually run. (ie whatever on localhost, but you're not running it) Did you try that case out?

Fang-:
I'll go ahead and merge this regardless. Mailer being unreachable is a bit of an edge case, anyway. But I do think we should track down and fix whatever's causing that hang.

shrugs:
> This still gives me the hanging behavior for clean, _not stubbed out_ mailer logic pointing to a mailer URL that doesn't actually run. (ie whatever on localhost, but you're not running it) Did you try that case out?

I didn't try that case, will take a peek on monday. perhaps an issue with the fetch code never timing out (maybe we need a `Promise.race`?)

PR #223, opened 2019-08-02
Better point "deduction" in Points view

Fang-:
By moving the deduction logic into the Points view, we simplify the Login view and avoid looking up owned points twice.

The Points view can then decide what to do.
- If there's more than one point we can interact with, display the points list
- If we can only interact with one point, and it isn't a pending transfer, forward to the Point page, and remove the Points page from history

Remaining work:
- [x] Account for outgoing transfers in the above logic. Need to think a little bit more about how to do that elegantly. Very tempted to just slap `useMemo` on top of everything in Points.js and hook it up that way.

### DIFF
--- a/src/views/Points.js
+++ b/src/views/Points.js
@@ -130,14 +130,14 @@ export default function Points() {
       Just: r => {
         r.value.matchWith({
           Error: () => null,
-          Ok: p => {
+          Ok: c => {
             let all = [
-              ...p.value.ownedPoints,
-              ...p.value.votingPoints,
-              ...p.value.managingPoints,
-              ...p.value.spawningPoints,
+              ...c.value.ownedPoints,
+              ...c.value.votingPoints,
+              ...c.value.managingPoints,
+              ...c.value.spawningPoints,
             ];
-            const incoming = p.value.incomingPoints.filter(
+            const incoming = c.value.incomingPoints.filter(
               p => !rejectedPoints.includes(p)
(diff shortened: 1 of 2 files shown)


Fang-:
This feels kinda bad, but maybe its ok... Happy to take refactoring suggestions.

Fang-:
@alexmatz please confirm the above is alright.

src/views/Points.js
```diff
@@ -88,6 +88,79 @@ export default function Points() {
     addRejectedPoint,
   ] = useRejectedIncomingPointTransfers();
 
+  const maybeOutgoingPoints = useMemo(() => {
+    return controlledPoints.matchWith({
```
shrugs: nit: just use an inline function `() => controlledPoints.matchWith(...)`

another nit: since we're returning a `Maybe` from a `Maybe` we can use `.chain` to only operate on the `Just` value like `controlledPoints.chain(value => values.matchWith(...))`

shrugs: we're actually doing that twice, also on like 98, so `c.value.ownedPoints.map(point => getDetails(point).map(details => ({point, hasTransferProxy(details)}))` would be cleaner

Fang-:
> only weird thing is that if you're in bridge and get a transfer now you have to re-login to see them

Confirmed with @alexmatz out of band that this is fine. Will do the nit changes in a bit!

PR #225, opened 2019-08-06
Mobile Related Fixes

shrugs:
- https://app.asana.com/0/1130701145004055/705147831874420/f
- https://app.asana.com/0/1130701145004055/1133939626401950/f
- https://app.asana.com/0/1130701145004055/705147831874424/f


package.json
```diff
@@ -72,7 +72,7 @@
     "checksum": "cd build && find . -exec shasum -a256 {} ';' > ../checksums.txt && cd ..",
     "zip": "zip -ur bridge-$npm_package_version.zip README.md bridge-https.py checksums.txt && cd build && zip -ur ../bridge-$npm_package_version.zip ./* && cd ..",
     "release": "npm install && npm run build && npm run checksum && npm run zip",
-    "pilot:ganache": "ganache-cli --blockTime 1 --networkId 1 -m 'benefit crew supreme gesture quantum web media hazard theory mercy wing kitten' > /dev/null &",
+    "pilot:ganache": "ganache-cli --blockTime 1 --networkId 1 --host '0.0.0.0' -m 'benefit crew supreme gesture quantum web media hazard theory mercy wing kitten' > /dev/null &",
```
Fang-: What's this about?

src/store/network.js
```diff
@@ -41,7 +41,7 @@ function _useNetwork(initialNetworkType = null) {
     switch (networkType) {
       case NETWORK_TYPES.LOCAL: {
         const protocol = isDevelopment ? 'ws' : 'wss';
-        const endpoint = `${protocol}://localhost:8545`;
+        const endpoint = `${protocol}://${document.location.hostname}:8545`;
```
Fang-: What prompted this change?

src/views/Login.js
```diff
@@ -70,6 +72,20 @@ export default function Login() {
 
   const [deducing, setDeducing] = useState(false);
 
+  // again, using breakpoint to decide if we're on mobile or not
+  // see ActivateCode for more details
+  const hardwareAllowed = useBreakpoints([false, true, true]);
```
Fang-: As per my comment in that asana task, unsure if disallowing hardware wallets on mobile is the right thing to do. Not sure what we gain from disallowing it. Hardware wallets might be the only sane/safe way of doing crypto-things on mobile! cc @jyng

package.json
shrugs: ah both this and the other change you pointed out are about allowing the eth nod eto be accessible from a mobile device on your wifi network. So you can go to `192.168.0.114:3000` or whatever and pull it up on your iphone. the eth node thing has been iffy, but works occasionally (?)

src/views/Login.js
shrugs: trezor at least definitely doesn't work on mobile, but ledger might?

package.json
Fang-: Ah cool, that seems good! (:

src/views/Login.js
Fang-: I'd argue that regardless of whether they are actually supported or not, we should still display them on mobile? Cultivate user awareness, and also future-proofs. But maybe Design feels strongly about likely-unusable things being too ugly in that context?

Fang-:
This should merge in recent changes from final-form, looks like.

shrugs:
I've made an executive decision on this cause afaik we haven't resolved the 'what do about mobile & hardware wallets' problem. So I've left the tab in place but instead disabled the individual accordion options. This gives us a balance between good UX (nobody clicks ledger or trezor expecting them to work) and visibility (they know that they can use their hardware wallets on a desktop device via the copy nudge).

image

Hopefully everyone's happy enough to let this ride for now, and we can revisit when we definitely know trezor or ledger works on mobile, etc.

PR #231, opened 2019-08-15
Make copy for Arvo keyfile download consistent

Fang-:
Also took the liberty of lightly modifying the detail copy so it's less repetitive with the primary text here.


shrugs:
LGTM, merge it 👍

PR #234, opened 2019-08-19
Ethereum design improvements

shrugs:
- [x] loading indicator


shrugs:
agreed that it's hacky—couldn't figure out a better solution in a reasonable amount of time, though

PR #235, opened 2019-08-19
feat: implement conditional ticket masking following passwordmasking.com's recs

shrugs:
well that was straightforward. love final-form now that it works (lol)

### DIFF
--- a/src/views/Login/Ticket.js
+++ b/src/views/Login/Ticket.js
@@ -3,7 +3,7 @@ import { Just } from 'folktale/maybe';
 import cn from 'classnames';
 import * as azimuth from 'azimuth-js';
 import * as kg from 'urbit-key-generation/dist/index';
-import { Grid, CheckboxInput } from 'indigo-react';
+import { Grid, CheckboxInput, Flex, ToggleInput } from 'indigo-react';
 import { FORM_ERROR } from 'final-form';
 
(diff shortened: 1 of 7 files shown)


shrugs:
wait I forgot to bring this behavior to the other palces we use ticket input like actiavtecode and ticket validation. 1 moment

shrugs:
@Fang- ok fixed

src/views/Login/Ticket.js
```diff
@@ -135,6 +137,8 @@ export default function Ticket({ className, goHome }) {
       point: impliedPoint || '',
       usePasshrase: false,
       useShards: false,
+      showTicket: true,
```
Fang-: wait, I missed this, but shouldn't this be default `false`, as per common password field convention?

shrugs: the website mentions that:

> When offered the choice of masked or unmasked passwords within the interface, the concept was identified as a feature not an error. Participants appreciated the usability benefits of clear text passwords and the security benefits of masked passwords. When participants used this option their trustingness of the website was unaffected. (Note the password was unmasked by default)

Fang-: Oh I hadn't remembered that note at the end. Very interesting. This is probably fine then.

PR #236, opened 2019-08-19
fix: tank messaging

shrugs:


shrugs:
@Fang- probably ready for a review, though I don't think this changes any of the semantics.

right now we seem to be following the pattern of "accept and return strings and only use BN within functions" which seems pretty reasonable. we could also try to store and use BNs entirely across the app, and only convert to strings in order to pass them to web3. my hunch as to why web3 stopped accepting BNs was to reduce the dependency on a specific bignum lib, since there's also BigNumber.js, and it shouldn't matter which library they use internally.

src/lib/tank.js
```diff
@@ -54,67 +54,61 @@ const ensureFundsFor = async (
...
-      } else {
-        console.log('tank: skipping remaining-funds check');
-        //TODO if we can't always (easily) provide a point, and the
-        //     fundTransactions call is gonna fail anyway, should we maybe
-        //     just not bother doing this check in the first place?
-      }
+  if (balance.gte(cost)) {
+    console.log(
+      `tank: already have sufficient funds: ${address} proposed ` +
+        `a transacton(s) that will cost ${cost.toString()}wei but it ` +
```
Fang-:
```suggestion
        `transaction(s) that will cost ${cost.toString()}wei but it ` +
```

src/lib/tank.js
```diff
@@ -54,67 +54,61 @@ const ensureFundsFor = async (
...
+
+    await waitForTransactionConfirm(web3, res.txHash);
+
+    const newBalance = await web3.eth.getBalance(address);
     console.log(
-      'tank: already have sufficient funds',
-      cost.toString(),
-      address,
-      balance.toString()
+      `tank: funds have confirmed: ${address} now has ` +
```
Fang-: Kinda tempted to suggest we should do a `newBalance > cost` check here, just in case of weirdness with the funding tx... Sane y/n?

src/lib/tank.js
```diff
@@ -54,67 +54,61 @@ const ensureFundsFor = async (
...
+    console.log('tank: funding failed, waiting on user funds...', e);
+    await waitForBalance(web3, address, cost, askForFunding, gotFunding);
   }
+
   return false;
 };
 
 // returns a promise that resolves when address has at least minBalance
 function waitForBalance(web3, address, minBalance, askForFunding, gotFunding) {
-  minBalance = toBN(minBalance);
```
Fang-: I put this here to ensure that `minBalace` is always a BN, so we don't run into issues with the `.gte()`. Pretty sure `toBN` on a BN is a cheap no-op. Is there a specific reason you want to take this out?

src/lib/tank.js
shrugs: seems like we ran into that in bug reports, so probably worth it, yeah

src/lib/tank.js
shrugs: I actually pass minBalance as a BN here, since this function is only used internally, but I realize that's not conforming to the contract where we input and output strings. will change it back

src/lib/tank.js
Fang-: I think just fall back to asking for funds in that case? Retrying the gas tank is another option, but not sure if that's objectively better.

shrugs:
@Fang- do you think this PR fixes https://app.asana.com/0/1130701145004055/1136984452086041/f ?

Fang-:
I expect that to have gotten solved by #239 already?

shrugs:
> I expect that to have gotten solved by #239 already?

I also thought so! but I thought that the BN fix was manually deployed while that error was generated. if not then yeah we can totally complete that card.

PR #238, opened 2019-08-20
Feat/copy things

shrugs:


shrugs:
the 'correct' way to do this, imo, is to throw a little 'copiable' icon at the end and a tooltip for feedback (like how AWS does it)

image

image


(note that we want to avoid reflowing the page at all costs)

PR #239, opened 2019-08-20
Use BNs for comparisons in tank lib

Fang-:
Minimum viable fix as currently deployed.

Remaining:
- [x] Find other potentially problematic comparisons
- [x] Make code cleaner


shrugs:
we might want to use `{ toBN } from 'web3-utils'` instead of `web3.utils` since we use that import elsewhere

pretty much every number we ever create that ends up in ethereum should be a BN. so the `cost`, `gasPrice`, etc variables should all use BN math.

Fang-:
Hmm `toWei` apparently doesn't like BNs as inputs. Not sure why, or how that didn't get caught during local testing. Should be all good new though!

PR #244, opened 2019-08-22
Actually set Trezor HD path on login

### LINKED ISSUE #240: "Cannot read property 'slice' of undefined"
Went to `littel-ponnys.azimuth.network` to try using bridge to do a personal breach so I could get ponnys back on the network. Went to the Networking tab. Clicked the "Trigger New Continuity Era" check box. Clicked the Reset Transaction button. Got "Cannot read property 'slice' of undefined" underneath.

Here's all the console output:
```
2useKeyfileGenerator.js:65 seed is nondeterminable for revision 1
AdminNetworkingKeys.js:85 deriving seed with revision 2
useKeyfileGenerator.js:76 keys do not match details for revision 1
useEthereumTransaction.js:59 TypeError: Cannot read property 'slice' of undefined
    at trezor.js:26
    at u (runtime.js:45)
    at Generator._invoke (runtime.js:271)
    at Generator.e.<computed> [as next] (runtime.js:97)
    at r (asyncToGenerator.js:3)
    at o (asyncToGenerator.js:25)
```

Fang-:
Fixes #240.

See commits for PR details.

### DIFF
--- a/src/lib/trezor.js
+++ b/src/lib/trezor.js
@@ -21,6 +21,10 @@ const trezorSignTransaction = async (txn, hdpath) => {
     transaction: trezorFormattedTxn,
   });
 
+  if (!sig.success) {
+    throw new Error(sig.payload.error);
+  }
+
   const payload = sig.payload;
 
   txn.v = Buffer.from(payload.v.slice(2), 'hex');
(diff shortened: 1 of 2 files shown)


Fang-:
That was the bugfix, yes. (:

(Specifically, the bug was using `hdPath` in one place, even though we had named the form (and used it elsewhere) as `hdpath`. The former spelling was more consistent with all other variables though, so I chose to rename the conforming cases instead of the non-conforming ones.)

shrugs:
ahhh yeah totally 👍

Issue #243, opened 2019-08-22
Networking key configuration not enabled right away


Fang-:
When you visit `AdminNetworkKeys`, you see this:

<img width="542" alt="image" src="https://user-images.githubusercontent.com/3829764/63546541-58c34000-c52a-11e9-83b2-95406a52abb3.png">

The button becomes clickable only after you toggle the checkbox, or open and close the "custom" dialog. It should just be clickable from the start.

(This is when logged in with a mnemonic. Not sure if it repros for ticket login, where key derivation is deterministic.)

Fang-:
The keyfile downloaded after doing this is always the same, too.

shrugs:
hmm, I'm getting odd behavior around the final-form render props: https://github.com/final-form/react-final-form/issues/605

but basically the issue is that

1) we're seeing this flickering
2) react seems to debounce effect running
3) the effect is being run on the invalid state
...
4) final-form stops 'flickering' once an input is made, for some reason, which is why things work after that.

shrugs:
still looking into it

PR #248, opened 2019-08-26
fix: correctly re-execute onValues after validation resumes

shrugs:
problem was that `form` is constant so we weren't correct reacting to the validation paused state, so `onValues` was never called once validation was unpaused (i.e. right after mount)

solves #243


shrugs:
shouldn't have committed that here, you're right, yeah. just removed them for clarity

PR #251, opened 2019-08-27
fix: parseInt returns NaN, but we expected it to throw

shrugs:
I was getting a weird behavior around forms when a number input was totally cleared and realized that parseInt doesn't throw, it returns `NaN` when it fails. Which means all of our error handling assumptions about parseInt throughout the app were wrong (i.e., my assumption). 

So this creates a throwing version of parseInt and then uses it everywhere that seems reasonable.


shrugs:
there's one occurrence in wallet.js that I didn't want to mess with cause it seemed complex

PR #255, opened 2019-08-27
fix: unify needs funds component and give pending affordance

shrugs:
finally feels good to DRY up this codepath

### DIFF
--- a/src/components/NeedFundsNotice.js
+++ b/src/components/NeedFundsNotice.js
@@ -1,0 +1,25 @@
+import React from 'react';
+import { fromWei } from 'web3-utils';
+
+import Highlighted from './Highlighted';
+import CopiableAddress from './CopiableAddress';
+import WarningBox from './WarningBox';
+import Blinky from './Blinky';
+
+export default function NeedFundsNotice({
+  address,
+  minBalance,
+  balance,
+  ...rest
+}) {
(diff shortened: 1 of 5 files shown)


src/components/NeedFundsNotice.js
```diff
@@ -0,0 +1,25 @@
...
+  balance,
+  ...rest
+}) {
+  return (
+    <WarningBox {...rest}>
+      <Highlighted warning>
+        Your ownership address <CopiableAddress>{address}</CopiableAddress>{' '}
+        needs at least {fromWei(minBalance)} ETH and currently has{' '}
+        {fromWei(balance)} ETH. The transaction will automatically resume once
+        ETH is available. Waiting... <Blinky />
```
Fang-: Is this maybe missing the trailing `{' '}`, or is that only in/around interpolation of any kind? The sentence reads a bit poorly anyway, should probably say "once _enough_ ETH is available".

shrugs: the trailing `{' '}` is only necessary when, by default, react would collapse the newline into no space at all. by default a newline in jsx between non-interpolated values will be treated as a normal space. so yeah, pretty much. my editor automatically adds these where necessary, though, so I'm trusting it a little bit.

PR #257, opened 2019-08-27
feat: error boundary mvp

shrugs:
in dev the create-react-app error boundary will overlay ours, but just dismiss it to see what a production user would see

also note: error boundaries only work with class components, hence the wrapper component.


src/components/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,72 @@
...
+      }),
+    [error]
+  );
+
+  return (
+    <div className="mw1 ph4">
+      <h2 className="mt8">
+        Bridge Error!{' '}
+        <span role="img" aria-label="a bridge on fire">
+          🌉🔥
```
Fang-: lol

src/components/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,72 @@
...
+        </span>
+      </h2>
+      <p className="mt4 f5">
+        We caught an error thrown by the Bridge application client-side code.
+        <br />
+        <b>Your points and assets are most likely safe.</b>
```
Fang-: >most likely

I'd either display this phrase with absolute confidence, or not at all.

Proposed alternative:  
"This does not affect your points and assets."

shrugs: sgtm. will wait on double-confirm from @alexmatz

Fang-: @alexmatz @vvisigoth I think this is good to go in after we get copy comments here.

Fang-:
This was previously reviewed and still looks sane, so gonna go ahead and merge.

PR #260, opened 2019-08-27
Always suggest whole gwei amounts

### LINKED ISSUE #259: Failing to convert decimals to BN
Interface won't accept email address.  Note:  Email address used in this case had a dot before the @ symbol. 

> Error: [number-to-bn] while converting number 3.3 to BN.js instance, error: invalid number value. Value must be an integer, hex string, BN or BigNumber instance. Note, decimals are not supported. Given value: "3.3"

Screen Shot 2019-08-27 at 9 24 04 AM

Fang-:
Should fix #259.


Fang-:
(Maybe we should actually support decimal gwei values, but let's at least get a quick fix in.)

shrugs:
was just about to fix this, nice

shrugs:
LGTM

PR #261, opened 2019-08-27
a11y pass

shrugs:
I'm doing an accessibility pass, mostly around forms and navigation. learning a11y as I go, so please double check my work.

notes:
- `aria-required` is tough to add since our 'required' validation is in the client-side validators, not within our view layer at all. plus some fields are conditionally required, which makes things a bit harder.
- because of our design token approach, it's impossible to implement generalized selectors for resetting `button` styles, since the `button { background-color: inherit; }` selector is more specific than `bg-red`. So 

references:
- https://bitsofco.de/using-aria-live/
  - using this site as a reference, I chose not to add the aria-live attribute to our warning messages, since they only recommend adding the attribute to static elements, and our warnings are not static at all. my hunch is that the new `role="alert"` on `ErrorText` will help a lot for screenreaders and handling errors.
- https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML
- https://www.w3.org/WAI/tutorials/forms/notifications/
- https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA21.html
- https://webaim.org/techniques/formvalidation/
- https://webaim.org/techniques/keyboard/

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,11 +2,13 @@
   "extends": [
     "react-app",
     "prettier",
-    "prettier/react"
+    "prettier/react",
+    "plugin:jsx-a11y/recommended"
   ],
   "plugins": [
     "prettier",
-    "react-hooks"
+    "react-hooks",
+    "jsx-a11y"
   ],
   "rules": {
     "prettier/prettier": "error"
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,7 @@
     "react-app",
     "prettier",
     "prettier/react",
-    "plugin:jsx-a11y/recommended"
+    "plugin:jsx-a11y/strict"
   ],
   "plugins": [
     "prettier",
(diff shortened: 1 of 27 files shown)


shrugs:
I apologize in advance for this PR, but since I'm learning as I go along it's really difficult to portion out changes and correctly structure my work.

jyng:
Have you taken a look at https://ui.reach.tech/? We're using that for indigo-react and has built-in accessibility. Could be some overhead on using a new framework.

shrugs:
if new indigo is accessible then that's good. will try and get current version to bare minimum (keyboard control) and then hope the rewrite to new indigo will wrap things up

Fang-:
> Have you taken a look at https://ui.reach.tech/?

image

Nice, very accessible. (^:

Good to know this is being tackled in indigo!

shrugs:
stale, closing

PR #266, opened 2019-08-30
Back button for logout

Fang-:
If the back button would log you out, it becomes the eject symbol instead of the back-arrow.

image

### DIFF
--- a/src/components/View.js
+++ b/src/components/View.js
@@ -7,6 +7,8 @@ import useBreakpoints from 'lib/useBreakpoints';
 import MiniBackButton from './MiniBackButton';
 import { useHistory } from 'store/history';
 
+const EXPECT_LOGOUT_WHEN_POPPING_AT_DEPTH = 2;
+
 // View is a top-level component that all Views must render to inherit styling
 function View({
   className,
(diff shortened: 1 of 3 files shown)


vvisigoth:
I like this and @galenwp was just noting this was a problem. @galenwp any objections to this icon? I personally am neutral to positive.

galenwp:
The icon is fine, but I think it should still say 'Log out' in text. It's
really annoying to log out by accident because you have to enter your
ticket again.

Fang-:
>text _and_ icon

Hmm not sure if I'm a fan...

image

src/components/View.js
```diff
@@ -28,6 +28,7 @@ function View({
   const goBack = useCallback(() => pop(), [pop]);
 
   const showBackButton = size > 1 && !!pop;
+  const backIsLogout = size === 2;
```
shrugs: should probably be a constant instead of magic variable since it relies on assumptions about our router depth

vvisigoth:
@galenwp @jyng should we decrease the weight of that "Log out"? Speak soon or this is going in, because either way is better than accidentally logging youself out...

Fang-:
image

Note that my changes make the "sidebar" in which the back button lives no longer fixed-size. This means that fluctuations in its size cause content to be pushed off-center for the "log out" case. This isn't ideal.

Fang-:
As discussed with @vvisigoth: last commit needs to be reverted.

PR #273, opened 2019-09-05
Metamask authentication

liam-fitzgerald:
Adds Metamask authentication to log into bridge.

Metamask doesn't support signing a tx before broadcast, so we "sign" a fake tx
and have metamask sign it upon broadcasting.

### DIFF
--- a/src/components/Buttons.js
+++ b/src/components/Buttons.js
@@ -25,7 +25,10 @@ export const OfflineButton = props => (
 );
 
 export const BootArvoButton = props => (
-  <OutButton href="https://urbit.org/docs/getting-started" detail="Boot your computer" {...props}>
+  <OutButton
+    href="https://urbit.org/docs/getting-started"
+    detail="Boot your computer"
+    {...props}>
     Boot Arvo
   </OutButton>
 );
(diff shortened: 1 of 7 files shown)


vvisigoth:
I'll confirm functionality, @Fang- @shrugs would you mind looking at style when you get a chance?

jyng:
I think we want to move towards a master-ticket focussed login and deemphasize other login methods. It gives more focus to a suggested path and makes the login page feel less crowded. Let's move forward with this.
login other
login other expanded

jyng:
@liam-fitzgerald Do you think this is doable?

liam-fitzgerald:
Absolutely will update

galenwp:
@jyng I could see us also putting master ticket shards under 'other'

vvisigoth:
Looks good, @liam-fitzgerald just pulled and ran with @jyng

Fang-:
Better late than never: this PR seems fine code-wise! Feels a bit weird to be doing the fake transaction thing, trying to think of cases where it could bite us but coming up blank. Probably Fine™.

> It gives more focus to a suggested path and makes the login page feel less crowded. Let's move forward with this.
> login other

That looks kinda Bad though. I think we passed the point where we need to go back to a dropdown or similar for this, @jyng?

PR #278, opened 2019-09-09
Bump urbit-ob and urbit-key-generation dependencies

jtobin:
Updates to recent versions of urbit-ob and urbit-key-generation.  Of note, they should now be imported via e.g.

    import ob from 'urbit-ob';
    import kg from 'urbit-key-generation';

so this also changes their imports accordingly.


g-a-v-i-n:
#277 resolves this

PR #286, opened 2019-09-19
Fix submit on enter

liam-fitzgerald:
Form submission on enter


liam-fitzgerald:
Maybe we should look at enforcing lint with CI? Would dovetail nicely with when i get around to tests

Fang-:
Agreed, having CI here double-check the style for you would be ideal.

PR #295, opened 2019-10-01
View +code inside bridge

liam-fitzgerald:
Allows for the viewing of the login code iff deterministic network keys are used.

Also fixes sigil generation issues.

### DIFF
--- a/src/lib/networkCode.js
+++ b/src/lib/networkCode.js
@@ -1,4 +1,5 @@
 import { hex2patp } from 'urbit-ob';
+import { createRing } from 'lib/keys';
 
 import jssha256 from 'js-sha256';
 
@@ -7,7 +8,7 @@ function shax(buf) {
   return Buffer.from(hashed);
 }
 
-function hexToArrayBuffer(hex) {
+function hex2buf(hex) {
   return Buffer.from(hex, 'hex').reverse();
 }
 
@@ -38,10 +39,10 @@ function shaf(buf, salt) {
   return xor(front, back);
 }
 
-export const generateCode = async privKeyHex => {
-  const privKey = hexToArrayBuffer(privKeyHex);
-  const salt = hexToArrayBuffer('73736170');
-  const hash = shax(privKey);
+export const generateCode = pair => {
+  const ring = hex2buf(createRing(pair));
+  const salt = hex2buf('73736170'); // salt is the noun %pass
+  const hash = shax(ring);
   const result = shaf(hash, salt);
   const half = result.slice(0, result.length / 2);
 
--- a/src/lib/networkCode.js
+++ b/src/lib/networkCode.js
@@ -1,0 +1,49 @@
+import { hex2patp } from 'urbit-ob';
+
+import jssha256 from 'js-sha256';
+
+function shax(buf) {
+  const hashed = jssha256.sha256.array(buf);
+  return Buffer.from(hashed);
+}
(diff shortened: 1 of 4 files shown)


vvisigoth:
Nice! Busy day today @Fang-  can you handle review and merge?

src/lib/networkCode.js
```diff
@@ -0,0 +1,49 @@
...
+  return Buffer.from(buf)
+    .reverse()
+    .toString('hex');
+}
+
+function shas(buf, salt) {
+  return shax(xor(salt, shax(buf)));
+}
+
+function xor(a, b) {
```
Fang-: Do we not have a xor implementation in one of the crypto things we import?

src/views/Point.js
```diff
@@ -103,6 +107,15 @@ export default function Point() {
           Admin
         </Grid.Item>
         <Grid.Divider />
+        <Grid.Item
+          full
+          as={ForwardButton}
+          accessory={code && <CopyButton text={code} />}
+          detail={code || notice}
```
Fang-: Shouldn't this also be `disabled` if `code` is falsy?

src/lib/useKeyfileGenerator.js
```diff
@@ -71,6 +74,8 @@ export default function useKeyfileGenerator(manualNetworkSeed) {
 
     const pair = deriveNetworkKeys(_networkSeed);
 
+    generateCode(pair.crypt.private + pair.auth.private + '42').then(setCode);
```
Fang-: The `'42'` here is probably some Urbit arcana? At the very least deserves a comment, if not a named `CONSTANT`.  
Might be better to move the entire concatenation here into `lib/networkCode` though, so that callers don't need to know anything about non-Bridge crypto.

joemfb: `42` here is actually `0x42`, ie `'B'`. This line is constructing a `+ring` for cryptosuite B. Do we know already have code to do this?

src/lib/networkCode.js
```diff
@@ -0,0 +1,49 @@
...
+  const result = shas(buf, salt);
+  const halfway = result.length / 2;
+  const front = result.slice(0, halfway);
+  const back = result.slice(halfway, result.length);
+  return xor(front, back);
+}
+
+export const generateCode = async privKeyHex => {
+  const privKey = hexToArrayBuffer(privKeyHex);
+  const salt = hexToArrayBuffer('73736170');
```
joemfb: Worth noting that this hex string is just `%pass`.

src/lib/networkCode.js
liam-fitzgerald: There's one in a transitive dependency, should I make it a direct dep and use that?

Fang-: Hmm, I think that's still best? Re-implementing primitives like this feels like trouble. Let's do it.

liam-fitzgerald: In a sick turn of events the dependency is buggier than the handrolled version. I could bring it in and hack around it, but it might just be better to leave it

Fang-: Agreed. At the first sign of "even more trouble", run!

liam-fitzgerald: Ready to go then!

vvisigoth:
@Fang- what say you?

PR #296, opened 2019-10-07
incorporate new card designs

liam-fitzgerald:
Switched to new card designs for viewing points.


src/lib/card.js
```diff
@@ -0,0 +1,191 @@
...
+  canvas.height = y * dpi200;
+  canvas.style.width = x + 'px';
+  canvas.style.height = y + 'px';
+
+  canvas.getContext('2d').scale(dpi200, dpi200);
+
+  return canvas;
+};
+
+// takes a
```
Fang-: This doesn't seem finished. Maybe we want comments for most of these functions/data structures as well. Comments in walk and rand are good!

src/components/Passport.js
```diff
@@ -46,139 +61,256 @@ const buildKeyType = permissions => {
...
+          display: 'flex',
+        }}>
+        <div
+          style={{
+            width: '100%',
+            display: 'flex',
+            alignItems: 'center',
+          }}>
+          {Just.hasInstance(point) && (
+            <MaybeSigil patp={patp.map(ob.patp).value} />
```
vvisigoth: I'm getting "patp.map" is not a function

vvisigoth:
Clicking into Point view shows a mangled sigil
Screenshot 2019-10-08 12 12 46

vvisigoth:
Also, no animation modes (besides "none") seem to work?

vvisigoth:
I have no proof of this, but it seems like the "size" of the sigil is being set on mini-sigil and not reset on the page transition?

vvisigoth:
tagging @g-a-v-i-n  to take a look, too.

src/lib/card.js
g-a-v-i-n: Resolved in 4543574

src/components/Passport.js
g-a-v-i-n: Resolved in 4543574

vvisigoth:
@liam-fitzgerald It looks like this is also still broken on the "Accept Invite" flow. I think that this might have something to do with the fact that the Passport is being passed a hook called "derivedPoint" in lieu of a prop. This crashes the Passport, since (I believe) this is expecting a point and not a hook. Can you take a look?

vvisigoth:
@liam-fitzgerald @g-a-v-i-n Check out this branch, https://github.com/urbit/bridge/tree/invite-dev

The last commit there comments out calls to gas tank/email server so that, if you create invites (to a dummy email) it will fail and print the ticket (like `~master-morzod-poldec-ropdyl`) to the screen. To redeem this, you can go to `http://localhost:3000/#master-morzod-poldec-ropdyl` or go to the activate screen and enter that ticket.

vvisigoth:
The above will make it easier to test the invite redemption flow locally.

vvisigoth:
Also, see @Fang- new commit that adds a new migration.js that starts your ganache state out with migrations, etc. https://github.com/urbit/azimuth/commit/a3539dfd626bbad897ac1ed12287df6be7427ce8

liam-fitzgerald:
Fixed up animations and invite flow.
There's CSS issue with the card though. It overflows its container on some screen sizes . Possible fixes are making the layout flow down, like it does for mobile, or adjusting the card component to resize nicely.

Safe to rebase @g-a-v-i-n ?

[
Screen Shot 2019-10-10 at 3 36 14 pm
](example)
Screen Shot 2019-10-10 at 3 39 14 pm

g-a-v-i-n:
@liam-fitzgerald on which breakpoint is this an issue?

g-a-v-i-n:
The second vertical image looks great, but make sure that the card + text are horizontally centered like other menus in Bridge.

vvisigoth:
@liam-fitzgerald can you handle the layout stuff ASAP? I'd like to get this built and deployed.

vvisigoth:
@liam-fitzgerald @g-a-v-i-n was able to fix the layout stuff, can you fix up these merge conflicts and merge mino back in? Doing so caused me to get errors right at the end of the transaction flow. It looks like there's a conflict between the +code PR and this work here and it manifests itself once the reticket flow is finished and you are redirected back to the logged-in page.

I'm not super familiar with this work, so I'd appreciate you taking a look and fixing this up.

Screenshot 2019-10-16 19 59 24

vvisigoth:
@liam-fitzgerald scratch the above, I realized that it was the +code PR so I reverted that and now everything is gravy. Please resubmit that PR.

g-a-v-i-n:
yehaw

PR #299, opened 2019-10-08
Nemesis copy

vvisigoth:
Updates progress notifications to make clear we're not stealing their money.


vvisigoth:
Yeah, if the "transaction" terminology, though "correct" is misleading if
you don't understand Ethereum.
—
~poldec-tonteg
http://urbit.org

PR #303, opened 2019-10-17
Revert "View +code inside bridge"

vvisigoth:
Reverts urbit/bridge#295

These changes introduced a crash on point acceptance.

Screenshot 2019-10-16 19 59 24


vvisigoth:
@liam-fitzgerald these appear to be the cause of the crashes, have confirmed that this fixes the crash taht I identified in the card PR.

PR #305, opened 2019-10-17
Fix/view code

liam-fitzgerald:
Fix crash on viewing of point that is not yet loaded/controlled

### DIFF
--- a/src/lib/useKeyfileGenerator.js
+++ b/src/lib/useKeyfileGenerator.js
@@ -32,9 +32,12 @@ export default function useKeyfileGenerator(manualNetworkSeed) {
   const [code, setCode] = useState(false);
 
   const _point = need.point(pointCursor);
-  const _details = need.details(getDetails(_point));
+  const details = getDetails(_point);
 
-  const networkRevision = convertToInt(_details.keyRevisionNumber, 10);
+  const networkRevision = details.matchWith({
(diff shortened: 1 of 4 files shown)


src/lib/useKeyfileGenerator.js
```diff
@@ -32,9 +32,12 @@ export default function useKeyfileGenerator(manualNetworkSeed) {
   const [code, setCode] = useState(false);
 
   const _point = need.point(pointCursor);
-  const _details = need.details(getDetails(_point));
+  const _details = getDetails(_point);
```
Fang-: I think this should be `details` then. The `_` in this context generally indicates "we unwrapped it from `Maybe` already".

src/lib/useKeyfileGenerator.js
```diff
@@ -32,9 +32,12 @@ export default function useKeyfileGenerator(manualNetworkSeed) {
...
 
-  const networkRevision = convertToInt(_details.keyRevisionNumber, 10);
+  const networkRevision = _details.matchWith({
+    Just: ({ value }) => convertToInt(value.keyRevisionNumber, 10),
+    Nothing: () => -1,
```
Fang-: The value `-1` never naturally pops up for these. To avoid stubbing our toes on edge cases, should we maybe use `0` here instead?

Fang-:
Code looks good to me now!

PR #317, opened 2019-10-28
Loading screen on points view

liam-fitzgerald:
Shows a loading indicator when chain state is loading, so that
Urbit HD wallets immediately show the (only) point view instead
of flashing through the point select view.

### DIFF
--- a/src/views/Points.js
+++ b/src/views/Points.js
@@ -206,6 +206,23 @@ export default function Points() {
     push,
   ]);
 
+  if (
+    loading ||
+    (allPoints.length === 1 &&
+      ownedPoints.length === 1 &&
+      outgoingPoints.length === 0)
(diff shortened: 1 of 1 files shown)


vvisigoth:
This still causes a flash of family view on loading an address with one point.

Kapture 2019-10-30 at 14 10 09

liam-fitzgerald:
fixed

vvisigoth:
Now the opposite. Eternal loading on single point ownership.

Kapture 2019-10-31 at 17 31 56

vvisigoth:
Still hanging.

To be clear, I'm testing by booting with clean chain state. Creating a ~zod, logging out and then logging back in.

vvisigoth:
Any progress on this one?

PR #334, opened 2019-11-06
Networking keys config: hex lengths

Fang-:
Web3's randomHex generates bytes, not characters. This makes sure we generate random seeds at a length consistent with the seeds produced by keygen.

Our buildHexValidator takes a length in characters, not bytes. This makes sure we validate accordingly.

(Arguably, hexValidator should take length in bytes also, but I don't want to go up-end that right now.)


liam-fitzgerald:
LGTM!

PR #335, opened 2019-11-07
Linear and conditional star release contract support

liam-fitzgerald:
Can view and withdraw stars from both contracts.

Drafted, waiting on the batching PR approval and also the addresses of the contracts on the Ropsten network.


vvisigoth:
Approved the batching PR and merged.

liam-fitzgerald:
ping @vvisigoth @Fang-

src/lib/starRelease.js
```diff
@@ -0,0 +1,86 @@
+import { linearSR, conditionalSR } from 'azimuth-js';
+import { Just, Nothing } from 'folktale/maybe';
+
+function mapRange(amount, cb) {
+  return [...Array(amount).keys()].map(cb);
+}
+
+// Preferentially withdraw from conditional contracts and use linear
+// to mop up the rest
+export function generateWithdrawTxs(contracts, batchLimits, amount, to) {
```
Fang-: wh-what

I don't understand at all why we're mixing the use of linear and conditional, both here and in `lib/useStarReleaseStore`. If this is somehow useful for creating the desired UI flows, I'd expect that to happen in those UI flows, not in a lib...

src/lib/starRelease.js
```diff
@@ -0,0 +1,86 @@
...
+    contracts,
+    batchLimits,
+    amount,
+    to
+  );
+  const lsrTxs = generateLinearWithdrawTxs(contracts, left, to);
+  return [...csrTxs, ...lsrTxs];
+}
+
+function generateConditionalWithdrawTxs(
```
Fang-: A more explicit loop (instead of recursion) might be a bit easier to follow here. Or at least annotate that this is running itself for every batch/limit.

src/lib/starRelease.js
```diff
@@ -0,0 +1,86 @@
...
+  batchLimits,
+  amount,
+  to,
+  offset = 0
+) {
+  // Successfully withdraw all requesed
+  if (amount <= 0) {
+    return [0];
+  }
+  // Return missing amount in car
```
Fang-: >car

?

src/views/Release/Active.js
```diff
@@ -0,0 +1,94 @@
...
+        ->
+      </Grid.Item>
+      <Grid.Divider />
+    </>
+  );
+}
+
+export default function ActiveView({ className, goPoint }) {
+  // const points = [2560, 2561, 2562];
+  const { controlledPoints } = usePointCache();
```
Fang-: Wait, so this just shows *all* the points the user controls? Does this replace the old point overview page?

src/views/Release/Active.js
```diff
@@ -10,11 +10,13 @@ import { useWallet } from 'store/wallet';
 
 import * as need from 'lib/need';
 import usePermissionsForPoint from 'lib/usePermissionsForPoint';
+import { useSyncOwnedPoints } from 'lib/useSyncPoints';
 import { buildKeyType } from 'lib/point';
 
 function ActiveViewRow({ point, goPoint }) {
   const { wallet } = useWallet();
   const address = need.addressFromWallet(wallet);
+  useSyncOwnedPoints([point]);
```
Fang-: Does the result of this not need to be assigned for something? If we're not using it, shouldn't we remove it?

liam-fitzgerald: It simply ensures that `point` is synced. It doesn't need to be assigned.

src/lib/starRelease.js
liam-fitzgerald: I put in the lib because I didn't want the UI code to become too cluttered. I thought it would be cleaner to hide the difference between the contracts in a lib, because from the bridge perspective, the two contracts don't really need to be exposed differently.

src/lib/starRelease.js
liam-fitzgerald: hangup from my lisping days, will change to head.

src/views/Release/Active.js
liam-fitzgerald: Just the stars.

src/lib/starRelease.js
Fang-: That sounds like fair game. Carry on!

src/views/Release/Active.js
Fang-: Right. Feels weird to have a "the stars you own" overview for this though, instead of just showing what's still in lockup. If the intent is to show "stars you got from this contract", then that's broken in the "transferred a star away" and "got a star from someplace else" cases.

Fang-:
Still confused about exactly what we want the "active stars" section here to do, but otherwise good to go as far as I'm concerned. Ping @vvisigoth for final approval.

vvisigoth:
Good with me.

PR #339, opened 2019-11-07
Individual breach language

vvisigoth:
Trigger New Continuity Era -> Breach Continuity


vvisigoth:
Changing this wording since it should include the word "Breach"

PR #342, opened 2019-11-16
prevent users bailing midway through activate

liam-fitzgerald:
Tries to stop users bailing midway through the activation flow. [Sadly](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload) we can't give any custom warning message.


liam-fitzgerald:
ping @Fang-

PR #350, opened 2019-12-07
Warn about unset networking keys when setting up pools

Fang-:
The Delegated Sending contract requires that networking keys for the star have been configured at least once. We weren't warning the user about that. This PR makes it so that we do. To avoid nasty star-isn't-live situations, we also warn if keys are _currently_ unconfigured.

(Yes, this means two separate warnings can show. Also idk why these are red instead of yellow. They don't actually stop you from doing this flow.)

image

This is the Bridge-side half of urbit/azimuth-js#55. Once that's merged and assimilated  into Bridge, the meaning of "available for use" here will change from "invite tx could crash" to "won't count as usable invite".


Fang-:
@vvisigoth I assume this has your blessing.

Issue #357, opened 2019-12-15
"invalid mnemonic" message could be more helpful


abliss:
If one word is misspelled (i.e. does not appear on the wordlist), the message should tell me which one. If the string doesn't parse because I accidentally put two spaces instead of one somewhere, it should tell me where. (These are perhaps better made as requests against bip39.js?)

Also, since the mnemonicToSeed is to be used instead of mnemonicToEntropy, there should (as recommended by the bip39 readme) be an option to bypass the check and proceed anyway (in case an alternate or obsolete wordlist was used).

PR #360, opened 2019-12-19
Mnemonic login QoL

### LINKED ISSUE #355: changing the default HD path clears valid mnemonic input
reproduce with:

1. paste in a valid mnemonic 
2. toggle passphrase / hdpath checkbox
3. backspace or insert a character in the hdpath field
4. the form will clear the input to the mnemonic and checkbox, but not the hdpath

Untitled

Fang-:
- Actually validate user-specified derivation paths.
- Fix #355.
- Resolve the "want to use any mnemonic" part of #357, by allowing that if you check a checkbox.
- Sneaks in some clean up of the `Senate.js`, which was spewing hella warnings.

### DIFF
--- a/src/views/Senate.js
+++ b/src/views/Senate.js
@@ -1,42 +1,26 @@
 import React, { useCallback, useMemo, useState } from 'react';
-import cn from 'classnames';
-import { Grid, Text, Input, P, LinkButton } from 'indigo-react';
+import { Grid, Text, P, LinkButton } from 'indigo-react';
 import * as azimuth from 'azimuth-js';
-import ob from 'urbit-ob';
 
 import { useNetwork } from 'store/network';
-import { usePointCache } from 'store/pointCache';
 import { usePointCursor } from 'store/pointCursor';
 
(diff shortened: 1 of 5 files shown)


src/views/Senate.js
```diff
@@ -185,15 +163,15 @@ export default function Senate() {
         </>
       );
     });
-  }, [open, polls, votingOn]);
+  }, [open, polls, votingOn, doVote]);
 
   const onDocumentChange = useCallback(({ valid, values }) => {
```
liam-fitzgerald: why comment these instead of removing?

Fang-: I agree it's not very tidy, but I imagine we'll want to get the "verify/create a document hash" functionality back in once we upgrade this to a more fully-featured senate page.

@vvisigoth we _are_ going to upgrade this to a more fully-featured senate page sometime next year, right?

vvisigoth:
yes
—
~poldec-tonteg
http://urbit.org

PR #369, opened 2020-01-08
New Invite Screen

liam-fitzgerald:
Added new invites screen and invite URL support.

cc: @vvisigoth @jyng


src/form/EmailChipInput.js
```diff
@@ -0,0 +1,120 @@
...
+  );
+
+  const addToChips = useCallback(() => {
+    setChips([...chips, value]);
+    setValue('');
+  }, [chips, value, setChips]);
+
+  const handleKeyDown = useCallback(
+    event => {
+      if (['Enter', 'Tab', ','].includes(event.key) && value !== '') {
```
Fang-: Can we make space do this too? ime that's commonly supported, and it doesn't have conflicts with standard/expected behaviors like tab and return do (next field and submit respectively).

src/components/InviteSigilList.js
```diff
@@ -0,0 +1,47 @@
...
+      Math.max(6 - _acceptedPoints.length - _pendingPoints.length, 0)
+    ).keys(),
+  ].map(() => Nothing());
+
+  const renderSigil = (points, colors, klassName) => {
+    return (
+      <>
+        {points.map((point, idx) => (
+          <div key={idx} className={cn(klassName, 'h9 w9')}>
+            <MaybeSigil patp={point} size={50} colors={colors} />
```
Fang-: As per design feedback, we should at least `title={point}` here.

src/views/Invite/InviteEmail.js
```diff
@@ -142,15 +142,12 @@ export default function InviteEmail() {
...
-      }
-    },
-    [getHasReceived]
-  );
+  const validateHasReceived = useCallback(async email => {
+    // const hasReceived = await getHasReceived(email);
+    // if (hasReceived) {
+    // return HAS_RECEIVED_TEXT;
+    // }
+  }, []);
```
Fang-: Also what's this about? This same logic now lives in `Views/Invite/Email.js` it seems, so maybe a remnant of the move?

src/components/InviteSigilList.js
liam-fitzgerald: done

src/form/EmailChipInput.js
liam-fitzgerald: done

src/views/Invite/InviteEmail.js
liam-fitzgerald: reverted these changes, but maybe better to just turf the old views

Fang-: If they're unused they should just go.

src/lib/useBlockWindowClose.js
```diff
@@ -15,3 +15,19 @@ export default function useBlockWindowClose() {
     };
   });
 }
+
+export function useConditionalBlockWindowClose(condition) {
```
Fang-: This doesn't keep me from going to a different page within Bridge though. Is that something we can prevent?

Fang-:
Changes seem good. Waiting on @jyng for design things.

jyng:
> On the Point page, the sigils feel pretty visually heavy compared to the rest of the page. I'm not sure what we get out of them being there right now. At the very least, maybe they should show names on-hover (with just a title attribute)?

This makes the interface quite cluttered and the added benefit is quite trivial. The designs here are also incorrect — the row shouldn't exist inside the invite group view.

> Also the "pending" highlighting is the only color we have on the page, but it's not really the most important part of the page. Perhaps we want to use grey instead of orange?

Adding members is an important feature we want to highlight, so when people add members they should see positive feedback. Whether or not the color works is a different conversation.

> On the Invite Group screen, the sigils are redundant with the sigil+name list that's underneath them. We probably shouldn't be rendering them there. It's kind of weird, too, that I can't send invites from that dedicated screen, even though I can from the smaller overview on the Points page.

 The sigil list is the detailed version of the sigil row from the previous screen. There’s inaccuracy here: the row should only exist in the home screen not the invite group screen. And yes ‘Add Members’ should be in the Invite Group Screen.

> On the Point page, we might want a separator underneath the "add members" section. It feels a little too intimate with the "admin" button (and the rest of that section) right now.

Yeah this is a good point. We should probably drop it two 2 steps in our spacing scale

jyng:
@liam-fitzgerald 
Some design fixes to be made:
Home page
- 'View' on hover should be a cursor
- Add Members should be a center button

Invite Group page
- There shouldn't be a sigil row
- There should be an Add Members button

Attached designs for reference

home sent
(Ignore the simplified navigation at the bottom, thats still work in progress)

invitegroup

Fang-:
> This makes the interface quite cluttered and the added benefit is quite trivial. The designs here are also incorrect — the row shouldn't exist inside the invite group view.

Fair, but that doesn't answer my concerns about this row existing in the first place.

> Adding members is an important feature we want to highlight, so when people add members they should see positive feedback.

Hmm I'm not sure I understand your translation from "adding members" to "pending invites" here. I get that they're causally related, and I agree that the action of adding members is worth highlighting (as it already is, there's a big black button on an otherwise mostly white page), but I don't see "pending invites" as equally important.

jyng:
Hmm, your questions seem to be regarding why 'invite group' is so emphasized. Let's move this into an email thread, your concern about the design is separate from Liam's work.

src/lib/useBlockWindowClose.js
liam-fitzgerald: I'd need to add lock/unlock functionality to the router

jyng:
@liam-fitzgerald 
I'm noticing in this PR that the tab spacing is off. in Tabs.js line 36 i see we're setting margin/padding, but it's not clear to me why this is for? The spacing seems fine without it.

liam-fitzgerald:
> @liam-fitzgerald
> I'm noticing in this PR that the tab spacing is off. in Tabs.js line 36 i see we're setting margin/padding, but it's not clear to me why this is for? The spacing seems fine without it.

Can you clarify this? I haven't changed the margin/padding in Tabs.js

jyng:
Yeah you haven't changed any spacing on this PR, but I had just noticed this and was curious what it's for because it creates an inset on the tab bar.

image

jyng:
Design Notes on Invite Screen
- The invite number should be 0 for pending invites
- You should see the pending invites when you click into Invite Group
- The textarea should be width: 100%
- We shouldn’t have the hover labels?? It’s just quite clunky

Fang-:
> We shouldn’t have the hover labels?? It’s just quite clunky

They're pretty big, I was thinking just native `title` tag hovers. But then, if we don't have those, what do we even want to accomplish with having the sigils there?

liam-fitzgerald:
> Yeah you haven't changed any spacing on this PR, but I had just noticed this and was curious what it's for because it creates an inset on the tab bar.
> 
> image

It was to center the tab options to match the designs. I'll get rid of that inset

jyng:
> They're pretty big, I was thinking just native title tag hovers. **But then, if we don't have those, what do we even want to accomplish with having the sigils there?**

Just getting a sense of activity inside your invite group is enough information here, the name is optional. However, I think using the title attribute is an improvement here, so let's keep it.

jyng:
LGTM

src/views/Invite/Cohort.js
```diff
@@ -181,7 +181,7 @@ export default function InviteCohort() {
         <Grid.Item full>
           <Flex align="center">
             <Flex.Item>
-              {_acceptedInvites + _pendingInvites} / {_totalInvites}
```
Fang-: Huh, I thought this was correct, in that these are the amount of invites you've actually _used_. idk though.

src/views/Invite/Url.js
```diff
@@ -0,0 +1,106 @@
...
+            progress={0.5}>
+            Generating Invite URL
+          </Grid.Item>
+        );
+      default:
+        return null;
+    }
+  }, [txStatus, error, generateInvite]);
+
+  const url = `https://bridge.urbit.org/#${get(invites, '[0].ticket', '')}`;
```
Fang-: This includes the leading `~`. We've traditionally (with the emails) been sending out invites in the format of `bridge.urbit.org/#some-ticket`, without leading `~`. I think it looks cleaner, and would remove it for the sake of consistency, but don't feel super strongly.

vvisigoth: Agree with @Fang- here, but that's my only quibble.

vvisigoth:
I bless.

vvisigoth:
I bless.

src/views/Invite/Url.js
liam-fitzgerald: fixed

liam-fitzgerald:
ping

Fang-:
Merging blessed PR!

PR #370, opened 2020-01-14
Sigil generator on Point Detail Screen

liam-fitzgerald:
Adds customisable sigil generator to point detail screen.

Drafting as this is based off the Cohort Screens, so blocked pending that merge.


src/views/SigilGenerator.js
```diff
@@ -0,0 +1,124 @@
...
+import { useLocalRouter } from 'lib/LocalRouter';
+import useSigilDownloader from 'lib/useSigilDownloader';
+
+import { composeValidator, buildNumberValidator } from 'form/validators';
+import BridgeForm from 'form/BridgeForm';
+import FormError from 'form/FormError';
+import SubmitButton from 'form/SubmitButton';
+import { NumberInput, ColorInput } from 'form/Inputs';
+
+const BG_COLORS = [
```
jyng: Could these be imported from indigo-tokens?

src/form/Inputs.js
```diff
@@ -196,3 +197,53 @@ export function EmailInput({ ...rest }) {
     />
   );
 }
+
+export const ColorInput = ({
```
jyng: Hm, why are we not using the indigo-react/inputs here?

src/form/SubmitButton.js
```diff
@@ -14,6 +14,7 @@ export default function SubmitButton({
   handleSubmit,
   ...rest
 }) {
+  const form = useFormState();
```
jyng: Could we try using the indigo-react/buttons here?

src/form/Inputs.js
```diff
@@ -197,53 +196,3 @@ export function EmailInput({ ...rest }) {
     />
   );
 }
-
-export const ColorInput = ({
```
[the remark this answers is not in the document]
liam-fitzgerald: So move this back into form/Inputs.js ? Indigo-react doesn't currently contain a color picker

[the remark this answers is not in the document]
jyng: Just off the top of my head, I would move color picker src/components (which im assuming are local components) and then inside the colorPicker.js I would use atomic components from indigo-react.

Issue #353, opened 2019-12-10
Reticketing using metamask broken


jyng:
<img width="643" alt="Screen Shot 2019-12-10 at 12 29 28 PM" src="https://user-images.githubusercontent.com/9557907/70566364-c9798100-1b48-11ea-8136-f5dbdb9315f4.png">

This is in Admin > Reticket

jyng:
<img width="804" alt="Screen Shot 2019-12-10 at 12 31 43 PM" src="https://user-images.githubusercontent.com/9557907/70566527-178e8480-1b49-11ea-8e03-cd398c9cba32.png">

Fang-:
Was this logged in with metamask?

jyng:
Correct

vvisigoth:
Drat!
—
~poldec-tonteg
http://urbit.org

shrugs:
confirmed: using the mnemonic from the same account is a workaround for the moment

g-a-v-i-n:
also true for trezor, not sure why signin matters.

Fang-:
Fixed in #371.

Issue #293, opened 2019-09-26
Reticket failing on Metamask/Trezor/Bridge


liam-fitzgerald:
Reticket doesn't work when we don't have control of the private key. In addition, we need to batch the transactions so the user isn't inundated with requests.

Fang-:
> Reticket doesn't work when we don't have control of the private key.

Is this because the flow is just poorly/inflexibly programmed? (ie assumes we're in a master-ticket-login situation)

liam-fitzgerald:
Yeah it assumes we have a private key and can sign transactions in browser, as opposed to delegating the signing elsewhere (metamask, Trezor, etc.)

Fang-:
It seems that neither [Trezor](https://github.com/trezor/connect/blob/develop/docs/methods/ethereumSignTransaction.md) nor [Ledger](https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-app-eth) support batch-signing transactions.

Fang-:
Fixed in #371.

PR #371, opened 2020-01-14
fix reticket flow to handle noncustodial wallets

liam-fitzgerald:
Reticket doesn't work when we don't have control of the private
key. This change makes the reticket flow use the signTransaction
and sendSignedTransaction helpers, which have special cases for
handling wallets where we don't have custody of the private key.

Also disables gas tank for metamask users.
Addresses #293 #353


src/lib/reticket.js
```diff
@@ -143,26 +148,44 @@ export async function reticketPointBetweenWallets({
...
-    return {
-      raw: hexify(stx.serialize()),
-      signed: stx,
-    };
-  });
+  console.log(txs);
+
+  let txPairs = await Promise.all(
+    txs.map(async (tx, idx) => {
+      const stx = await signTransaction({
```
Fang-: I suppose the UX for this in the previously-not-supported cases is just repeated pop-ups/prompts to do the signing?

vvisigoth:
I just tested with Ledger and doestn' work, I've attached errors. @liam-fitzgerald do you have a hardware wallet to test this with? If not, email me.

Screen Shot 2020-01-14 at 4 16 23 PM

Screen Shot 2020-01-14 at 4 16 49 PM

Fang-:
@vvisigoth Just in case, can you confirm you're running on latest Ledger firmware? #364 updated the dependencies for Ledger, that may or may not break older firmwares.

vvisigoth:
OK, updated firmware and now I have different problems.

Screen Shot 2020-01-14 at 4 37 52 PM

Screen Shot 2020-01-14 at 4 38 01 PM

vvisigoth:
Talked to @Fang- offline. He's double checking we have correct firmware support.

vvisigoth:
OK, testing on Ledger. I fixed some of these issues by using the latest firmware here https://github.com/urbit/bridge/pull/372 . Still running into the "U2_LEDGER DEVICE INELLIGIBLE" error. The error seems to pop up after I've approved the "Fields" of the tx on the ledger device, but before I've been asked to "Approve and Send" the first transaction.

Other transactinos still work, just falling over on the Reticket flow.

src/lib/reticket.js
Fang-: Perhaps this is too eager in sending transactions to the device, and we want to actually run this in sequence? ie, maybe not rely on `.map` here? Not sure, but seems like something worth trying if we don't have any other leads.

liam-fitzgerald: thanks for the heads up. yeah the popups/prompts just repeat

liam-fitzgerald:
Latest commit works with my ledger, would you be able to source a trezor for testing?

Fang-:
@vvisigoth I won't have access to a Trezor (or any hw wallet) until next week. Can you give this a test run?

liam-fitzgerald:
ping @Fang- @vvisigoth do either of you have access to a trezor yet?

Fang-:
Just tested with a Trezor. Doing this from a just-reloaded Bridge page works fine. When I tried doing this directly after accepting an incoming transfer, I got the following error:

image

But all those transactions still succeeded, so not sure what was up here. This also seems like a more general issue (potentially stale nonces) rather than a flow-specific one, and/or might have been present in the first place.

Merging this as confirmed fixed!

PR #385, opened 2020-02-03
Invite: clean up old screens, fix gas tank usage bug

Fang-:
Primarily, 811b75f: we weren't passing in any transaction objects, leading to failing funds requests. #369 really did a number on the logic here...

This also cleans up the files deprecated by that PR. I think we no longer have dead invite code lying around anymore now. @liam-fitzgerald please confirm.


vvisigoth:
I've looked at this, but haven't tested against a dev setup of email sender (I had it set up on computer that died). Have you done so, @Fang-

Fang-:
This doesn't touch email sending, but I did test this on a ropsten setup (which is closer to realnet than local is).

vvisigoth:
Sorry, typo, meant gas tank.

Fang-:
Either case, yes, this was tested to work against ropsten.

PR #389, opened 2020-02-06
greeting: point to new invite screen

Fang-:
The `INVITE` route got renamed to `INVITE_COHORT` during #369. Just `INVITE` points to the sigil page somehow...


Fang-:
On reflection, maybe this link should be removed entirely? We show the invite functionality on the point overview page now, which is where this would show up.

liam-fitzgerald:
Remove is probably the way to go

Fang-:
Oops! Busy times are no excuse to let something this simple slip down the cracks.

Went and removed the entire thing.

PR #395, opened 2020-02-13
Change sponsor screens

liam-fitzgerald:
Have implemented the new change sponsor screens in preparation for the hosting release. Restructures the Point view to include an Urbit OS routing tree.


src/views/UrbitOS/ChangeSponsor.js
```diff
@@ -0,0 +1,199 @@
...
+    [validateFormAsync]
+  );
+
+  const validate = useMemo(() => {
+    const sponsorSize =
+      azimuth.getPointSize(point) === azimuth.PointSize.Star ? 1 : 2;
+    return composeValidator(
+      {
+        sponsor: buildPointValidator(sponsorSize, [
+          validateMinimumPatpByteLength(sponsorSize),
```
Fang-: You want it to be the exact size required, not a minimum size, right?

src/views/UrbitOS/Home.js
```diff
@@ -0,0 +1,188 @@
...
+        {showSponsor && (
+          <>
+            <Grid.Item
+              full
+              as={ForwardButton}
+              detail="A sponsor finds new peers in your network"
+              accessory={<u>Change</u>}
+              onClick={goChangeSponsor}>
+              <span className="mono">{sponsor}</span>
+              <span className="f7 bg-black white p1 ml2 r4">SPONSOR</span>
```
Fang-: Why is this here? This kind of element isn't used anywhere else yet, it jumps out and I'm not sure why. cc @jyng

src/views/UrbitOS/Home.js
```diff
@@ -0,0 +1,188 @@
...
+          onClick={toggleShowKeys}>
+          View Networking Keys
+        </Grid.Item>
+        {showKeys && <Grid.Item full as={NetworkingKeys} point={point} />}
+      </Grid>
+    </>
+  );
+}
+
+// eslint-disable-next-line
```
Fang-: ?

src/views/UrbitOS/Home.js
```diff
@@ -0,0 +1,188 @@
...
+function Hosting() {
```
Fang-: All of the hosting stuff here seems like stubs, and isn't used/displayed anywhere. We should probably leave the code out for now, especially for this PR.

src/views/UrbitOS/Home.js
jyng: We want to start merging some design language between Bridge and OS1. What other ways could we highlight that the star is the sponsor?

jyng:
1) Did we decide to remove 'Unset'?

2)'Login Code' should live under Urbit OS and not Network. The alert text here should be in 'sans' and not 'mono'

3) Urbit OS/Change Sponsor: The input box placeholder should be a star patp. It should also be a galaxy patp for stars.

4) With regards to the email about 'sponsor request list' scope, I've added a new screen in Figma to display 'sponsor request' (requestor's view). 

5) @Fang- I designed a sponsor screen for them to request, but either it hasn't be done or its WIP. @liam-fitzgerald ?

src/views/UrbitOS/Home.js
Fang-: > We want to start merging some design language between Bridge and OS1.

Oh, that's fair game I suppose. It had also somehow escaped my no

Good question re alternatives. I would like for this to mimic the login code design, which would put "Sponsor" in the head, and `~value` in the detail, but that would mean getting rid of the explanation in the detail... Maybe "Sponsor: `~value`" in the head then?

More broadly speaking, imo, this kind of small block next to text feels a lot like a category or warning or other kind of status indicator, and not at all like it's _the_ canonical descriptor for the data I'm looking at.

src/views/UrbitOS/Home.js
```diff
@@ -0,0 +1,188 @@
...
+        <Grid.Item full className="mv7 f5">
+          Network
+        </Grid.Item>
+        <Grid.Divider />
+        {showSponsor && (
+          <>
+            <Grid.Item
+              full
+              as={ForwardButton}
+              detail="A sponsor finds new peers in your network"
```
jyng: The detail of a Grid.Item should always have a mt2, unless in special cases.

liam-fitzgerald:
closing, this work will arrive w/ hosting

Issue #397, opened 2020-02-20
Transaction failed on dev environment while testing linear release


vvisigoth:
![Bridge](https://user-images.githubusercontent.com/1609955/74891970-3a08ff00-533d-11ea-8957-ac4a65caa9e3.gif)

fails with this error in the console

<img width="670" alt="Screen Shot 2020-02-19 at 5 26 48 PM" src="https://user-images.githubusercontent.com/1609955/74891997-49884800-533d-11ea-958a-6066ef38b388.png">

@liam-fitzgerald

liam-fitzgerald:
Weird, error trace means we're trying to do an invalid transaction on the EVM. Will try and repro

liam-fitzgerald:
So it's erroring because the linear star release contract doesn't have as many stars deposited as it should, i.e. the contract promises 10 stars, but only 5 were actually deposited. Will rectify, but it shouldn't be urgent.

PR #402, opened 2020-02-27
fix partially deposited release contracts

liam-fitzgerald:
Previously, if the contracts contained less stars than were promised,
then the starRelease store would misreport the number of stars
currently available for withdraw and allow for an erroneous tx to be
attempted.

Fixes: #397

### DIFF
--- a/src/lib/starRelease.js
+++ b/src/lib/starRelease.js
@@ -65,21 +65,36 @@ function generateLinearWithdrawTxs(contracts, amount, to) {
 }
 
 export async function getConditional(contracts, address) {
-  const [commitment, batches] = await Promise.all([
+  const [commitment, batches, remaining] = await Promise.all([
     conditionalSR.getCommitment(contracts, address),
     conditionalSR.getBatches(contracts, address),
+    conditionalSR.getRemainingStars(contracts, address),
   ]);
 
   if (commitment === null || batches === null) {
     return { total: 0, available: 0, withdrawn: 0, batchLimits: [] };
   }
 
-  const { total } = commitment;
-  const withdrawLimits = await Promise.all(
-    batches.map((_, idx) =>
-      conditionalSR.getWithdrawLimit(contracts, address, idx)
-    )
-  );
+  const forfeited = await conditionalSR.getForfeited(contracts, address);
+  const total = remaining.length;
+  let balance = total;
+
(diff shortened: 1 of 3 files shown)


src/lib/starRelease.js
```diff
@@ -65,21 +65,36 @@ function generateLinearWithdrawTxs(contracts, amount, to) {
...
-    batches.map((_, idx) =>
-      conditionalSR.getWithdrawLimit(contracts, address, idx)
-    )
-  );
+  const forfeited = await conditionalSR.getForfeited(contracts, address);
+  const total = remaining.length;
+  let balance = total;
+
+  const withdrawLimits = await batches.reduce(async (p, _, idx) => {
+    const acc = await p;
```
Fang-: What is `acc` supposed to stand for? Something like `batchSize` would be much more readable. (Also why do we assume they're promises? Shouldn't `batches` be an array of numbers?)

src/lib/starRelease.js
```diff
@@ -65,21 +65,36 @@ function generateLinearWithdrawTxs(contracts, amount, to) {
...
-  const withdrawLimits = await Promise.all(
-    batches.map((_, idx) =>
-      conditionalSR.getWithdrawLimit(contracts, address, idx)
-    )
-  );
+  const forfeited = await conditionalSR.getForfeited(contracts, address);
+  const total = remaining.length;
+  let balance = total;
+
+  const withdrawLimits = await batches.reduce(async (p, _, idx) => {
```
Fang-: I don't think doing the balances for the batches in sequence like this (and potentially saying "not enough balance for this one batch") is entirely accurate, since you can withdraw from any batch: you don't necessarily empty them in order. Probably not a huge deal to pretend that's the case though.

src/lib/starRelease.js
liam-fitzgerald: acc -> accumulator.
the product of batches.reduce is a promise returning an array of numbers.

src/lib/starRelease.js
liam-fitzgerald: the UI doesn't support out of order withdrawal so it's kinda moot

src/lib/starRelease.js
Fang-: Oops, I was reading this as it it were a `.map`.

PR #407, opened 2020-03-04
New family view

liam-fitzgerald:
Adds new family view


src/views/Points.js
```diff
@@ -18,11 +19,175 @@ import pluralize from 'lib/pluralize';
...
 import Footer from 'components/Footer';
 import { ForwardButton } from 'components/Buttons';
 import CopiableAddress from 'components/CopiableAddress';
 import NavHeader from 'components/NavHeader';
+import Tabs from 'components/Tabs';
+import PointList from 'components/PointList';
+
+import Locked from './Points/Locked';
+
+function Pending({ incomingPoints, outgoingPoints }) {
```
Fang-: Kinda weird that `Locked` gets its own file, but `Pending` doesn't. It's "only" a 100 lines or so, but that's still nearly a quarter of `Points`. (Guess you'd want to move `Active` into its own file then also, for consistency, but it's so small that it seems hardly worth it.)

src/views/Points/Locked.js
```diff
@@ -0,0 +1,192 @@
+import React, { useCallback, useMemo } from 'react';
```
Fang-: Isn't this file more or less just `views/Release/Locked`? We can deprecate/remove that one then, right?

Fang-:
ping on these comments and the merge conflict @liam-fitzgerald

PR #409, opened 2020-03-10
New login screen to make non-master ticket logins easier to navigate

jyng:
Screen Shot 2020-03-09 at 7 40 59 PM

image

Screen Shot 2020-03-09 at 7 40 17 PM

### DIFF
--- a/src/views/Login.js
+++ b/src/views/Login.js
@@ -78,7 +78,7 @@ export default function Login() {
               className="b-solid b1 b-black"
               center
               onClick={() => setisOther(true)}>
-              Metamask, Mnemonic, Hardware Wallet, etc...
+              Metamask, Mnemonic, Hardware Wallet, etc
             </Grid.Item>
             <Grid.Item full onClick={goToActivate} className="mv10 t-center f6">
               <span className="gray4">New Urbit ID? </span>
--- a/src/views/Login.js
+++ b/src/views/Login.js
@@ -63,7 +63,9 @@ export default function Login() {
       <Grid>
         <Grid.Item full as={Crumbs} routes={[{ text: 'Bridge' }]} />
         <Grid.Item full as={Text} className="flex justify-center mt4">
-          <Grid.Item as={Text} className="gray3">Urbit ID /&nbsp;</Grid.Item>
+          <Grid.Item as={Text} className="gray5">
+            Urbit ID /&nbsp;
+          </Grid.Item>
           <Grid.Item as={Text}>Login</Grid.Item>
         </Grid.Item>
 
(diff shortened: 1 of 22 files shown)


matildepark:
Surely just trailing ellipses from “Hardware Wallet…” is better than “etc.” in a button?

— 
~haddef-sigwen
https://urbit.org

galenwp:
Yeah 'etc...' seems a bit odd. 'Other options (Metamask, Ledger, etc)' or similar.

vvisigoth:
YEAH
—
~poldec-tonteg
http://urbit.org

vvisigoth:
Otherwise, i think this is vast improvement, thnx. Approved based on changing ellipsis.

jyng:
Fixed

jyng:
Edited

src/views/Login.js
```diff
@@ -61,37 +59,54 @@ export default function Login() {
...
+            </Grid.Item>
+          </>
+        )}
+        {isOther && (
+          <Grid.Item
+            as={LinkButton}
+            onClick={() => setisOther(false)}
+            full
+            className="t-center underline f6 mt8">
+            Back
```
Fang-: Should this just say "Master Ticket" or similar instead, to have more symmetry with the "Metamask, Mnemonic..." button?

src/views/Login.js
```diff
@@ -61,37 +59,54 @@ export default function Login() {
...
+              as={Button}
+              className="b-solid b1 b-black"
+              center
+              onClick={() => setisOther(true)}>
+              Metamask, Mnemonic, Hardware Wallet...
+            </Grid.Item>
+            <Grid.Item full onClick={goToActivate} className="mv10 t-center f6">
+              <span className="gray4">New Urbit ID? </span>
+              <LinkButton>Activate</LinkButton>
+            </Grid.Item>
```
Fang-: Not sure why we don't just show this on both pages (ie regardless of `isOther`), but I can see how it's only strictly needed for the landing conditions.

(Also the space after "ID?" bother me. (^: )

src/views/Login/Ticket.js
```diff
@@ -168,71 +227,58 @@ export default function Ticket({ className, goHome }) {
...
 
             <Grid.Item full as={FormError} />
 
-            <Grid.Item full as={SubmitButton} handleSubmit={handleSubmit}>
+            <Grid.Item
+              full
+              as={SubmitButton}
+              center
+              handleSubmit={handleSubmit}>
               {isWarning =>
```
Fang-: Maybe this is as good a time as any to tear this out. See also #405. (But we might want further discussion on that.)

jyng: Yeah while I was styling this, I felt like it was an unnecessary layer of abstraction, but what do I know lol

src/views/Login.js
jyng: Yeah, this crossed my mind, but I felt like this page is for advanced users who are familiar with crypto. On the flipside, we haven't done a good job of getting people to use Master Ticket (even if they bought a planet from a third-party)

Issue #346, opened 2019-12-02
After personal breach, keyfile downloaded is old


vvisigoth:
I did a personal breach (from  3 --> 4) and immediately DL'ed the keyfile. Keyfile was life 3.

Went back to family view, renavigated to the keyfile page and it DL'ed the correct one.

Fang-:
Hmm, cannot reproduce. Immediately downloading the keyfile on-rekey always gives me a result consistent with future downloads. I'm assuming this was with a master ticket login? That's what I tried with, as most other methods give non-deterministic keyfiles, that aren't redownloadable. (Though they should be, within the same session... did we intentionally remove that?)

galenwp:
Just experienced this. Not sure what I can offer in terms of repro — but exactly as described by vvisigoth.

galenwp:
I wonder whether it can somehow be caused by ETH network congestion

jtobin:
Encountering this again myself.

matildepark:
Seems like this is the same issue as #137. Just noting ~

Fang-:
Closing as duplicate of #137. See also the question there: anyone know of new cases of this since #423?

PR #423, opened 2020-03-24
generate keyfile as part of network key tx

liam-fitzgerald:
We generate the keyfile as part of the network key transaction to
hopefully prevent downloading stale.

Hopefully mitigates #346 and #137 although I am unable to reproduce
it.

### DIFF
--- a/src/views/Admin/AdminNetworkingKeys.js
+++ b/src/views/Admin/AdminNetworkingKeys.js
@@ -74,6 +74,7 @@ function useSetKeys() {
 
   const {
     available: keyfileAvailable,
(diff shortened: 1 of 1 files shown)


Fang-:
Can you explain the reasoning behind this fix? I'm having trouble seeing what problematic case this solves, and how exactly it solves that.

liam-fitzgerald:
If the new keyfile takes a while to generate, then the downloadkeyfilebutton will serve a stale keyfile while the new keyfile is generating. Hence we don't show the downloadkeyfile button until we've finished generating a new one

PR #427, opened 2020-03-24
router: add senate to routeNames

### LINKED ISSUE #424: Senate redirects to Sigil?
on localhost (master)
Screen Shot 2020-03-24 at 11 29 55 AM

Fang-:
Fixes #424.

This was missing here, even though we used `ROUTE_NAMES.SENATE` in `router.js`. Apparently `undefined` points to the sigil page? Huh.


liam-fitzgerald:
we should probably add sigil generator to routenames as well

Fang-:
Good catch, didn't even realize that was missing too.

PR #431, opened 2020-03-27
Auth Token support

liam-fitzgerald:
Adds auth token support. Generated as part of login. Supports all wallet types. 

Also reworks metamask support so that if present and enabled, it is used as a provider for web3, instead of the infura provider.

Some questions:
- Copy re: login methods, hardware/metamask wallet users will need to sign the auth request before login
- Thinking about maybe deprecating direct hardware support and making hardware wallets go through metamask. If we did this we could use [EIP712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) which is better suited to our usecase.

### DIFF
--- a/src/lib/authToken.js
+++ b/src/lib/authToken.js
@@ -1,0 +1,14 @@
+import * as secp256k1 from 'secp256k1';
+
+import { keccak256 } from './wallet';
+
+const MESSAGE = 'Bridge Authentication Token';
+
+export const getAuthToken = ({ wallet, walletType, walletHdPath, web3 }) => {
+  const { signature } = secp256k1.sign(
+    keccak256('\x19Ethereum Signed Message:\n' + MESSAGE.length + MESSAGE),
+    wallet.privateKey
+  );
+
+  return signature;
+};
--- a/src/lib/authToken.js
+++ b/src/lib/authToken.js
@@ -20,15 +20,6 @@ function signMessage(privateKey) {
   return ethSignature;
 }
 
-function signatureToHex(v, r, s) {
-  const ethSignature = new Uint8Array(65);
-  ethSignature.set(Buffer.from(r, 'hex'));
-  ethSignature.set(Buffer.from(s, 'hex'), 32);
-  ethSignature[64] = v;
-
-  return `0x${Buffer.from(ethSignature).toString('hex')}`;
-}
-
 export const getAuthToken = async ({
   wallet,
   walletType,
@@ -39,8 +30,7 @@ export const getAuthToken = async ({
     return web3.eth.personal.sign(MESSAGE, wallet.address, '');
   }
   if (walletType === WALLET_TYPES.LEDGER) {
-    const { v, r, s } = await ledgerSignMessage(MESSAGE, walletHdPath);
-    return signatureToHex(v, r, s);
+    return ledgerSignMessage(MESSAGE, walletHdPath);
   }
   if (walletType === WALLET_TYPES.TREZOR) {
     return trezorSignMessage(MESSAGE, walletHdPath);
--- a/src/lib/authToken.js
+++ b/src/lib/authToken.js
@@ -1,10 +1,13 @@
 import * as secp256k1 from 'secp256k1';
 
-import { keccak256 } from './wallet';
+import { keccak256, WALLET_TYPES } from './wallet';
 
 const MESSAGE = 'Bridge Authentication Token';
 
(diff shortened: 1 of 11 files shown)


src/lib/authToken.js
```diff
@@ -0,0 +1,54 @@
...
+  walletType,
+  walletHdPath,
+  web3,
+}) => {
+  if (walletType === WALLET_TYPES.METAMASK) {
+    return web3.eth.personal.sign(MESSAGE, wallet.address, '');
+  }
+  if (walletType === WALLET_TYPES.LEDGER) {
+    const { v, r, s } = await ledgerSignMessage(MESSAGE, walletHdPath);
+    return signatureToHex(v, r, s);
```
Fang-: Would it make sense to move this `signatureToHex` into `ledgerSignMessage`, or nah? Seems nice to have more-consistent interfaces for these across the different wallet types.

src/store/wallet.js
```diff
@@ -46,6 +51,32 @@ function _useWallet(initialWallet = Nothing(), initialMnemonic = Nothing()) {
   const [networkSeed, setNetworkSeed] = useState(Nothing());
   const [networkRevision, setNetworkRevision] = useState(Nothing());
 
+  const [authToken, setAuthToken] = useState(Nothing());
```
Fang-: I thought nobody liked the `maybe` objects and we were gonna not use them for new code? They're relatively common across the rest of `wallet.js` still though, so probably better for local consistency, but it's exposed, too...

src/store/wallet.js
```diff
@@ -46,6 +51,32 @@ function _useWallet(initialWallet = Nothing(), initialMnemonic = Nothing()) {
...
+
+  const { web3 } = useNetwork();
+
+  useEffect(() => {
+    (async () => {
+      if (
+        !Just.hasInstance(wallet) ||
+        !Just.hasInstance(web3) ||
+        NONCUSTODIAL_WALLETS.has(walletType)
```
Fang-: What's with this last check? Doesn't `getAuthToken` support noncustodial wallets just fine?
It's not immediately obvious why we insist on doing `setAuthToken` within the login page code for noncustodial wallets.

liam-fitzgerald: Because the signing for noncustodial wallets is async and can block indefinitely and we want to prevent login until the signing occurs. To do this we await getAuthToken() inside the Login page for all noncustodial wallets, and generate the signing on login for the others.

src/store/wallet.js
liam-fitzgerald: Yeah maybe kind of sucks but it's pervasive across the entire codebase. We ought to re-evaluate how/where/when we store data in the general but I don't think this PR is the right place to have that conversation

liam-fitzgerald:
also just remembered I don't have a trezor, would you be able to test that for me?

src/lib/trezor.js
```diff
@@ -36,4 +36,19 @@ const trezorSignTransaction = async (txn, hdpath) => {
   return txn;
 };
 
+export const trezorSignMessage = async (message, hdPath) => {
+  const sig = await TrezorConnect.ethereumSignTransaction({
```
Fang-:
```suggestion
  const sig = await TrezorConnect.signMessage({
```

liam-fitzgerald:
Another pass?

src/views/Login/Ledger.js
```diff
@@ -164,7 +164,8 @@ export default function Ledger({ className, goHome }) {
         application. If you're running on older firmware, make sure the "browser
         support" option is turned on. To sign transactions, you'll also need to
         enable the "contract data" option. You will be prompted to sign a
-        message. This allows Bridge to operate correctly.
+        message. This allows Bridge to operate correctly. Never sign this hash
+        outside of Bridge.
```
Fang-, line 168: > Never sign this hash

Should be "sign this message", right? `Trezor.js` does this correctly.

This warning feels kinda ehh, but there really isn't much more we can do here. Maybe the message should be "urbit bridge exclusive token" or similar to drive this home, and look fishy when prompted for by other sites?

liam-fitzgerald: Ledgers display a hash when a message is requested to be signed, so I thought consistency with the hardware trumped copy consistency.

Fang-: Fair thinking, but iirc Trezors display (the beginning of) the message itself, so those two aren't consistent even with each other. I think in such cases it's best to err on the side of describing the underlying action accurately? If they have trouble, you want them to search for "ledger sign message", etc etc.

PR #432, opened 2020-03-31
One Click Login

liam-fitzgerald:
Dependant on urbit/urbit#2619 to work correctly.
Allows users to login to their points from bridge. Requires HTTPS and DNS to be setup. HTTPS is unfortunately a necessity because the browser tends to have a fit about mixed content, especially with fetches in the background.


galenwp:
I'm curious what the state of this is — it'd be amazing to be able to demo this.

Fang-:
The urbit-side PR has been merged, but not yet OTA'd I don't think. Also this PR (besides still being marked draft? @liam-fitzgerald what's pending?) depends on and targets the hosting PR. That's getting real close, but I wouldn't depend on it going in before demo tomorrow.

You might be able to get away with checking out this branch and just building/running it locally.

liam-fitzgerald:
You certainly could check it out and run it locally. If you'd like help, feel free to reach out. 

It's marked draft because I don't really like the structure of it, I wanted to suss out future requirements for hosting to see if I'm taking this in the correct direction. Still works though.

galenwp:
Got it, got it. It's okay to not be able to get it in. I'll talk to Ed and see what he wants to do.

PR #433, opened 2020-03-31
wallet-gen: add deterministic invite code generator

liam-fitzgerald:
Adds a deterministic ticket generator. Uses the point and a seed to generate a ticket by hashing the seed and salting with the point to be generated. Using a hash as a deterministic RNG should be fine because we aren't really going deep with the entropy.

### DIFF
--- a/src/lib/walletgen.js
+++ b/src/lib/walletgen.js
@@ -16,14 +16,16 @@ import {
 
 const SEED_LENGTH_BYTES = SEED_ENTROPY_BITS / 8;
 
+const getEntropyBits = point =>
+  point < MIN_STAR
+    ? GALAXY_ENTROPY_BITS
+    : point < MIN_PLANET
+    ? STAR_ENTROPY_BITS
+    : PLANET_ENTROPY_BITS;
+
 // returns a promise for a ticket string
 export const makeTicket = point => {
-  const bits =
-    point < MIN_STAR
-      ? GALAXY_ENTROPY_BITS
-      : point < MIN_PLANET
-      ? STAR_ENTROPY_BITS
-      : PLANET_ENTROPY_BITS;
+  const bits = getEntropyBits(point);
 
   const bytes = bits / 8;
   const some = new Uint8Array(bytes);
@@ -46,12 +48,7 @@ export const makeTicket = point => {
 };
 
 export const makeDeterministicTicket = (point, seed) => {
-  const bits =
-    point < MIN_STAR
-      ? GALAXY_ENTROPY_BITS
-      : point < MIN_PLANET
-      ? STAR_ENTROPY_BITS
-      : PLANET_ENTROPY_BITS;
+  const bits = getEntropyBits(point);
 
   const bytes = bits / 8;
 
--- a/src/lib/walletgen.js
+++ b/src/lib/walletgen.js
@@ -16,7 +16,7 @@ import {
 
 const SEED_LENGTH_BYTES = SEED_ENTROPY_BITS / 8;
 
-const getEntropyBits = point =>
+const getTicketBitSize = point =>
   point < MIN_STAR
     ? GALAXY_ENTROPY_BITS
     : point < MIN_PLANET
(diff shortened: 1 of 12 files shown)


src/lib/walletgen.js
```diff
@@ -44,6 +45,27 @@ export const makeTicket = point => {
   });
 };
 
+export const makeDeterministicTicket = (point, seed) => {
+  const bits =
+    point < MIN_STAR
+      ? GALAXY_ENTROPY_BITS
+      : point < MIN_PLANET
+      ? STAR_ENTROPY_BITS
+      : PLANET_ENTROPY_BITS;
```
Fang-: This feels like it should be its own utility function. It's used in `makeTicket` also.

liam-fitzgerald:
trying for smaller PRs :) Just wanted a signoff on the actual crypto alg itself.

src/lib/walletgen.js
```diff
@@ -15,14 +16,16 @@ import {
 
 const SEED_LENGTH_BYTES = SEED_ENTROPY_BITS / 8;
 
+const getEntropyBits = point =>
```
Fang-: Perhaps pedantic, but I'd `getTicketBitSize`, since we use it to determine that, not the "entropy" that goes in.

liam-fitzgerald: teach me the dark arts of naming things good

Fang-: [Use this forbidden txt wisely.](https://github.com/urbit/urbit/blob/master/doc/spec/flw.txt)

liam-fitzgerald:
Whole thing is basically done, just waiting for designs from @jyng 

can I get a pass on this @Fang- 

demo: 
(design will change )
Screen Shot 2020-04-15 at 3 40 06 pm

src/views/Login/Ledger.js
```diff
@@ -78,7 +78,7 @@ export default function Ledger({ className, goHome }) {
           walletHdPath: addHdPrefix(values.hdpath),
         });
 
-        setAuthToken(authToken);
+        setAuthToken(Just(authToken));
```
Fang-, line 81: I can't remember why this is a `Maybe`. Don't we have this in all post-login cases?  
(Feel like I asked this already at some point, but didn't take notes well enough. Sorry!)

src/views/Invite/Cohort.js
```diff
@@ -164,31 +166,50 @@ function CohortMemberExpanded({ point, className, ...rest }) {
...
+
+      if (owner.keys.address !== _details.transferProxy) {
+        setCode(Just(null));
+      } else {
+        setCode(Just(ticket));
+      }
     };
     fetchWallet();
-  }, [authToken, point, active]);
+  }, [codeVisible]);
```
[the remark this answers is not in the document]
liam-fitzgerald: that's how we only derive it on click to reveal

src/views/Login/Ledger.js
liam-fitzgerald: It was more for consistency with the rest of useWallet

src/views/Invite/Cohort.js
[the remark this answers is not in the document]
Fang-: But we only derive if the check on those other things passes, right? And they might (theoretically) change independently from `codeVisible`...
If the concern here is around re-deriving even though we already know it, maybe we can check for the value of `code` itself too?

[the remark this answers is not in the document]
liam-fitzgerald: good point. i'll add code to the deps list. This is actually symptomatic of a deeper issue where the every access to the point cache is considered changed after any of the points change. I'm unsure whether or not this requires us to drastically change the way that we hold data in the cache though.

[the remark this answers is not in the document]
Fang-: > the every access to the point cache is considered changed after any of the points change

Oof, yes. This is likely (part of) what's causing performance issues for logins with many related assets. Definitely worth thinking about!

Fang-:
big oops on letting this sit for so long. gonna put it in!

Issue #422, opened 2020-03-21
Error Report: Cannot create property 'message' on string 'Node error: {"code":-32000,"message":"nonce too low"}'


ixv:
Got this while trying to verify a master ticket from an invite code.

## Error

```
TypeError: Cannot create property 'message' on string 'Node error: {"code":-32000,"message":"nonce too low"}'
    at renderAdditionalInfo (https://bridge.urbit.org/static/js/main.21612c58.chunk.js:2261:385)
    at PassportTransfer (https://bridge.urbit.org/static/js/main.21612c58.chunk.js:2261:1914)
    at dh (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:159040:7)
    at Jh (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:159516:7)
    at mj (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:161789:86)
    at jj (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:160777:11)
    at Z (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:160634:15)
    at $i (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:160493:16)
    at https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:158219:17
    at push.exports.unstable_runWithPriority (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:162914:12)
```

In Bridge version 2.3.5.

liam-fitzgerald:
@Fang- nonce is usually gas tank shenanigans can you check?

Fang-:
Looks like the error printing logic in `renderAdditionalInfo` (in both `PassportTransfer` and `ResetExecute`) doesn't account for the case where `generalError` is a string rather than an object.

That said, the error that was trying to get printed should have been caught if we made use of the gas tank, and shouldn't occur if we didn't.

The context is weird, too. @ixv you got this during the verification step, or directly after, when it's telling you to wait for a bit?

liam-fitzgerald:
[...]

PR #441, opened 2020-04-05
fix: use visibility hidden complementary column

shrugs:
...so that it takes up the same amount of space

before:

image

after:

image

### DIFF
--- a/src/components/View.js
+++ b/src/components/View.js
@@ -90,7 +90,13 @@ function View({
       </Flex.Item>
 
       <Flex.Item style={{ visibility: 'hidden' }}>
-        {showBackButton && !backIsLogout && <MiniBackButton />}
+        {showBackButton && !backIsLogout && (
+          <MiniBackButton
+            hpadding={!isMobile}
+            vpadding={isMobile}
+            isExit={backIsLogout}
+          />
+        )}
       </Flex.Item>
     </Flex>
   );
--- a/src/components/View.js
+++ b/src/components/View.js
@@ -89,7 +89,9 @@ function View({
         <Flex.Item as={Footer.Target} />
       </Flex.Item>
 
-      <Flex.Item style={{ width: '48px' }} />
+      <Flex.Item style={{ visibility: 'hidden' }}>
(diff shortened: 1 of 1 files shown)


jyng:
I think you mean to make it look like the 'before'?, correct?

shrugs:
notice the spacing to the right of the entire content container, and how the container is correctly centered in the after photo. the problem was that the right spacer div was a constant with of 48px but the back button is smaller

(i took a look at handling the passport styling in #442, if that's what you're looking at)

jyng:
Ah okay, I see what you mean now and how there's two separate issues. Thanks for noting that.

Would it be easier to match the buttons with the horizontal spacing of the card?

shrugs:
if we use the aspect ratio approach in #422 they'd be aligned, yeah. but otherwise imo it makes more sense for the passport to respond to its available space than for it to drive the available width for the rest of the UI

jyng:
I don't think the aspect ratio approach looks acceptable on mobile, i feel like this might require editing how many dots are shown on mobile. This seems like considerable amount of work?

That's why I feel like we could set a max-width (derived from card width) and have everything else wrap to that. It seems much easier. No?

shrugs:
i'm not sure i fully understand your suggestion, but if you're confident go for it

as i understand it, if the width of the passport is driving the available space for the rest of the content, we will in most cases end up with either extra padding between the content and the edges of the screen (in the cases where the passport is smaller) and content radically overflowing the mobile screen (which is what would happen if we chose to make the content the width of the passport, for example in the first image of #442)

another option, to avoid squishing the hash, is to just have a bunch of sigil and text sizing breakpoints for each screen, so that it's best effort on each screen size. the passport might overflow (as it does right now) but at least it wouldn't overflow _as much_.

Fang-:
@jyng I'm guessing this is deprecated by current "new bridge design" work?

jyng:
I thought this was fixed?

I'm looking on the live site and not seeing this design issue anymore.

Fang-:
Okay, closing then! Thanks regardless, @shrugs. (:

PR #444, opened 2020-04-06
ActivateCode: hide ticket input if it's implied

### LINKED ISSUE #443: UX Fix [Gray Activation Code Text]
I've received a few reports that the activation code workflow is not intuitive. Typically we've been trained to ignore gray text and the prepopulated activation code is gray. I find people asking me "where's my activation code?" because they assume the text in the field is just filler text when it's actually their activation code.

I suggest hiding the input box altogether if the activation code is prepopulated and directing the user's attention to advancing to the next step in the process.

liam-fitzgerald:
Hides the ticket and changes the copy if there's an implied point
from the URL, in order to minimise user cognitive load. We don't
skip this page entirely because we have to make network and crypto
calls and doing this upon pageload would ruin time to interactive.

cc: @kennyrowe

Fixes #443
localhost_3000_ (2)

### DIFF
--- a/src/views/Activate/ActivateCode.js
+++ b/src/views/Activate/ActivateCode.js
@@ -157,7 +157,7 @@ export default function ActivateCode() {
         </Grid.Item>
         <Grid.Item full as={P} className="mb2">
           Someone has invited you to claim your Urbit identity and join the
-          network. Enter your activation code to continue.
+          network. {!impliedTicket && 'Enter your activation code to continue.'}
         </Grid.Item>
         <BridgeForm
           validate={validate}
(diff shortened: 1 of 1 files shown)


liam-fitzgerald:
what browser @Fang- ? Do you have a running ganache? Works fine on mine(tm)

Fang-:
I'm on Firefox. Had modified `Bridge.js` to use the Ropsten network instead, so I don't think local ganache mattered (but it likely got started anyway).

liam-fitzgerald:
odd, just tried to repro w/ firefox and ropsten and worked fine. owned should never be null, so maybe check node_modules?

Issue #437, opened 2020-04-02
Feature request: Allow "invite group" field understand line breaks


rmariani:
It'd be great if a big stack of linebreak-separated emails could just be copy-pasted into the Bridge invite group field. Right now, however, it interprets such a list as one gigantic email entry with a space for each linebreak.

PR #460, opened 2020-05-08
input: allow pasting comma-separated email lists

Fang-:
For easier sending of larger batches of invites. We already split on `,` and ` `  keypresses when writing these things, so pasting those in should count as separators, too.


liam-fitzgerald:
maybe add \n to the regex and we can close #437

Fang-:
The description of that suggests it's already interpreting the newlines as spaces. But seems good practice to just mirror whatever separators we accept as input, anyway, so I added `\n` and `\t` here.

liam-fitzgerald:
looks good!

PR #476, opened 2020-06-11
copy: touch up urbit id actions copy

Fang-:
currently:
image

This page is a mess.

"Download keys"? What keys? People are often looking for networking keys, maybe this is it? Why would you need the master ticket? (Because it's not, it's the passport.)

"Networking keys required"? Required for what? It lets you "set keys", but why is that even on this page? Isn't it an urbit os thing?

Consider this more an RFC than a PR, happy to take better changes here.

### DIFF
--- a/src/views/UrbitID/Home.js
+++ b/src/views/UrbitID/Home.js
@@ -88,7 +88,7 @@ export default function UrbitIDHome() {
             · Master Key Authentication required.
           </B>
         }>
-        Download Keys
+        Re-download Passport
       </Grid.Item>
       <Grid.Divider />
       <Grid.Item
@@ -136,7 +136,7 @@ export default function UrbitIDHome() {
             full
             as={ForwardButton}
             className="f6"
-            accessory={<span className="underline pointer">Set Keys</span>}>
+            accessory={<span className="underline pointer">Set Networking Keys</span>}>
             Network Keys Required
           </Grid.Item>
           <Grid.Divider />


Fang-:
Ping @jyng. If this is deprecated by your ongoing Bridge work, maybe we can backport some of the copy?

Fang-:
ping @jyng again, see above.

jyng:
Sorry that i let this issue languish.

1) We're trying to make the neologisms around Bridge a lot clearer, so we're not using passport anymore and just calling it what it is. I think we could say "Download All Keys" here.

2) The main use cases right now is just logging into Urbit OS. A supplementary, but important, use case is security protocol and your master key is the most precious of them all. I'm curious as to why people are looking for their networking keys. That seems like a path that shouldn't be required to start using their OS (metamask users are an exception)

3) "Set Keys" should redirect you to Set Network Keys which lives inside Urbit OS. This sort of alert is a good pattern to have inside Bridge. The secondary confusion might be why Spawn Key requires Network Key to be set. I'm assuming thats logic written in Azimuth. I think confusions like these could be solved be redirecting to an FAQ as Spawning is lower priority than just getting into Urbit OS.

Fang-:
> I'm curious as to why people are looking for their networking keys. That seems like a path that shouldn't be required to start using their OS

You explicitly need your networking keys/keyfile to boot your ship.

Fang-:
This was touched up in #592.

Issue #481, opened 2020-06-26
Error Report: Cannot create property 'message' on string 'Node error: {"code":-32000,"message":"already known"}'


sphinxx0000:
I was trying to claim a planet sent to me by Tlon support team. Going to try again. Working from Chrome on Mac OS

## Error

```
TypeError: Cannot create property 'message' on string 'Node error: {"code":-32000,"message":"already known"}'
    at renderAdditionalInfo (https://bridge.urbit.org/static/js/main.b6a701ac.chunk.js:2322:385)
    at PassportTransfer (https://bridge.urbit.org/static/js/main.b6a701ac.chunk.js:2322:1914)
    at dh (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:159072:7)
    at Jh (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:159548:7)
    at mj (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:161821:86)
    at jj (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:160809:11)
    at Z (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:160666:15)
    at $i (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:160525:16)
    at https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:158251:17
    at push.exports.unstable_runWithPriority (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:162946:12)
```

In Bridge version 2.5.0.

vvisigoth:
@Fang- have you seen this before?

Fang-:
Looks like it was trying to show an error message and errored out while doing so.

"already known" suggests the transaction(s) succeeded though. We try to catch these, since they're generally a harmless side-effect of helping you pay for the transaction. I wonder why this thought it was an actual error...

> Going to try again.

If your invite no longer works, you can probably log in with the master ticket that you downloaded.

Fang-:
Closing as duplicate of #478. Thanks for filing!

Issue #480, opened 2020-06-26
Error Report: sigil.js cannot render @p of length 8. Only lengths of 1 (galaxy), 2 (star), and 4 (planet) are supported at this time.


vshvedov:
## Error

```
construct@[native code]
_construct@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:236630:26
Wrapper@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:92325:23
ConfigError@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:56227:79
sigil@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:56300:26
dh@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:159072:8
mj@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:161764:13
jj@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:160809:13
Z@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:160666:17
Z@[native code]
$i@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:160525:17
$i@[native code]
https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:158251:18
https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:162946:13
yf@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:158246:9
O@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:158236:5
fj@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:160573:24
Sd@https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:157491:7
Sd@[native code]
```

In Bridge version 2.5.0.

Fang-:
Closing as duplicate of #344.

Issue #479, opened 2020-06-23
Breadcrumb navigation crashes on residents page


conradconrad:
running Chrome Version 83.0.4103.106 (Official Build) (64-bit) accessing https://bridge.urbit.org/

from ~star/residents clicking on ~star at top of page crashes bridge. the back arrow works

```
TypeError: Cannot read property 'data' of undefined
    at https://bridge.urbit.org/static/js/main.b6a701ac.chunk.js:1332:911
    at Object.useMemo (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:159401:9)
    at useMemo (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:155616:16)
    at useRouter (https://bridge.urbit.org/static/js/main.b6a701ac.chunk.js:1332:862)
    at HistoryProvider (https://bridge.urbit.org/static/js/main.b6a701ac.chunk.js:1347:196)
    at dh (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:159072:7)
    at Jh (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:159548:7)
    at mj (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:161821:86)
    at jj (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:160809:11)
    at Z (https://bridge.urbit.org/static/js/2.a3f8391b.chunk.js:160666:15)
```

In Bridge version 2.5.0.

PR #482, opened 2020-06-27
various bridge cleanup

liam-fitzgerald:
some fixes: 
txn: always throw an Error on failure
sigil: check patp has sigil before render
residents: fix broken crumb navigation

fixes: #478 #481 #480 #344 #479 #484

### DIFF
--- a/src/components/MaybeSigil.js
+++ b/src/components/MaybeSigil.js
@@ -10,7 +10,8 @@ import AspectRatio from './AspectRatio';
 export default function MaybeSigil({ className, patp, size, ...rest }) {
   const validPatp = patp.matchWith({
     Nothing: () => null,
-    Just: p => (ob.isValidPatp(p.value) ? p.value : null),
+    Just: p =>
+      (ob.isValidPatp(p.value) && p.value.length) < 15 ? p.value : null,
   });
 
   return validPatp ? (
--- a/src/components/MaybeSigil.js
+++ b/src/components/MaybeSigil.js
@@ -11,7 +11,7 @@ export default function MaybeSigil({ className, patp, size, ...rest }) {
   const validPatp = patp.matchWith({
     Nothing: () => null,
     Just: p =>
-      (ob.isValidPatp(p.value) && p.value.length) < 15 ? p.value : null,
+      ob.isValidPatp(p.value) && p.value.length < 15 ? p.value : null,
   });
 
   return validPatp ? (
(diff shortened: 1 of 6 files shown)


vvisigoth:
Thanks!

A
—
~poldec-tonteg
http://urbit.org

src/lib/txn.js
```diff
@@ -92,7 +92,7 @@ const signTransaction = async ({
   } else if (walletType === WALLET_TYPES.TREZOR) {
     await trezorSignTransaction(stx, walletHdPath);
   } else if (walletType === WALLET_TYPES.METAMASK) {
-    return metamaskSignTransaction(stx, wallet.address);
+    return metamaskSignTransaction(utx, wallet.address);
```
Fang-, line 95: Can you explain this change? This cascades down into the changes in `metamaskSignTransaction`, which seem to give us less guarantees about the format of the data we pass to Metamask. But maybe I'm wrong and/or Metamask doesn't care?

src/components/MaybeSigil.js
```diff
@@ -10,7 +10,8 @@ import AspectRatio from './AspectRatio';
 export default function MaybeSigil({ className, patp, size, ...rest }) {
   const validPatp = patp.matchWith({
     Nothing: () => null,
-    Just: p => (ob.isValidPatp(p.value) ? p.value : null),
+    Just: p =>
+      (ob.isValidPatp(p.value) && p.value.length) < 15 ? p.value : null,
```
Fang-, line 14: Doesn't this change the invalid patp case?

```
> false < 15
true
```

src/lib/useRouter.js
```diff
@@ -89,7 +89,8 @@ export default function useRouter({
     routes,
   ]);
   const data = useMemo(() => {
-    return last(routes).data || NULL_DATA;
+    return get(last(routes), 'data', {});
+    // return last(routes).data || NULL_DATA;
```
Fang-, line 93: maybe:

```suggestion
    return get(last(routes), 'data', NULL_DATA);
```

src/views/Residents.js
```diff
@@ -158,6 +158,7 @@ export default function Residents() {
   const point = need.point(pointCursor);
   const { getResidents, syncResidents } = usePointCache();
 
+  const goBack = useCallback(() => pop(), [pop]);
```
Fang-, line 161: Does the indirection here actually change anything? Wild if so.

Fang-:
@liam-fitzgerald would love to get these fixes out. Can you respond to my questions/comments?

src/views/Residents.js
liam-fitzgerald: it does, shockingly

src/lib/txn.js
liam-fitzgerald: by not putting it in a Tx object we preserve some properties which allows us to use web3 instead of manually sending the RPC request. This means that the error being thrown is consistently an `Error` object and not some other weird jankery.

PR #488, opened 2020-07-13
Merge #433 into master

Fang-:
Didn't realize that was still targeting lf/auth-token, so re-merging that into master.


Fang-:
@liam-fitzgerald pretty sure I got the conflict resolution on this merge right. lmk if you see anything that's off.

PR #493, opened 2020-07-16
migrations: fix broken azimuth imports

pkova:
Developing locally with `npm run pilot` is broken because of these imports since the truffle migrations never run.

### DIFF
--- a/migrations/1_migration.js
+++ b/migrations/1_migration.js
@@ -1,17 +1,11 @@
-var Azimuth = artifacts.require('azimuth-solidity/contracts/Azimuth.sol');
-var Polls = artifacts.require('azimuth-solidity/contracts/Polls.sol');
-var Claims = artifacts.require('azimuth-solidity/contracts/Claims.sol');
-var Censures = artifacts.require('azimuth-solidity/contracts/Censures.sol');
-var Ecliptic = artifacts.require('azimuth-solidity/contracts/Ecliptic.sol');
-var DelegatedSending = artifacts.require(
-  'azimuth-solidity/contracts/DelegatedSending.sol'
-);
-var LinearStarRelease = artifacts.require(
-  'azimuth-solidity/contracts/LinearStarRelease.sol'
-);
-var CSR = artifacts.require(
-  'azimuth-solidity/contracts/ConditionalStarRelease.sol'
-);
+var Azimuth = artifacts.require('azimuth-solidity/Azimuth');
+var Polls = artifacts.require('azimuth-solidity/Polls');
+var Claims = artifacts.require('azimuth-solidity/Claims');
+var Censures = artifacts.require('azimuth-solidity/Censures');
+var Ecliptic = artifacts.require('azimuth-solidity/Ecliptic');
+var DelegatedSending = artifacts.require('azimuth-solidity/DelegatedSending');
+var LSR = artifacts.require('azimuth-solidity/LinearStarRelease');
+var CSR = artifacts.require('azimuth-solidity/ConditionalStarRelease');
 
 const user1 = '0xD53208cf45fC9bd7938B200BFf8814A26146688f';
 const rateUnit = 50;
@@ -91,7 +85,7 @@ module.exports = async function(deployer) {
     escapeHatchDate
   );
 
-  const lsr = await deployer.deploy(LinearStarRelease, azimuth.address);
+  const lsr = await deployer.deploy(LSR, azimuth.address);
 
   const own = await ecliptic.owner();
   switch (process.env.WITH_TEST_STATE) {


pkova:
@Fang-

pkova:
Also, I originally thought these changes required something like my stuff over at https://github.com/urbit/azimuth/pull/30, but I was wrong. The contents of this PR work with our without the azimuth stuff.

Fang-:
After wiping node_modules, I'm getting this error:

```
./node_modules/azimuth-js/contracts.js
Module not found: Can't resolve 'azimuth-solidity/build/contracts/Azimuth.json' in '/Users/fang/repos/urbit/bridge/node_modules/azimuth-js'
```

Which looks like azimuth-js suffers from the same issues... which makes me wonder why these things are no longer nested in the `/contracts` folder in the first place.

@jtobin I recall you wrestling with this some time ago. Do you know what's up with that/when it changed? Maybe we want to change that back in azimuth-solidity, rather than updating all its dependents.

pkova:
```
git clone https://github.com/pkova/bridge.git
npm install
npm run pilot
```
works for me.

jtobin:
> Which looks like azimuth-js suffers from the same issues... which makes me wonder why these things are no longer nested in the /contracts folder in the first place.

Huh, I actually can't repro on azimuth-js (after wiping `node_modules` and `package-lock.json`, then `npm install && npm test`).  Works fine.

> @jtobin I recall you wrestling with this some time ago. Do you know what's up with that/when it changed? Maybe we want to change that back in azimuth-solidity, rather than updating all its dependents.

I wouldn't have changed the location of the built contracts in azimuth-solidity (I would rev the major version for that).  Is this a truffle thing maybe?  Bridge is using `truffle-5.1.16` as a dev dependency; did the way `artifacts.require` handles paths change?

jtobin:
Just to summarize the present mechanics here as I understand them:

Pulling in azimuth-solidity v1.2.x as a dependency and installing with `npm install` will compile the contracts with truffle v4.1.11 on installation, leaving the artifacts accessible in `node_modules/azimuth-solidity/build/contracts`.  Those can then just be `require`'d outright, as is done in azimuth-js via e.g.:

```
const eclipticAbi = require('azimuth-solidity/build/contracts/Ecliptic.json').abi;
```

Bridge is using a somewhat different setup though; truffle v5.1.16 is a dev dependency here, and I believe the migration script invokes it to compile the contracts itself, then retrieving *those* artifacts via `artifacts.require`.  To be clear, these are different from the artifacts that are already present in via the azimuth-solidity installation, and I'm actually wondering if this migration operation *overwrites* them.

I'm actually surprised the changes here work, since the contract sources should still be located in `azimuth-solidity/contracts` regardless, and that's what it looks like `artifacts.require` wants to be pointed at.

jtobin:
And hmm, it looks like `artifacts.require` is a little weird and does some of its own namespacing.  Not sure what is actually happening behind the scenes here:

> At the beginning of the migration, we tell Truffle which contracts we'd like to interact with via the artifacts.require() method. This method is similar to Node's require, but in our case it specifically returns a contract abstraction that we can use within the rest of our deployment script. The name specified should match the name of the contract definition within that source file. **Do not pass the name of the source file, as files can contain more than one contract.**

pkova:
It's just hardcoded: https://github.com/trufflesuite/truffle/blob/692eac39c5af9c3cb2709eb1e96147cbbbd45b99/packages/resolver/lib/sources/npm.ts#L30

jtobin:
Right ok, so truffle just inserts `build/contracts` on its own (and doesn't do any of its own compiling, i.e. does not overwrite v4.1.16's artifacts).  It also ignores the `.sol`.

I guess the idiomatic Truffle-approved way to do this should really be:

```
var Azimuth = artifacts.require('azimuth-solidity/Azimuth');
var Polls = artifacts.require('azimuth-solidity/Polls');
var Claims = artifacts.require('azimuth-solidity/Claims');
var Censures = artifacts.require('azimuth-solidity/Censures');
var Ecliptic = artifacts.require('azimuth-solidity/Ecliptic');
var DelegatedSending = artifacts.require('azimuth-solidity/DelegatedSending');
var LinearStarRelease = artifacts.require('azimuth-solidity/LinearStarRelease');
var CSR = artifacts.require('azimuth-solidity/ConditionalStarRelease');
```

I can confirm this works fine for me, @Fang-.

jtobin:
I've been trying to use the GitHub web UI to suggest changes but can't figure it out, so here's a `git-apply`-able patch:

```
diff --git a/migrations/1_migration.js b/migrations/1_migration.js
index 1f25a58..9c27e00 100644
--- a/migrations/1_migration.js
+++ b/migrations/1_migration.js
@@ -1,17 +1,11 @@
-var Azimuth = artifacts.require('azimuth-solidity/contracts/Azimuth.sol');
-var Polls = artifacts.require('azimuth-solidity/contracts/Polls.sol');
-var Claims = artifacts.require('azimuth-solidity/contracts/Claims.sol');
-var Censures = artifacts.require('azimuth-solidity/contracts/Censures.sol');
-var Ecliptic = artifacts.require('azimuth-solidity/contracts/Ecliptic.sol');
-var DelegatedSending = artifacts.require(
-  'azimuth-solidity/contracts/DelegatedSending.sol'
-);
-var LinearStarRelease = artifacts.require(
-  'azimuth-solidity/contracts/LinearStarRelease.sol'
-);
-var CSR = artifacts.require(
-  'azimuth-solidity/contracts/ConditionalStarRelease.sol'
-);
+var Azimuth = artifacts.require('azimuth-solidity/Azimuth');
+var Polls = artifacts.require('azimuth-solidity/Polls');
+var Claims = artifacts.require('azimuth-solidity/Claims');
+var Censures = artifacts.require('azimuth-solidity/Censures');
+var Ecliptic = artifacts.require('azimuth-solidity/Ecliptic');
+var DelegatedSending = artifacts.require('azimuth-solidity/DelegatedSending');
+var LSR = artifacts.require('azimuth-solidity/LinearStarRelease');
+var CSR = artifacts.require('azimuth-solidity/ConditionalStarRelease');
 
 const user1 = '0xD53208cf45fC9bd7938B200BFf8814A26146688f';
 const rateUnit = 50;
@@ -91,7 +85,7 @@ module.exports = async function(deployer) {
     escapeHatchDate
   );
 
-  const lsr = await deployer.deploy(LinearStarRelease, azimuth.address);
+  const lsr = await deployer.deploy(LSR, azimuth.address);
 
   const own = await ecliptic.owner();
   switch (process.env.WITH_TEST_STATE) {
```

pkova:
Fixed!

pkova:
Merge this my dudes.

Fang-:
I can finally merge now that my boss gave me permission! (^:

PR #495, opened 2020-07-17
tx: fill out the rawTx before sending the rawTx

Fang-:
I think d20b79d/#482 introduced this. Apparently we didn't test the non-metamask case here.


Fang-:
Gonna go ahead and self-merge this as the fix seems dumb & straightforward, and the impact quite non-trivial.

PR #498, opened 2020-07-17
tx: catch more "tx known" errors

### LINKED ISSUE #497: Error on Invite Claiming, but Master Ticket still works
I went to claim an invite and got an error on Step 3 of invite claiming.

```
Error: Node error: {"code":-32000,"message":"already known"}
    at Function.validate (web3-providers.umd.js:127)
    at HttpProvider._callee$ (web3-providers.umd.js:810)
    at tryCatch (runtime.js:45)
    at Generator.invoke [as _invoke] (runtime.js:274)
    at Generator.prototype.<computed> [as next] (runtime.js:97)
    at asyncGeneratorStep (asyncToGenerator.js:3)
    at _next (asyncToGenerator.js:25)
```

I was, however, able to log in with the downloaded collateral and everything works.

Screen Shot 2020-07-17 at 11 34 01 AM

Fang-:
Fixes #497.

Whoever made it so that error codes are always the same for every error, but messages vary per node implementation, needs to be trialed.

### DIFF
--- a/src/lib/txn.js
+++ b/src/lib/txn.js
@@ -120,7 +120,9 @@ const sendSignedTransaction = (web3, stx, doubtNonceError) => {
         // that it's because the tank already submitted our transaction.
         // we just wait for first confirmation here.
         const message = typeof err === 'string' ? err : err.message || '';
-        const isKnownError = message.includes('known transaction: ');
+        const isKnownError =
+          message.includes('known transaction') ||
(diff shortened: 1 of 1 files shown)


Fang-:
> do we know which nodes have these odd codes?

Hmm, considering we only talk to Infura, maybe things just changed out from under us. Equally bad, I suppose.

PR #505, opened 2020-07-21
keyfiles: make all keyfiles deterministic

liam-fitzgerald:
uses the authToken to create a deterministic keyfile that will be valid as long as the (point, revision, authToken) triple remains the same. This allows for redownloading of keyfiles that were set on a non master ticket wallet, and also viewing of the login code for non-master ticket wallets. 

cc: @matildepark might address some of your qualms re: non master ticket wallets being second class citizens in bridge.

### DIFF
--- a/migrations/1_migration.js
+++ b/migrations/1_migration.js
@@ -119,6 +119,7 @@ module.exports = async function(deployer) {
         await lsr.deposit(user1, offset);
         await csr.deposit(user1, offset + 1);
       }
+      break;
     case 'INVITES':
       await ecliptic.createGalaxy(0, own);
       await ecliptic.configureKeys(0, '0x123', '0x456', 1, false);
@@ -130,6 +131,7 @@ module.exports = async function(deployer) {
       await ecliptic.spawn(131328, own);
       await ecliptic.spawn(512, own);
       await sending.setPoolSize(256, 65792, 1000);
+      break;
     case 'RESIDENTS':
       await ecliptic.createGalaxy(0, own);
       await ecliptic.configureKeys(0, '0x123', '0x456', 1, false);
@@ -141,6 +143,7 @@ module.exports = async function(deployer) {
         await ecliptic.spawn(offset + 256, own);
         await ecliptic.transferPoint(offset + 256, user1, false);
       }
+      break;
 
     default:
       return;
--- a/src/lib/keys.js
+++ b/src/lib/keys.js
@@ -6,6 +6,7 @@ import * as serial from '../nockjs/serial';
 import * as kg from 'urbit-key-generation';
 
 import { eqAddr, addHexPrefix } from './wallet';
+import { shas } from './networkCode';
 
 // the curve param for the network keys
 export const NETWORK_KEY_CURVE_PARAMETER = '42';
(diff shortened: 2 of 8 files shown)


liam-fitzgerald:
(obviously if you wish to use a custom network seed you still may)

matildepark:
Damn, where’d all my qualms go

galenwp:
king

migrations/1_migration.js
```diff
@@ -119,6 +119,7 @@ module.exports = async function(deployer) {
         await lsr.deposit(user1, offset);
         await csr.deposit(user1, offset + 1);
       }
+      break;
```
Fang-, line 122: \>sneaking unrelated changes into the commit  
stop doing this lol

(But it's a good change, so I'll once again allow it for now.)

src/lib/keys.js
```diff
@@ -141,27 +142,33 @@ const deriveNetworkSeedFromMnemonic = async (
   return Just(await kg.deriveNetworkSeed(mnemonic, passphrase, revision - 1));
 };
 
+/**
+ * @param {string} authToken
+ * @param {number} number
```
Fang-, line 147:
```suggestion
 * @param {string} authToken
 * @param {number} revision
 * @param {number} point
```

src/lib/keys.js
```diff
@@ -141,27 +142,33 @@ const deriveNetworkSeedFromMnemonic = async (
...
+  point,
+  authToken,
   revision,
 }) => {
   if (Just.hasInstance(urbitWallet)) {
     return await deriveNetworkSeedFromUrbitWallet(urbitWallet.value, revision);
   }
 
-  if (Just.hasInstance(wallet) && Just.hasInstance(authMnemonic)) {
-    return await deriveNetworkSeedFromManagementMnemonic(
```
Fang-, line 159: If we're logged in with the management mnemonic, we can still derive from that as per the wallet structure just fine. Is there a reason we shouldn't?

src/lib/keys.js
```diff
@@ -141,27 +142,33 @@ const deriveNetworkSeedFromMnemonic = async (
...
-      wallet.value,
-      authMnemonic.value,
-      details,
-      revision
-    );
+  if (Just.hasInstance(authToken)) {
+    return deriveNetworkSeedFromAuthToken(authToken.value, point, revision);
```
Fang-, line 171: The arguments for the function say `revision, point` though, not the other way around.

Should probably make ordering of these consistent with `attemptNetworkSeedDerivation` anyway.

Fang-:
Hmm. I'm trying to repro issues around downloading keyfiles, stale keyfiles, etc. and working off this PR because that seems more useful. I think this is exacerbating* a problem that was already there: we generate keys using the _current_ revision, not the _next_, leading to setting the same keys, leading to revision not increasing, leading to no-op transactions.

This is weird, because we certainly do use `revision + 1` for deriving, as per the following:

https://github.com/urbit/bridge/blob/f6b288d51ee584f6c5086ae9db2ae9dac0ee9a02/src/views/UrbitOS/NetworkingKeys.js#L62-L67

(* note how manual (which includes random) seeds do not care about the revision number. With this PR, I believe most (if not all?) login methods use derived seeds instead, causing these problems to be much more common.)

I spent a fair amount of time today trying to figure out how, if the `buildNetworkSeed` logic there uses `revision + 1`, we end up generating a transaction with keys for `revision` instead. And... I'm not sure yet.

`useKeyfileGenerator` sees a lot of calls, putting logs in its `generate` callback sees it run a bunch of times over the course of a typical "reset networking keys" flow. Just theory rn, but perhaps something there is overwriting some in-store value, leading to the transaction being regenerated with those other values?

I'll need to spend more time hunting this down. @liam-fitzgerald if you have any thoughts/suggestions on this, I'd love to hear them.

Fang-:
Oh, well, this is thoroughly confusing.

https://github.com/urbit/bridge/blob/d93d23e02405017f976ad48559af7501e98d6126/src/lib/keys.js#L133-L141

Guess we've always done "derive current keys at revision n-1" and "keys that will get revision n at n-1" etc. That's... not great? Why did we do that?

We can course-correct for that by doing "keys that have revision n are derived at n", at the cost of retrying derivation at n-1 if n doesn't get us keys that match chain. Seems tempting... But also hesitant to touch any of this for fear of just breaking it in a different way.

Fang-:
Alright, cc9183b (which I so boldly pushed to your PR) resolves my last remaining issue with this.

Changing to the auth mnemonic even in the "logged in with management mnemonic" case broke compatibility with previously generated keyfiles, and moved away from the uhdw scheme where we could still easily stick to it.  
The commit I just added reverts us to the previous behavior, except that failure there (ie, "mnemonic is not current management proxy") falls back to the auth token.

It smells a little bit fishy to me that we were not doing uhdw-style deterministic networking keys for _all_ mnemonic logins in the first place, but I don't think it's the biggest deal.

I'll go ahead and self-merge because I want to get all of these fixes out before the weekend. @liam-fitzgerald please do still take a look at this, just in case I'm missing anything weird.

Issue #496, opened 2020-07-17
Sigil margins are inconsistent & incorrect


Fang-:
Just observed this in v2.5.3. Did some npm wrangling prior to this, not sure what I kicked, but some sigils get rendered kinda poorly. Stars get weird margins in the passports and mini-passports, and sigils in input fields don't have any margin/padding at all.

<img width="491" alt="image" src="https://user-images.githubusercontent.com/3829764/87812613-c1ad5800-c860-11ea-98ae-283724789b5b.png">

<img width="248" alt="image" src="https://user-images.githubusercontent.com/3829764/87812510-94f94080-c860-11ea-9721-5854f0a6e60a.png">

Fang-:
I solved the "no margins" problem in 059ed84, but still don't know what's up with the weird cutoff situation. @g-a-v-i-n any idea where the problem with this could be? The top level `<g>` in the svg has weird `right` and `bottom`(?) values, setting those to 0 fixes this:

![sigil-futz](https://user-images.githubusercontent.com/3829764/87818746-65036a80-c86b-11ea-8b0b-e289ad7000bf.gif)

JohnELester:
I just noticed this with ~dasfeb.

![1](https://user-images.githubusercontent.com/25461543/87829722-94a37a00-c84d-11ea-8e98-e244195d2a59.png)

![2](https://user-images.githubusercontent.com/25461543/87829729-979e6a80-c84d-11ea-8f29-60637ca5a4d5.png)

g-a-v-i-n:
looks like an issue with sigil-js, will take a look

g-a-v-i-n:
you likely need to use the following in the config
```js
            width: size,
            height: size / 2,
            margin: false,
```

Fang-:
@g-a-v-i-n that doesn't change the visual result here, and the values I noted are still there...

g-a-v-i-n:
Would you kindly point me to a branch?

Fang-:
master is fine, not much development is happening outside of that.

g-a-v-i-n:
What do you mean by `right` and `bottom` values? Are you referring to CSS, style properties or something else? I'm having a hard time getting to the bottom of this because I don't know the Bridge codebase and sigils render as expected on the sigil-js test page.

Fang-:
In the gif, you can see the outer `<g>` has a transform matrix of `0.5,0,0,0.5,-32,16`, which I think is "width, height, top, left, right, bottom"?

[...]

PR #506, opened 2020-07-21
invites: handle error cases slightly better

### LINKED ISSUE #455: Shouldn't pass objects as errors to final-form
## Context

I had just verified transaction for sending an email invitation for a planet 

I was waiting for the "Sending Invitation" action to complete, and this problem spontaneously occurred.  It happened several times.  I am unable to send an email invitation.

## Error

`Objects are not valid as a React child (found: object with keys {ticket, signedTx, rawTx}). If you meant to render a collection of children, use an array instead.`

```
Error: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7Bticket%2C%20signedTx%2C%20rawTx%7D&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings. 
    at mg (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:158621:38)
    at z (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:158715:7)
    at rb (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:158781:13)
    at https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:158898:23
    at S (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:159470:26)
    at mj (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:161811:306)
    at jj (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:160777:11)
    at Z (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:160634:15)
    at $i (https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:160493:16)
    at https://bridge.urbit.org/static/js/2.9ed8dd63.chunk.js:158219:17
```

In Bridge version 2.3.5.

Fang-:
Also sneaks in a sigil-js upgrade, as partial fix for #496.

This makes invite error handling slightly better, by making it actually render errors, instead of crashing on you. So this fixes #455. But it's still not great...

@liam-fitzgerald consider the following, imagining that only some of the invite transactions succeeded, while some fail:

https://github.com/urbit/bridge/blob/82af1413565f05839b2551cf5451c461cce4c5e5/src/lib/useInviter.js#L186-L197

https://github.com/urbit/bridge/blob/82af1413565f05839b2551cf5451c461cce4c5e5/src/views/Invite/Email.js#L85-L91

We error out without sending emails about the invite transactions that succeeded. This is obviously Not Good (though slightly better now that the user can recover the invite codes).

I tried fixing this, but when trying to communicate intermediate and partial failure states to the user, but found myself stuck in react-final-form hell: I can only communicate successes or failures, nothing in between.

So, RFC: should I just add some additional state for these situations and set those as invite logic proceeds, similar to `setStatus`? Or is there some way to do this that's more friendly to current architecture? ("It's fscked, just do whatever" is a fair answer, too.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1815,6 +1815,16 @@
         "loader-utils": "^1.1.0"
       }
     },
+    "@tlon/sigil-js": {
+      "version": "1.4.1",
+      "resolved": "https://registry.npmjs.org/@tlon/sigil-js/-/sigil-js-1.4.1.tgz",
+      "integrity": "sha512-+ajy1lqChGzpFzYE03IhSBmn+IugcgjSVhMurWN6CUb50Wc4z7epPpNk2xFvyfY5BilQlgtxDvMpIcNS54Iw2Q==",
+      "requires": {
+        "invariant": "^2.2.4",
+        "svgson": "^4.0.0",
+        "transformation-matrix": "2.1.1"
+      }
+    },
     "@types/babel__core": {
       "version": "7.1.9",
       "resolved": "https://registry.npmjs.org/@types/babel__core/-/babel__core-7.1.9.tgz",
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "@hot-loader/react-dom": "^16.8.6",
     "@ledgerhq/hw-app-eth": "^5.5.0",
     "@ledgerhq/hw-transport-u2f": "^5.5.0",
+    "@tlon/sigil-js": "^1.4.1",
     "@welldone-software/why-did-you-render": "^3.2.3",
     "async-retry": "^1.2.3",
     "azimuth-js": "^0.19.0",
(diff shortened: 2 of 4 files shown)


liam-fitzgerald:
could probably just make the `FORM_ERROR` prop a more complex object for and include the recovery in that.

Fang-:
> could probably just make the FORM_ERROR prop a more complex object

How do I produce that (displaying it in the UI) without `return`ing out of the `onSubmit` call though? I'm under the impression you don't want to return from that until you're actually completely done with form processing...

(Looking for a review on what's already here regardless of this, to be clear. Want to get these and other recent fixes out sooner rather than later.)

Fang-:
Gave this another look over, gonna go ahead and self-merge. This should at least make things _feel_ a little better when they fail. I'll go and make an issue for the remaining error handling work I mentioned above.

PR #517, opened 2020-08-13
libs: move out into separate bridge-libs repository

Fang-:
Crude initial detaching of (not quite fully) generic Ethereum and Azimuth library functions.

Intent here is to have [a repository](https://github.com/urbit/bridge-libs) for general-use Bridge-adjacent functions, like all the things we do around keyfile generation, invite handling, reticketing flows etc.

After some fiddling (notice minor adjustments to some function interfaces) this seems to mostly work. Didn't manage to make it through the invite flow yet, but not sure if that's directly related. Will look into this later.

This isn't the most tidy situation right now. Network types don't bridge (heh) the gap very elegantly, lib structure/design is basically non-existent, etc. There's also weird situations around dependencies and their versions. Proper solution there is to just upgrade everything to latest, but there's breaking changes involved in that...


Fang-:
> August 2020

hah, closing. cc @nerveharp for perusal

PR #518, opened 2020-08-13
login: improve shard form input handling

### LINKED ISSUE #516: Unable to access Bridge using 2/3 shard multisig
From a user: I have the first shard and the third shard. When I try to enter bridge by putting the first shard into the first form field and the third shard into the third form field, the UI says "This field is required" on the second form field. When I try to put the third shard into the second form field instead of into the third field, it says the third field is required. This amounts to me being unable to access Bridge unless I have all three of my shards, even though I'm using 2/3 multisig. This is unacceptable. Please help!

Fang-:
- Consider empty shard fields as valid.
- If too many shard fields are empty, present an error to the user, asking them to provide at least two shards.
- Coerce shards to be either full strings, or `undefined`. Keygen currently doesn't handle the empty string case well.

Fixes #516.

### DIFF
--- a/src/form/validators.js
+++ b/src/form/validators.js
@@ -3,6 +3,7 @@ import { some } from 'lodash';
 import {
   validateNotEmpty,
   validatePatq,
+  validateShard,
   validateMnemonic,
   validateHdPath,
   validatePoint,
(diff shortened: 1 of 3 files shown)


vvisigoth:
king
—
~poldec-tonteg
http://urbit.org

vvisigoth:
@tacryt-socryp tell your "user"

tacryt-socryp:
Thanks, will do.

PR #520, opened 2020-08-17
ui: allow setting gas price >100 gwei

Fang-:
By bumping the max up to 400.

This is a pretty arbitrary number. I've seen ethgasstation.info mention prices over 300 gwei for fast transactions, but haven't even seen it touch 400. The etherscan.io stats for average gas price per day don't go above 300.

tldr this should be plenty unless things get really wild.


liam-fitzgerald (approved):
based and gas-pilled

PR #521, opened 2020-08-17
constants: (nearly) limitless suggested gas price

Fang-:
I imagine we still want some value here, as a sanity check, so I'm making it match #520.

Since this now goes way past what the gas tank is willing to fund, we'll probably want to update the copy of the "pls provide your own funds" message, as has been discussed over email, prior to merging this.

Notably, _this impacts the invite flow_, in that it's much more likely to ask users to bring their own funds, in the middle of onboarding. Might make for a rough experience.

### DIFF
--- a/src/components/NeedFundsNotice.js
+++ b/src/components/NeedFundsNotice.js
@@ -17,7 +17,9 @@ export default function NeedFundsNotice({
       <Highlighted warning>
         Your ownership address <CopiableAddress>{address}</CopiableAddress>{' '}
         needs at least {safeFromWei(minBalance)} ETH and currently has{' '}
-        {safeFromWei(balance)} ETH. The transaction will automatically resume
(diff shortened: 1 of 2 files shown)


liam-fitzgerald (approved):
Step on the gas!

Fang-:
@vvisigoth I think it's wise to consider merge of this blocked on copy changes making it in, especially considering this hits the invite flow. The email thread on this hasn't quite concluded yet though. I can make a call on the minimum viable copy change if you want, or just wait it out.

Fang-:
Updated the copy as per email thread. @vvisigoth please take a look.

image

It seems a bit on the wordy side to me, but it's probably better, at least?

I tried displaying the extra copy conditionally, based on the gas price, but that would need to be threaded through lots of places... so I opted for the less invasive change for now, displaying this exact copy always.

PR #522, opened 2020-08-25
Bridge "flows"

Fang-:
Introduces "flows" that can be selected by writing to `window.flow`.
Supports passing in success and failure callbacks.

Implements such a flow for sending an invite containing a specific
planet to send.

As per the commit body:

> Borrows lots of logic from useInviter. Should be refactored to share it.
> Might be worth trying to refactor JustInvite and PassportTransfer to
> share their "render progress screen" logic.

I played around with this for a bit, but ultimately didn't get anywhere...  
Deduplicating the transaction-related logic between `JustInvite` and `useInviter` doesn't get you all that much, and leads to a weird pattern where we do unpacking of all these `Just` wrappers in one place, only to use them in another.

The thing we probably want to do here is generalize the "transaction(s) construction & sending with progress indicator" logic that pops up in `reticket`, `useInviter` and `JustInvite`, along with the rendering logic for those in `PassportTransfer` and `JustInvite` (and `Invite`?). I'm... not sure what the right approach for that is though, and especially the latter half seems heavily influenced by design work on this.  
Please let me know if you have thoughts!

For testing this, consider applying f7ab291.

Drafting because even if we're fine not refactoring for now, this still needs frontend work. (cc @yosoyubik?)

image

Logging in immediately starts working, only rendering a tiny progress bar & message rn. If we want a "view and confirm" screen, probably better to add more detail to the login page?

image

(Also happy to take bikeshedding comments wrt the naming of this category of thing. No idea if there's a word commonly used for this.)


liam-fitzgerald:
This is just a comfy way to batch transactions, no?

Fang-:
Not really. The use case for which this was developed is "have landscape generate an invite (and add the planet to some group), then use bridge to send it". So this isn't about batching per se. It's about allowing outside openers to tell Bridge what specific thing they want the user to do within this Bridge session. The "single transaction" case seems more common, and is what the JustInvite flow does.

Fang-:
Worth noting that the current approach to this (using the `window` object) doesn't actually work in cross-domain situations (thanks @pkova). We can resort to POSTing flow data, or similar approaches, but I don't think we can have Bridge talk back to the opener easily.

liam-fitzgerald (approved):
Seems sane for now. We really should acquire a real router at some point though

PR #523, opened 2020-09-08
invite: send leftover eth back to the gas tank

Fang-:
At the end of reticketing, if we used the gas tank, and used less gas
than what we got funding for, send the remaining ETH back to the gas
tank for redistribution.

This will help us lower the costs of funding these a bit. We get back the overhead without slowing down the flow for the user. [For example.](https://ropsten.etherscan.io/address/0x68dc899c009a13218de51e258e6d50331b17be91)

There's an argument to be made that this should be built into /lib/tank instead of done ad-hoc, but that change seems much more involved. The bulk of our overhead comes from the invite claiming flow anyway, this change is worth much less for individual transactions.

### DIFF
--- a/src/lib/reticket.js
+++ b/src/lib/reticket.js
@@ -203,10 +203,16 @@ export async function reticketPointBetweenWallets({
 
   progress(TRANSACTION_PROGRESS.TRANSFERRING);
 
-  await sendAndAwaitAllSerial(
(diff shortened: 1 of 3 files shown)


Fang-:
ping @liam-fitzgerald

liam-fitzgerald (approved):
LGTM!

PR #525, opened 2020-09-16
transaction: allow transfer without reset

Fang-:
By checking an "I transferred this to myself" checkbox.

Currently, accepting a transfer _always_ resets the point, clearing out all proxy addresses, networking keys, and breaching the ship. When you're just transferring from yourself to another address you control though, you generally want to keep all those things the same, and definitely not need to deal with a breach of your ship. This allows you to, well, not reset on-transfer.

image

We may or may not want to tweak that copy. cc @jyng

### DIFF
--- a/src/views/AcceptTransfer.js
+++ b/src/views/AcceptTransfer.js
@@ -98,7 +98,7 @@ export default function AcceptTransfer() {
               full
               as={CheckboxInput}
               name="noReset"
-              label="I transferred this to myself"
+              label="Retain proxies and key configuration, in case of transferring to self"
               disabled={inputsLocked}
             />
           )}
--- a/src/views/AcceptTransfer.js
+++ b/src/views/AcceptTransfer.js
@@ -1,6 +1,6 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useMemo } from 'react';
 import cn from 'classnames';
-import { Grid, Text } from 'indigo-react';
+import { Grid, Text, CheckboxInput } from 'indigo-react';
 import * as azimuth from 'azimuth-js';
 
 import { useNetwork } from 'store/network';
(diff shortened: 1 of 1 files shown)


liam-fitzgerald (approved):
LGTM!

jyng:
Hm I wonder if this can be automated. Could Bridge look to see if the `From` and `To` address are the same and not reset the point?

jyng:
Oh nvm, you're saying its a different address that you own. Sorry.

I do think that copy should be tweaked. There might be a UI change, but unlikely. Let's not merge this until I hand you the change.

Fang-:
Hah yes, that's right. Alright, I'll await your review. (:

galenwp:
"Check this box to not reset the proxies, in the case you're transferring to yourself."

Fang-:
ping @jyng

syzygyzer:
ping @jyng

Fang-:
This is actually ping @Fang-, since @jyng already gave me an answer out of band. I'll try and find some time to make the discussed changes here tomorrow or early next week.

syzygyzer:
Alright thanks - I was going through the open docs PRs and https://github.com/urbit/docs/pull/948 is waiting on this one, which was why I pinged.

vvisigoth:
I think that we should just do @Fang- original suggestion with @galenwp copy suggestion. @jyng idea of putting this option under "Advanced" is confusing because 1) the option only exists on this tx and 2) it's good to remind users that this is an option. Merge and deploy @Fang-

Fang-:
image

matildepark:
I find this really confusing to look at because I don't know why I would want to retain proxies and key configuration or why that would matter when transferring to self. But, unsolicited, and it's out there.

syzygyzer:
I wanted this feature when I was transferring my planet from one wallet to another and was required to breach when doing so.

PR #528, opened 2020-10-09
txs: consider transaction broadcasted instantly

Fang-:
This removes the annoying 5-30 second wait before the state change on the
"send transaction" button, now instead making the button unclickable right away.

cc @vvisigoth


Fang-:
ping @liam-fitzgerald

liam-fitzgerald (approved):
LGTM!

PR #533, opened 2020-11-19
Add claims and altid support

xiphiness:
This PR adds support for alt-id and arbitrary claims.

This is associated with the broader antechamber initiative, described [here](https://gist.github.com/jalehman/1a3142dc30cc730159ab2daf9bcb726e)

The core functionality is all finished in this PR. The explainer components are not quite there. There is no solution to notifying a point that another point has made an alt-id claim to them without some data availability that comes from EVM event reduction, as the Claims contract does not support such arbitrary querying.

### DIFF
--- a/src/components/Passport.js
+++ b/src/components/Passport.js
@@ -21,7 +21,7 @@ function makeSigil(size, patp, colors) {
     full: true,
     patp,
     colors,
-    size
+    size,
   };
 
   return sigil(config);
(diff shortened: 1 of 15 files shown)


Fang-:
Noting for the record that we'll probably hold off on merging this until developer friendliness improvements to the Claims contract have been deployed.

> There is no solution to notifying a point that another point has made an alt-id claim to them

I'd say we do not want a solution to that in the first place, considering the use case here is claiming to be yourself, and so in legit cases, you always knowing which claims to make.

PR #534, opened 2020-11-24
bitcoin: implement bitcoin view

pkova:
As per the the bounty at https://grants.urbit.org/bounties/794126751-bitcoin-key-derivation

Depends on urbit/urbit-key-generation#92.


pkova:
Checking with a real bitcoin dude:

@timlucmiptev BIP-174 PSBT:s are the right tool for the job regarding unsigned transactions, right?

src/views/Bitcoin.js
```diff
@@ -0,0 +1,103 @@
...
+  );
+};
+
+export default function Bitcoin() {
+  const { pop } = useLocalRouter();
+  const { urbitWallet } = useWallet();
+
+  const [signedTx, setSignedTx] = useState();
+
+  const privKey = urbitWallet.unsafeGet().bitcoin.keys.private;
```
pkova, line 46: If we don't have urbitWallet here, just explode. It's checked in `Point.js`, wanted to give these keys as props to my component but didn't figure out how to do it with this routing setup.

pkova:
Out of band todos for this one after discussing with @jalehman and @timlucmiptev:

1. Exposition for the bitcoin screen.
2. Remove unnecessary `@ux` dojo formatting.
3. Proper xPub formatting.
4. Sync how bitcoinjs-lib handles PSBT:s with @timlucmiptev.

galenwp:
very cool

Fang-:
Might be interesting to look at #522, would be nice if landscape and other clients could open directly into the relevant Bitcoin flow in Bridge, instead of requiring manual user input beyond their login details.

jalehman:
> Might be interesting to look at #522, would be nice if landscape and other clients could open directly into the relevant Bitcoin flow in Bridge, instead of requiring manual user input beyond their login details.

Great point @Fang- — this is an ideal use-case.

pkova:
Fixed all the todos listed above, the exposition text might need some work though. This is now more easily runnable since urbit/urbit-key-generation#92, but generating suitable PSBT:s from the xPub is a little bit of a pain. I used https://github.com/Coldcard/psbt_faker.

Gonna look at the stuff @Fang- posted next.

winstongary272:
When it comes to trading forex,binary options and bitcoins I recommend (liam sophia) as the best because she can earn $9000 weekly with $700 also contact her if you need help on how to recover your lost funds from scam brokers feel free to contact her today via;liamsophia39@gmail.com or you can also message her on WhatsApp +1 (814)503-0528.

src/views/Bitcoin.js
```diff
@@ -0,0 +1,154 @@
...
+          Bitcoin
+        </Grid.Item>
+        <Grid.Item full as={WarningBox}>
+          We do not recommend using this feature for significant amounts of
+          money, since your private keys are exposed to the browser.
+        </Grid.Item>
+        <Grid.Item full className="mt4" as={Text}>
+          Below is the bitcoin extended public key derived from your point
+          {' ' + pointName}. You can paste it to Landscape to generate an
+          unsigned transaction.
```
jalehman, line 105: I'd remove the reference to Landscape here.

Fang-: From the copy here in general, it's also not clear what this does. What kind of transaction will Landscape generate? Why do I need to go somewhere else to generate that, when it looks like I have all my cryptographic information right here?

src/lib/validators.js
```diff
@@ -132,3 +133,26 @@ export const validateUnique = arr => {
...
+    return 'Invalid Partially Signed Bitcoin Transaction';
+  }
+};
+
+export const validateSignablePsbt = hd => base64 => {
+    const psbt = bitcoin.Psbt.fromBase64(base64);
+    return !psbt.data.inputs.some((input, idx) => {
+      try {
+        //  removing already derived part, eg 'm/84'/0'/0'/0/0' becomes '0/0'
+        const path = input.bip32Derivation[0].path.substring(12);
```
Fang-, line 150: Is the path here a constant or something? Seems a tad dangerous to assume things about the path's prefix/length.

(Also, in the comment, probably don't bother with the enclosing single quotes. It's kind of confusing to read in this context lol.)

src/lib/validators.js
```diff
@@ -132,3 +133,26 @@ export const validateUnique = arr => {
...
+        const prv = hd.derivePath(path).privateKey;
+        psbt.signInput(idx, bitcoin.ECPair.fromPrivateKey(prv));
+        return true;
+      } catch (e) {
+        return false;
```
Fang-, line 155: To make sure I understand correctly: this would crash if the input was, uh, partially signed with a different private key?

src/lib/validators.js
```diff
@@ -132,3 +133,26 @@ export const validateUnique = arr => {
...
+  try {
+    bitcoin.Psbt.fromBase64(base64);
+  } catch (e) {
+    return 'Invalid Partially Signed Bitcoin Transaction';
+  }
+};
+
+export const validateSignablePsbt = hd => base64 => {
+    const psbt = bitcoin.Psbt.fromBase64(base64);
+    return !psbt.data.inputs.some((input, idx) => {
```
Fang-, line 147: Instead of negating, might be simpler to remove the negation here, and say `|| 'No inputs were signed'` below? That feels like the more common pattern.

src/views/Bitcoin.js
```diff
@@ -0,0 +1,154 @@
...
+  const empty = new bitcoin.Psbt();
+
+  const [signedTx, setSignedTx] = useState('');
+  const [psbt, setPsbt] = useState(empty);
+
+  const pointName = ob.patp(need.point(pointCursor));
+
+  const privKey = need.urbitWallet(urbitWallet).bitcoin.keys.private;
+  const pubKey = need.urbitWallet(urbitWallet).bitcoin.keys.public;
+  const chain = need.urbitWallet(urbitWallet).bitcoin.keys.chain;
```
Fang-, line 47:
```suggestion
  const { privKey, pubKey, chain } = need.urbitWallet(urbitWallet).bitcoin.keys;
```

src/views/Bitcoin.js
```diff
@@ -0,0 +1,154 @@
...
+
+  const hd = bitcoin.bip32.fromPrivateKey(
```
Fang-, line 49: `hd` just means "hierarchically determined" in my mind so, here and elsewhere, might be worth renaming to `btcWallet` or something?

src/views/Bitcoin.js
```diff
@@ -0,0 +1,154 @@
...
+    Buffer.from(privKey, 'hex'),
+    Buffer.from(chain, 'hex'),
+    bitcoin.networks.bitcoin
+  );
+
+  const zPub = xpubToYpub(
```
Fang-, line 55: So is it a ypub or a zpub?

My understanding of these is that they're just xpubs but for different bitcoin wallet/signing(?) protocols. Are they interchangeable like this, for our use case?

src/views/Bitcoin.js
```diff
@@ -0,0 +1,154 @@
...
+
+  const onValues = ({ valid, values, form }) => {
+    if (valid) {
+      const newPsbt = bitcoin.Psbt.fromBase64(values.unsignedTransaction);
+      const hex = newPsbt.data.inputs
+        .reduce((psbt, input, idx) => {
+          //  removing already derived part, eg 'm/84'/0'/0'/0/0' becomes '0/0'
+          const path = input.bip32Derivation[0].path.substring(12);
+          const prv = hd.derivePath(path).privateKey;
+          return psbt.signInput(idx, bitcoin.ECPair.fromPrivateKey(prv));
```
Fang-, line 77: The validator does this logic also, should probably be refactored into a `/lib/bitcoin` or whatever. Might be a good place to put `xpubToYpub` as well.

src/views/Point.js
```diff
@@ -130,6 +130,8 @@ export default function Point() {
     isOwner,
   } = useCurrentPermissions();
 
+  const canBitcoin = Just.hasInstance(urbitWallet);
```
Fang-, line 133: Considering this doesn't depend on the specific point in any way, but rather on the login credentials, it's a bit weird that this displays on the point page, alongside point-specific menus.

imo it should at least _also_ be displayed on the `Points` overview page, and maybe shown as in a "separate" category somehow on this individual point page. cc @urcades, @jyng

Fang- (changes requested):
Not sure how to test this, I can't get psbt_faker to give me anything I can paste in here to make it work.

Looking at the flow now, #522 might not be ideal for this, since you're going bridge->landscape->bridge instead of just landscape->bridge, and needing to log in twice is a bit awkward...

src/lib/validators.js
pkova: Rofl at the single quotes, fixed.

The paths in the transaction generated in urbit will always begin with `m/84'/0'/0'`, and if not, this will just fail to sign.

`btcWallet` is derived with `m/84'/0'/0'`, so when we look at the input derivation path of the PSBT we have to derive the remaining part.

src/lib/validators.js
pkova: The point is to allow signing if one or more of the inputs are able to be signed. I noticed that the validation here supported it, but the actual signing in `Bitcoin.js` didn't, so I added 61d0d8e.

src/lib/validators.js
pkova: The the "!test && error" pattern used throughout this file isn't my jam either, but that's what this file seems to be all about.

src/views/Bitcoin.js
pkova: Fixed!

src/views/Bitcoin.js
pkova: Fixed!

src/views/Bitcoin.js
pkova: Actually never mind, let's keep it as it is.

`private` is a reserved word in javascript :(

src/views/Bitcoin.js
pkova: Yeah it's a zpub, I'm just dumb, fixed.

They're kinda interchangeable, but we're doing BIP84 so the correct one to use is zpub.

src/views/Point.js
pkova: Yeah didn't think about this at all, the BTC wallet doesn't have anything to do with points at all! Definitely should be on the `Points` page somehow or in a separate category. Added c8e8454 but need to think about this for a bit.

src/lib/validators.js
Fang-: +1 for trying to match the surrounding code, but I hereby give you permission to flaunt local convention and just do what seems most sane to you. (:

src/views/Bitcoin.js
Fang-: Not sure how that would prevent you from applying the suggested fix or something similar. It's okay to get a little bit creative with the naming. (;

src/lib/validators.js
Fang-: nvm i'm wrong, it's not equivalent.

Fang-:
Paging @urcades, @jyng. Still pending design review, and input on [how to present access to this](#discussion_r570969274). It's a bit weird to display this on the individual point page, but we might also not want to open the points overview page just for this. (Currently, point overview is only shown if you control multiple points or have stars in lockup, otherwise skipping straight to the individual point page.)

Changes so far look good @pkova. There's still a couple unresolved review items from last time, but they're all small ones. I'll probably touch this soon, for flows integration, so that landscape can open Bridge into this.

Fang-:
I went ahead and integrated #522 into this, in 514f31c on the `m/btc-flow` branch.  
No idea how to get the form to _process_ the given `initialValues` instead of just showing them in the form. Maybe we want an explicit "sign this" button anyway, or maybe there's a way to be clever about form behavior? cc @liam-fitzgerald

urcades:
pulling in the PR now locally, will check this out and review soon

Fang-:
For the record, the "try it out" instructions that worked for me:

1. log in as some point,
2. open the bitcoin view,
3. copy out the zpub, convert it to an xpub using [this page](https://jlopp.github.io/xpub-converter/)
4. use [psbt_faker](https://github.com/Coldcard/psbt_faker) to generate a fake tx, `psbt_faker test.psbt 'xpubPasteItHere' -6`
5. copy the contents of `test.psbt` into the input field in bridge

@urcades lmk if you need any bridge setup or testing help, I'll be around.

urcades:
Hello, I've been unable to serve the built bridge interface, but I've been given enough background to provide some feedback on the current interface (shown above).

Right now, here's a rough sketch of the basic progression we expect when performing a bitcoin transaction ... for now***:

Screen Shot 2021-02-26 at 8 09 04 PM

The leftmost screen in this image displays the last stage/step of the "Landscape Crypto Invoicing" flow, which is to produce an unsigned transaction hash, which would be auto-copied to the user's clipboard (or manually copied, if auto-copying is blocked for some reason) upon clicking the "Copy Transaction and Sign in Bridge" button.

This button would open up a window to bridge.

After logging in using whatever method a pilot has associated with their Urbit ID, they should be brought immediately to the bitcoin page shown above. The copied transaction can then be copied into the "Unsigned Transaction" field.

We should add an explicit button under this field so that an urbit pilot can explicitly press a "Sign Transaction" button after pasting the transaction in the field to produce the signed transaction hash.

This modified form design can be structured as such:

Screen Shot 2021-02-26 at 9 03 15 PM

Basically, I think we should make the two sections for 1) **Setting Extended Pubkey in Landscape** and 2) **Transaction Signing** very, very distinct, and make it obvious that the first section is set only once (when setting up a wallet), while the second section is used every time a transaction is made.

Also, the second signing section should really have distinct 'states' for each stage of the signing process.

Would appreciate any feedback on the above.

***So, it goes without saying that I think this entire process needs to be way, way better than it's currently being built, and ideally not require the manual copy-pasting of hashes between Landscape and Bridge. The UX of this is really not up to par with how we're building up Landscape, and while I acknowledge that the work is being completed to the specification outlined by the grant that enabled it, we should ensure that anything slotted into landscape meets the highest bar of usability.

urcades (approved):

Fang- (approved):
Alright, after talking out of band, I think this is all good now. Thanks so much for your work and patience @pkova!

Merging this into a `bitcoin` branch for flows integration, final touches etc.

PR #553, opened 2021-02-16
useGasPrice: use safeLow && surface expected wait

liam-fitzgerald:
Capture d’écran 2021-02-16 à 12 17 32 PM

use safeLow and surfaces time to completion from the gasstation api

### DIFF
--- a/src/lib/getSuggestedGasPrice.js
+++ b/src/lib/getSuggestedGasPrice.js
@@ -22,10 +22,12 @@ export default async function getSuggestedGasPrice(networkType) {
 
         // ethgasstation returns values in floating point, one order of magitude
         // more than gwei. see: https://docs.ethgasstation.info
-        const suggestedGasPrice = Math.ceil(json.fast / 10); // to gwei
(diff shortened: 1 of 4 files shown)


src/lib/getSuggestedGasPrice.js
```diff
@@ -22,10 +22,12 @@ export default async function getSuggestedGasPrice(networkType) {
 
         // ethgasstation returns values in floating point, one order of magitude
         // more than gwei. see: https://docs.ethgasstation.info
-        const suggestedGasPrice = Math.ceil(json.fast / 10); // to gwei
+        const suggestedGasPrice = Math.ceil(json.safeLow / 10); // to gwei
+
+        const waitTime = json.safeLowWait;
 
         // we don't want to charge users more than the gas tank funds
-        return Math.min(suggestedGasPrice, MAX_GAS_PRICE_GWEI);
```
Fang-, line 28: You'll want to update the other cases in this file to produce similar results. Right now, trying this with Ropsten selected straight-up crashes. You're probably fine suggesting 15 seconds for Ropsten and one second for local. I think you'd pass `undefined` in the `catch` case here?

Issue #554, opened 2021-02-21
Error Report: Cannot read property 'map' of null


hdzdwlrs:
## Context

Logging into bridge with metamask

A wallet containing multiple planets

## Error

```
TypeError: Cannot read property 'map' of null
    at Object.Ok (https://bridge.urbit.org/static/js/main.b2800745.chunk.js:2263:2859)
    at InternalConstructor.matchWith (https://bridge.urbit.org/static/js/2.de79ebd6.chunk.js:33518:29)
    at https://bridge.urbit.org/static/js/main.b2800745.chunk.js:2263:2771
    at InternalConstructor.chain (https://bridge.urbit.org/static/js/2.de79ebd6.chunk.js:65133:14)
    at https://bridge.urbit.org/static/js/main.b2800745.chunk.js:2263:2750
    at Object.Mh [as useMemo] (https://bridge.urbit.org/static/js/2.de79ebd6.chunk.js:174847:7)
    at push.exports.useMemo (https://bridge.urbit.org/static/js/2.de79ebd6.chunk.js:170996:14)
    at Points (https://bridge.urbit.org/static/js/main.b2800745.chunk.js:2263:2728)
    at oh (https://bridge.urbit.org/static/js/2.de79ebd6.chunk.js:174592:7)
    at di (https://bridge.urbit.org/static/js/2.de79ebd6.chunk.js:175212:7)
```

In Bridge version 2.5.12.

Fang-:
Alright, found out I can consistently reproduce this by first loading Bridge, and then (after pageload) changing the network that's selected in Metamask.  
This wasn't an issue before because changing the selected network in Metamask would make it reload the page. Metamask stopped doing recently.

Pushing in the simple fix for that today, but it's unclear that covers all cases of this error. #559 mentioned still encountering the issue after reloading the page.

In other words, this issue is a symptom of _some_ Ethereum connectivity issue. Why read calls would give `null` results instead of throwing exceptions, I do not yet understand...

thelifeandtimes:
Closing out as stale. As of v3.2.14, Metamask has changed it's 'active network' model such that you can have any number of networks enabled at once and the network you are interacting with is selected at the time of transacting, rather than connecting.

PR #560, opened 2021-03-19
Prevent "Cannot read property 'map' of null" jank

Fang-:
As mentioned in https://github.com/urbit/bridge/issues/554#issuecomment-802810525, a confirmed trigger for this crash is Metamask changing the target chain out from under us. So we add a handler for that, and fully reload the Bridge page, to help us get off to a clean start.

Also as previously mentioned, I suspect this isn't the only trigger for the crash, but I'm struggling to come up with further reproduction. To help users out, I added a smol error state/screen.

image

That links to [this issue template](https://github.com/urbit/bridge/issues/new?body=%3C%21--+Please+provide+some+context.+Do+you+have+Metamask+installed%3F+What+login+method+did+you+use%3F+--%3E&title=Failed+to+read+the+blockchain.). Even if we don't get super useful information out of those reports, still useful to know that they're happening, and whether or not Metamask is always in the picture.

### DIFF
--- a/src/Bridge.js
+++ b/src/Bridge.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { useEffect } from 'react';
 import { hot } from 'react-hot-loader/root';
 import { Just, Nothing } from 'folktale/maybe';
 import { IndigoApp } from 'indigo-react';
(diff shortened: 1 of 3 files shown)


Fang-:
Going ahead and cowboy-merging this, since I'm preparing a release rn and it can't be more broken that what's already live.

PR #563, opened 2021-03-26
release: support transferring lockups

Fang-:
Adds support for ~~another footgun~~ transferring of lockups.

We get transfer approval. Remember that transfers of lockups are mandatory two-phase transfers, where the sender first "approves" of the recipient taking the lockup:

image
After tx confirmation:
image

(I wanted to make the url here click-to-copy, but that makes it overflow to the right instead of wrap, and I'm too dumb to fix that.)

And then we get transfer acceptance through a url-parameter-activated flow, which finalizes the transfer, and forwards the recipient to the lockup page. We implement this as a flow to work around UI complications. (Namely, "when/where would we show a button for this?") If you're transferring a lockup to someone, it seems a reasonable assumption you're talking to them already anyway.

image

First three commits here are cherry-picked from #522 as currently lives on the bitcoin branch and have already been reviewed.


liam-fitzgerald (approved):
LGTM!

Fang-:
Added a little notification to the point transfer page, for galaxies who have stars in lockup.

image

Pretty trivial, so going ahead and merging it in!

PR #569, opened 2021-05-12
senate: support voting on upgrade proposals

Fang-:
Adds a page for voting on contract upgrade proposals. This largely mimics the behavior of the existing document voting page. Adds a senate navigation page so both are accessible separately.

image

After clicking into the upgrades page:

image

Also tweaks both these voting pages to not show the "open" list and transaction buttons, if there are no open polls.

The two voting pages share some amount of code/logic, but I'm more concerned with getting this out than refactored right now. Unifying the two pages under the hood might be a good contributor task, though new Bridge designs might call for this being replaced wholesale.


src/views/Senate/Documents.js
```diff
@@ -0,0 +1,220 @@
...
+  const doVote = useCallback(
+    (hash, accept) => {
+      construct(hash, accept);
+      setVotingOn({ hash, accept });
+    },
+    [construct, setVotingOn]
+  );
+
+  const didVote = useCallback(() => {
+    polls[votingOn].hasVoted = true;
```
liam-fitzgerald, line 112: Does this rerender correctly? I'd be surprised if it did because the object references are the same. It's more idiomatic to use pass a function to setPolls and update the state immutably. Plus, it lets you drop polls from the dependency list

src/views/Senate/Upgrades.js
```diff
@@ -0,0 +1,195 @@
...
+    }
+    return [open, closed];
+  }, [loading, proposals, polls, now]);
+
+  const doVote = useCallback(
+    (address, accept) => {
+      construct(address, accept);
+      setVotingOn({ address, accept });
+    },
+    [construct, setVotingOn]
```
liam-fitzgerald, line 100: setters are unnecessary in dependency lists

src/views/Senate/Upgrades.js
```diff
@@ -0,0 +1,195 @@
...
+  );
+
+  const didVote = useCallback(() => {
+    polls[votingOn].hasVoted = true;
+    setPolls(polls);
```
liam-fitzgerald, line 105: See other comment

src/views/Senate/Upgrades.js
```diff
@@ -0,0 +1,195 @@
...
+            <Grid.Item full as={Grid}>
+              <Grid.Item full as={P}>
+                Open upgrade polls:
+              </Grid.Item>
+              {openList}
+            </Grid.Item>
+
+            <BridgeForm>
+              {({ onSubmit, values }) => (
+                // <>
```
liam-fitzgerald, line 181: Unnecessary?

src/views/Senate/Upgrades.js
```diff
@@ -0,0 +1,195 @@
...
+
+  const [proposals, setProposals] = useState([]);
+  const [polls, setPolls] = useState({});
+  const [loading, setLoading] = useState(true);
+  const [votingOn, setVotingOn] = useState({});
+
+  //TODO update every second?
+  const now = Math.round(new Date().getTime() / 1000);
+
+  useMemo(async () => {
```
liam-fitzgerald, line 61: What's with the use of useMemo for side effects instead of useEffect?

liam-fitzgerald (changes requested):
Some questions about rerenders and general style nits

src/views/Senate/Documents.js
Fang-: It does, actually! (After fixing the other ways in which the logic here was broken, that is...) But I'll update based on your advice regardless. Thanks!

src/views/Senate/Upgrades.js
Fang-: Probably past me not knowing what he was doing and struggling with async useEffect? Fixed!

liam-fitzgerald (approved):

PR #573, opened 2021-06-07
bitcoin: add flow for viewing bitcoin extended key

pkova:
We want to encourage people to do the key-sensitive BTC stuff with Bridge instead of wildly throwing around their master ticket in the BTC wallet app, so we need a flow for the xpub as well.

Also cleaned up whatever the linter complained about in `Login.js` and more importantly bumped `azimuth-solidity` to `1.2.3`, the development environment wouldn't compile otherwise.

### DIFF
--- a/src/views/Login.js
+++ b/src/views/Login.js
@@ -1,51 +1,21 @@
 import React, { useCallback, useState } from 'react';
-import { P, H4, H5, Grid, Text, Button, Flex, LinkButton } from 'indigo-react';
+import { Grid, Text, Button, Flex, LinkButton } from 'indigo-react';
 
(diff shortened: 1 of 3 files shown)


src/views/Login.js
```diff
@@ -63,11 +63,16 @@ export default function Login() {
       switch (flow.kind) {
         case COMMANDS.TAKE_LOCKUP:
           push(names.ACCEPT_LOCKUP);
+          break;
```
Fang-, line 36: Good catch, thanks!

Fang- (approved):
This all looks good. Thank you! Release later today probably.

PR #582, opened 2021-06-24
store: more nuanced syncing

Fang-:
This renames some of the syncing functions for clarity, and syncs less aggressively on the points list screen.

Previously, when loading the points list, we would fetch all kinds of details that only come into play once the user actually navigates to a point. This isn't a huge deal for two or three points, but for accounts that control double- or triple-digit points (not uncommon in the "star spawning planets" case), this would result in thousands of requests to the Ethereum node.

(Yes, thousands. This change brings it back to hundreds, but Bridge is still obviously doing way too much work here. Further changes will follow, some day...)

Requesting review from two people here, but review from either would be sufficient as well.

### DIFF
--- a/src/components/Passport.js
+++ b/src/components/Passport.js
@@ -7,7 +7,7 @@ import 'style/anim.css';
 import * as need from 'lib/need';
 import { chunkStr, Matrix, walk, rand } from 'lib/card';
 import usePermissionsForPoint from 'lib/usePermissionsForPoint';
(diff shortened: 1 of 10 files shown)


liam-fitzgerald (approved):
LGTM!

tomholford (approved):
@Fang- Nice work. Next time we chat, curious to hear more details about this change, I think it would be educational.

PR #584, opened 2021-06-28
deps: modernization and upgrades

### LINKED ISSUE #410: Should update web3 (toWei errors in minified builds only)
image

tomholford:
# Context

The JS ecosystem moves quickly 🏎️  This PR brings Bridge up-to-date in several crypto libs, and removes unused development libs (since replaced by latest version of `react-scripts`). It also seems to resolve the issue we were having with unreliable builds.

# Changes
- upgrade: keccak, keythereum, secp256k1, truffle, web3
- remove: react-hot-loader, @hot-loader/react-dom
- converted a handful of files to Typescript and added typings
- reorganized some crypto utils and constants from the Wallet.ts file
- update README with correct Node version

# Testing

Since there are no automated tests (yet!), we manually tested these changes by following these steps:

1. `rm -rf node_modules/` - first, remove existing deps
2. `npm install` - reinstall JS packages and rebuild native extensions
3. `npm run build` - generate production-optimized build
4. `npm run pilot` - confirm that local development environment starts successfully

Once confirming the app was building and loading successfully, we then tested the following flows:
- Mnemonic Login
- Galaxy Creation
- Star Creation
- Planet Creation
- Invite Acceptance
- Passport Download
- New Master Ticket (from Passport) Login

Fixes #410


src/lib/flowCommand.ts
```diff
@@ -21,9 +21,16 @@ const COMMANDS = {
   XPUB: 'xpub',
 };
 
+type FlowType = {
+  kind?: 'takeLockup' | 'btc' | 'xpub';
+  lock?: 'linear' | 'conditional';
+  from?: string;
+  utx?: string;
+} | null;
```
Fang-, line 29: Based and typespilled.

But this doesn't quite capture what we want/expect here. The keys are mutually exclusive based on the `kind`. I think you want something like:

```suggestion
type FlowType =
  | { kind: 'takeLockup', lock: 'linear' | 'conditional', from: string }
  | { kind: 'btc', utx: string }
  | { kind: 'xpub' }
  | null;
```

src/lib/constants.ts
```diff
@@ -74,9 +74,33 @@ const PROFILE_STATES = {
 
 const PROGRESS_ANIMATION_DELAY_MS = 500; // .animated-width
 
+const DEFAULT_HD_PATH = "m/44'/60'/0'/0/0";
+const ETH_ZERO_ADDR = '0x0000000000000000000000000000000000000000';
+const ETH_ZERO_ADDR_SHORT = '0x0';
```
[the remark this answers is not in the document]
Fang-: I guess it's nice that files are now more clear about whether they import wallet constants vs wallet logic.

src/lib/utils/crypto.ts
```diff
@@ -0,0 +1,39 @@
+import { reduce } from 'lodash';
+import { isAddress, keccak256, toHex } from 'web3-utils';
+import { publicToAddress } from 'ethereumjs-util';
+import { ETH_ZERO_ADDR, ETH_ZERO_ADDR_SHORT } from 'lib/constants';
+
+export const stripHexPrefix = (hex: string) => {
+  return hex.startsWith('0x') ? hex.slice(2) : hex;
+};
+
+export const addHexPrefix = (hex: string) => {
```
Fang-, line 10: These are all just hex/Ethereum address functions, so the `crypto.ts` name feels a bit misleading. Perhaps `utils/address.ts`?

src/store/lib/useControlledPointsStore.js
```diff
@@ -61,6 +61,7 @@ export default function useControlledPointsStore() {
         );
       }
     } catch (error) {
+      console.error(error);
```
Fang-, line 64: If we want to leave this in (which is reasonable) we might want to tack on a little message.

```suggestion
      console.error('failed to fetch controlled points', error);
```

src/lib/utils/crypto.ts
```diff
@@ -0,0 +1,39 @@
...
+    (acc, char, idx) =>
+      parseInt(hash[idx], 16) >= 8 ? acc + char.toUpperCase() : acc + char,
+    '0x'
+  );
+};
+
+export const eqAddr = (addr0, addr1) =>
+  toChecksumAddress(addr0) === toChecksumAddress(addr1);
+
+export const bufferedPublicKeyToAddress = (publicKey: Buffer): string => {
```
Fang-, line 37: >type in the function name

That's bad naming sense! Can't we just call this `publicToAddress` and shadow/overload the original? I think everywhere we want an address we want it as a `0xabc..` hex string?

src/types/bridge-wallet.type.ts
```diff
@@ -0,0 +1,6 @@
+import { BIP32Interface } from 'bitcoinjs-lib';
+
+interface BridgeWallet extends BIP32Interface {
+  address: string;
+  passphrase: string;
+}
```
Fang-, line 6: This looks vaguely incorrect (we may not always know the `passphrase`), but it's also not used anywhere afaict. Probably don't want to add dead code.

Fang- (commented):
Nice! Tested this locally, seems to work fine for all the broad/important cases. Some questions and comments below, and out of band.

src/lib/flowCommand.ts
tomholford: Good point, this communicates the actual use case.

Like the `BridgeWallet` type, this type was created in the midst of a debugging session, but ultimately I didn't up using it :) So, will remove it from the PR ✂️

Fang- (approved):
As discussed out of band, this now works just fine across flows and devices, as far as I can tell. Great work here! Can't wait to finally push out minimized builds again. (:

PR #587, opened 2021-06-30
store: finish syncOwnedPoint -> syncExtras rename

Fang-:
In #582 we had renamed some syncing functions, but neglected to replace-all for the `syncOwnedPoint` case. This completes what was started.


tomholford (approved):
LGTM

PR #592, opened 2021-07-05
ux: Urbit OS view and flow enhancements

tomholford:
# Context

As part of the larger overall L2 / PKI effort, several pieces of the existing flow need to be tweaked to accommodate the upcoming changes. This is also a good chance to address a few low-hanging fruit changes based on feedback from existing users and the Tlon design team.

# Changes

- extract `Hosting` component to its own file; stop rendering it (for now) in the Urbit OS Home view
- consistent look and feel of alerts across pages
- conditionally render menu options and notifications to improve UX
- copy and design updates based on user and design team feedback

# Preview

The screenshots below depict the different views before and after the changes in this PR.

### Before
image

### After
image

### Before

image

### After
image

### Before

image

### After

image

### Before

image

### After

image


jyng:
I just changed the copy here from `Network` to `Urbit OS. This looks good on my end, @yosoyubik you wanna take a quick look at the code?

jyng (approved):

src/types/PointDetailsType.ts
```diff
@@ -0,0 +1,28 @@
+// Response returned by azimuth#getPoint
```
yosoyubik, line 1: wondering what's the best way to handle this for the layer-2 types... 

[`roller.getPoint`](https://urbit.github.io/roller-rpc-client/index.html#getpoint) returns this type: https://urbit.github.io/roller-rpc-client/interfaces/point.html)

tomholford: Good question. One idea is to take advantage of Typescript [namespaces](https://www.typescriptlang.org/docs/handbook/namespaces.html); alternatively, we can go with something simpler like naming them explicitly (e.g., `L1Point`, `L2Point`). It's less complex and quicker to go with the latter solution, but potentially more maintainable in the long run to use namespaces. WDYT?

yosoyubik: I like simple :D I'll update the Roller to return an L2Point

yosoyubik (approved):

src/types/PointDetailsType.ts
```diff
@@ -0,0 +1,28 @@
...
+  0: string;
+  1: string;
+  2: string;
+  3: string;
+  4: string;
+  5: string;
+  6: string;
+  7: string;
+  8: string;
+  9: string;
```
Fang-, line 12: I think all of these have named equivalents below, and so we don't care about the numbered accessors? Might be fine to just leave these out, they're not really helpful.

src/views/UrbitOS/Home.tsx
```diff
@@ -29,6 +31,10 @@ export default function UrbitOSHome({ manualNetworkSeed }) {
   const showSponsor = azimuth.getPointSize(point) !== azimuth.PointSize.Galaxy;
   const toggleShowKeys = useCallback(() => setShowKeys(s => !s), [setShowKeys]);
 
+  // The initial key revision number is 0, and increments when set
+  const hasSetNetworkingKeys = details.keyRevisionNumber !== '0';
+  const networkKeysAction = hasSetNetworkingKeys ? 'Reset' : 'Initiate';
```
Fang-, line 38: "Initiate networking keys" feels a tad strange compared to "initialize networking keys", but maybe it's fine.

We may also want "re-set" (note the dash) here instead, to emphasize changing rather than just wiping?

But I also haven't spoken to @jyng about phrasing around these things recently, feel free to push back.

Fang- (approved):
Some nits, but otherwise looks good!

You would've been fine to just nuke the Hosting page's code entirely, it'll stick around in git history. But keeping it in a file is fine too if we intend to get back to it within, uh, a year or something? We just shouldn't let dead code pile up is all.

src/views/UrbitOS/Home.tsx
tomholford: Any of these are fine with me, will defer to @jyng on how to proceed from here.

jyng: Initialize is better. Lets change that here and also in the alert copy. I'm not sure what you mean Re-set, it seems like it would be strangely confusing to the user. We can also revisit this later.

Fang-: > I'm not sure what you mean Re-set, it seems like it would be strangely confusing to the user. 

As in, you're not "factory _reseting_" your keys, but rather "_setting_ them to a new value". But you're right, we can change whenever if we feel it's better.

Fang- (approved):
Took the liberty of pushing some smol fixes. All good now! (:

PR #594, opened 2021-07-14
dx: add Ropsten mode for local development

tomholford:
# Changes

- new npm scripts: 
  - `npm run pilot-ropsten` - spins up a local instance of Bridge, pointing to Ropsten testnet
  - `npm run build:ropsten` - builds the site bundle pointing to Ropsten
  - `npm run release:ropsten` - same as the current `release` script, with one difference: calls the `build:ropsten` script
- indicate visually if using Ropsten (see screenshot below)

# Preview

image


Fang-:
This is nice!

I often do a Ropsten build & deploy prior to the "real" launch. Could you add a command for that?

Additionally, it would be nice if the version number displayed on the login page could get an `r` appended to it when configured to point to Ropsten.

src/views/Login.tsx
```diff
@@ -124,7 +125,10 @@ export default function Login() {
             className="us-none pointer">
             <span className="underline">Offline</span> ↗
           </Flex.Item>
-          <Flex.Item className="gray4">v{version}</Flex.Item>
+          <Flex.Item className="gray4">
+            v{version}
+            {isRopsten ? 'r' : ''}
+          </Flex.Item>
```
Fang-, line 131: This is good, but is there a way to modify the `version` directly, rather than just affecting its displayed value? We also use it in the `zip` command, as the filename.

(This isn't actually a big deal, considering I never build .zip releases for Ropsten, but just wondering if there's a higher-level approach here.)

tomholford: Good thinking, just made a few tweaks to allow this value to be set via an env var (`REACT_APP_BRIDGE_VERSION`); it will fallback to the version from `package.json` if unset. And, it will use this value for the UI and zip filename.

PR #595, opened 2021-07-16
urbit-os, urbit-id: fix non-functional back button

pkova:
Bruh the back buttons don't work.


Fang- (approved):
Good catch, thank you!

PR #599, opened 2021-07-27
ux: restore Access Code to Urbit OS view

### LINKED ISSUE #598: Ability to view/copy login code missing from Bridge
For the past week, the 'Login code' section has been missing from the 'Urbit OS' page on Bridge when logged into a planet account with the master ticket. 

Not in 'Urbit OS':
no-login-code-bug

Not in 'Urbit ID':
no-login-code-bug-2

System: Seen on both Chrome and Safari. Confirmed by multiple users.

tomholford:
# Context

As part of a recent [refactor of the Urbit OS view](https://github.com/urbit/bridge/pull/592), a button to copy one's Access Code was removed; this PR restores it :)

Fixes #598 

# Changes

- Refactor `CopyButton` component to use `@tlon/indigo-react` `Icon`
- Test coverage for new `AccessCode` component
- Dependencies for automated testing

# Preview

optimized


src/views/UrbitOS/AccessCode.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+describe('AccessCode', () => {
+  describe('when code is not available', () => {
+    // Value is initialized as `false` in `useKeyfileGenerator`
+    const testCode = false;
+
+    it('does not render the component', () => {
+      render(<AccessCode code={testCode} />);
+      const accessCode = screen.queryByText(
+        'This is your code to access Urbit OS'
```
Fang-, line 13: Wouldn't it be better to just test the rendered document for being empty or not? The way the tests are currently written makes them depend on (and thus change with) the exact copy used, which I imagine we want to avoid? (I don't actually know much about interface testing, so you tell me!)

Fang- (changes requested):
There is no visual feedback about what happens when I click the `CopyButton`. The address copy button does give such feedback, for example:

image

Also have a comment about the way the tests are written. They're cool though, thanks for adding. And the code itself looks fine!

Also awaiting copy approval on this.

src/views/UrbitOS/AccessCode.test.tsx
tomholford: >Wouldn't it be better to just test the rendered document for being empty or not?

Found a way to do so thanks to [the docs](https://testing-library.com/docs/guide-disappearance/#asserting-elements-are-not-present), good suggestion.

>The way the tests are currently written makes them depend on (and thus change with) the exact copy used, which I imagine we want to avoid?

It could be a good smoke test in the truthy case, it's a trivial way of ensuring we don't insert a typo in the copy while working on that file. Also, [it's recommended](https://testing-library.com/docs/guiding-principles) by `testing-library` to use text strings in tests, because it's simulates how users navigate the real app.

src/components/Buttons.js
```diff
@@ -23,3 +25,14 @@ export const OfflineButton = props => (
...
 );
+
+export const CopyButton = ({ didCopy = false, ...props }) => (
+  <Button
+    accessory={
+      <WithTooltip content={didCopy ? 'Copied!' : 'Copy'}>
+        <Icon icon="Copy" size="18px" color={'black'} />
+      </WithTooltip>
+    }
+    {...props}
```
Fang-, line 36: Pretty sure we want `WithTooltip` around the entire button. Right now I can click the entire thing, but only see the feedback if I hover the icon.

Also, the tooltip feels a bit on the large side, font-wise. Might be nice to shrink it down to the size of the button's description.

src/views/UrbitOS/AccessCode.tsx
```diff
@@ -14,9 +14,7 @@ const AccessCode = ({ code }: AccessCodeArgs) => {
         detail="This is your code to access Urbit OS"
         as={CopyButtonWide}
         text={code}
-        data-testid="access-code"
-        // Allow the Copy Button Tooltip to flow over the Grid border
-        style={{ overflow: 'unset' }}>
+        data-testid="access-code">
```
Fang-, line 17: Dang these actually end up in the html of production builds. ):

Fang- (approved):
All good! Thank you. (:

PR #600, opened 2021-07-31
ux: WalletConnect login

### LINKED ISSUE #601: Support WalletConnect such that we can log in using Gnosis Safe

tomholford:
Closes #601

# Changes

- adds [WalletConnect](https://walletconnect.org/) login flow

# Preview

## Ready to Connect

image

## Connected, ready for Auth

image

## Login Flow

This was tested using the [WalletConnect online test wallet](https://test.walletconnect.org/)

scaledOutput

# Testing

To verify functionality, I downloaded several apps for both iOS and Android. After some trial and error, I realized that many wallet apps automatically reject Signing Transactions if the connected wallet has a zero balance. Once I connected a main net wallet with some funds, I was able to confirm that both Trust wallets for iOS and Android (in addition to the reference implementation in the Test Wallet) are working as expected. As for the remaining unknown wallets, I stopped testing once I found 3 working peer implementations so we can get this out the door and focus on Roller.

In summary, confirming that this works across several WalletConnect peer implementations (reference Test Wallet, Trust, and TokenPocket). Thus, if other wallets do have issues once we ship to production, we should encourage affected users to file an issue with vendor's repo.

image


tomholford:
@Fang- A few items for when you have a moment:

-  When initially loading the WC Login component, some event listeners are instantiated to handle incoming events from the peer device. After the User authenticates and is redirected to the Home view, should the application re-instantiate WC event listeners? If it does, then if the User disconnects on their WC peer, Bridge could log the user out as well. However, this might be unexpected behavior, as it might violate the principal of least surprise. Would a user expect to be auto-logged out of Bridge by disconnecting on the peer device?
- Have tested this so far with WC Test Web Wallet, and both Coinomi and Rainbow for iOS. The test wallet and Rainbow are pretty reliable, but Coinomi would crash on occasion when authenticating. Do you have any devices with which you could also test the flow? Would be helpful to collect additional data points
- The remaining piece - allowing the user to set a custom HD Path - is dependent on using the `BridgeForm` components built-in behaviors (such as `onSubmit`); currently the implementation is hacking around it. Got stuck when I was trying to figure it out the other day, will take another crack at it today. Depending on how things go, I might be reaching out again :)

tomholford:
@jyng If you have a moment, please LMK if you have any thoughts or feedback, thanks

jyng:
1) We should use the same button type as the other options.
2) I seem to only be able to click "Connect" once. If I close it and try to click "Connect" again, nothing happens. No log in console either.

tomholford:
> 1. We should use the same button type as the other options.
Got it, will make that change 👍 

> 2. I seem to only be able to click "Connect" once. If I close it and try to click "Connect" again, nothing happens. No log in console either.
Okay thanks for sharing this, will investigate to see if I can reproduce the issue. What browser and WalletConnect peer are you using?

tomholford:
@jyng Addressed both issues, thanks for your feedback! Please LMK if you have any other thoughts, thanks

image

jyng:
> Connect a peer that support WalletConnect
Is 'peer' a term that WalletConnect uses? If not, I think 'wallet' should replace it.

src/lib/useWalletConnect.ts
```diff
@@ -0,0 +1,202 @@
...
+      address,
+      connector,
+      walletType: WALLET_TYPES.WALLET_CONNECT,
+    });
+
+    // Inspired by the MetamaskWallet implementation :)
+    // TODO: We should refactor / unify how we handle wallets
+    const wallet: WalletConnectWallet = {
+      address,
+    };
```
Fang-, line 98: I mean, we do `setWallet` with a consistent type, don't we? But there is some necessary inconsistency between wallets when it comes to tx signing/sending and such.

src/lib/useWalletConnect.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    setAuthToken(Just(authToken));
+    setWallet(Just(wallet));
+    setWalletHdPath(hdPath);
```
Fang-, line 102: Does this ever come into play anywhere? It looks like in the WalletConnect case, we just connect to some specific account, but don't know anything about its origin/corresponding path. Might be misleading to set it/provide a login input for it.  
I think we only really use it in hardware wallet scenarios, where we need to tell the device to sign with a specific account?

Fang- (changes requested):
When I try to generate & sign a (planet acceptance) transaction, I get `can't access property Symbol.iterator, source is undefined`. This is with a WalletConnect test wallet.

As mentioned out of band, I can't get this to work with Ledger Live (it opens the app, but the app doesn't actually do anything). May or may not be something we have control over.

galenwp:
very happy to see walletconnect going in!

src/lib/useWalletConnect.ts
tomholford: Good call, removed this 👍

src/lib/useEthereumTransaction.js
```diff
@@ -119,13 +128,15 @@ export default function useEthereumTransaction(
...
             gasLimit,
+            txnSigner,
           })
         )
       );
 
       setSignedTransactions(txns);
       setState(STATE.SIGNED);
     } catch (error) {
+      console.error(error);
```
Fang-, line 139:
```suggestion
      console.error('error signing txs', error);
```

We might not always be able to see a legible trace of where the error came from, so labeling it seems nice.

Fang- (commented):
Changes herein all look good. I'll want to go and test all the various login/signing methods to make sure no breakage is sneaking by, but otherwise this is in a good state.

What we should probably do, considering the flakiness of some WalletConnect clients, is add a tiny little disclaimer in its login box, saying something like "WalletConnect is a young protocol, not all wallets may work fully" or something along those lines, to give our "yes we know" responses to the inevitable "my wallet doesn't connect!" reports a little more oomph.

tomholford:
> Changes herein all look good. I'll want to go and test all the various login/signing methods to make sure no breakage is sneaking by, but otherwise this is in a good state.
> 
> What we should probably do, considering the flakiness of some WalletConnect clients, is add a tiny little disclaimer in its login box, saying something like "WalletConnect is a young protocol, not all wallets may work fully" or something along those lines, to give our "yes we know" responses to the inevitable "my wallet doesn't connect!" reports a little more oomph.

Good thinking. Since the action is obvious (click the big Connect button), replaced the previous help text with the disclaimer:

image


Also @Fang- , added a *Testing* section to the PR description above. In summary:
>Confirming that this works across several WalletConnect peer implementations (reference Test Wallet, Trust, and TokenPocket).

Feeling pretty comfortable about this now, will re-request review 🔁

Fang- (approved):
Alright, I think this is good for release. Let's do this!

urcades:
this is huge, thanks so much @tomholford

PR #602, opened 2021-08-12
star release: fix string<->int shenanigans

Fang-:
For linear releases, this was causing us false positives on the "does
this address have a lockup" check. Subsequently calling the withdrawe
limit function for an address without a lockup would result in a
contract-side exception being thrown due to the division by zero. This
exception wasn't getting handled properly, and prevented users with
conditional lockups from actually viewing those in Bridge.

So now we make sure to diligently convert received values into actual
integers, instead of the strings we get them as.

For conditional releases, this wasn't actively causing any problems
(that we were aware of), but we make similar changes there anyway,
for peace of mind.

### DIFF
--- a/src/lib/starRelease.js
+++ b/src/lib/starRelease.js
@@ -82,21 +82,28 @@ export async function getConditional(contracts, address) {
     conditionalSR.getRemainingStars(contracts, address),
   ]);
 
-  if (commitment === null || batches === null) {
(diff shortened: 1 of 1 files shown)


tomholford (approved):
Enjoyed the detailed comments, it made this easy to follow.

In Bridge V2 we will favor Typescript when possible to prevent this common JS 'feature' :)

PR #605, opened 2021-08-16
wallet: make publicToAddress more robust

### LINKED ISSUE #604: Can't log in with private key
Tried on both Brave (shields down & up) & Chrome. As soon as it is pasted:

```
Error: This method only supports Buffer but input was: [redacted]
    at Object.t.assertIsBuffer (https://bridge.urbit.org/static/js/2.1631767b.chunk.js:2:1375645)
    at t.pubToAddress (https://bridge.urbit.org/static/js/2.1631767b.chunk.js:2:1790671)
    at f (https://bridge.urbit.org/static/js/main.05e71d06.chunk.js:1:40345)
    at new j (https://bridge.urbit.org/static/js/main.05e71d06.chunk.js:1:68666)
    at https://bridge.urbit.org/static/js/main.05e71d06.chunk.js:1:49777
    at https://bridge.urbit.org/static/js/main.05e71d06.chunk.js:1:38488
    at Ml (https://bridge.urbit.org/static/js/2.1631767b.chunk.js:2:3499838)
    at t.unstable_runWithPriority (https://bridge.urbit.org/static/js/2.1631767b.chunk.js:2:3517651)
    at zi (https://bridge.urbit.org/static/js/2.1631767b.chunk.js:2:3440009)
    at Pl (https://bridge.urbit.org/static/js/2.1631767b.chunk.js:2:3499299)
```

Bridge v2.9.5

Fang-:
Apparently Uint8Arrays aren't quite enough like Buffers as far as
ethereumjs-util is concerned. Here we make sure we always pass in a
Buffer by creating one from whatever argument was passed in.

Arguably this needs fixing as the callsites instead, happy to discuss.

Fixes #604.

### DIFF
--- a/src/lib/utils/address.ts
+++ b/src/lib/utils/address.ts
@@ -35,5 +35,5 @@ export const eqAddr = (addr0, addr1) =>
   toChecksumAddress(addr0) === toChecksumAddress(addr1);
 
 export const publicToAddress = (publicKey: Buffer): string => {
-  return toHex(pubToAddress(publicKey, true).toString('hex'));
+  return '0x' + pubToAddress(Buffer.from(publicKey), true).toString('hex');
 };


ngzax:
@Fang- : Big thanks for this patch. However. ;)

I cloned this repo, did a `npm ci`, use mkcert to create the local certs for https, and then used `python bridge-https.py` and then navigated the `https://localhost:4443` I get a dir list.

image

Fang-:
@ngzax Suffering will continue until Bridge improves...

And I can't even repro because my python environment is messed up. Lovely.

Dumbest fix for this might just be changing [this](https://github.com/urbit/bridge/blob/08a6c155cbc574b23d1552ddfb72ad66efc9bf93/src/Bridge.js#L25) to say `.MAINNET` and then running with `npm run pilot`, which should serve over https.

Fang-:
Actually, does it maybe work if you `cd build` prior to `python ../bridge-https.py`?

ngzax:
yeah, `npm ci` is just "install exactly according to package.json, don't update anything." i will try your suggestions above and report back.

ngzax:
ok. good news. I have validated both that this method of running bridge from local build works and that your patch now allows me to log in with my private key. I will open a new PR with an update to the README to reflect the current state of bridge.

FWIW, all this arose because I am trying to get rid of this wallet that I can only access via PK and move it to my Ledger H/W wallet, but that's not seeing to be so great right now either.

ngzax:
https://github.com/urbit/bridge/pull/607

tomholford (approved):

PR #608, opened 2021-08-16
dx: add `pilot-mainnet` script

tomholford:
also, address linter issue

### DIFF
--- a/src/Bridge.js
+++ b/src/Bridge.js
@@ -21,9 +21,9 @@ import GlobalErrorBoundary from 'components/GlobalErrorBoundary';
 
 const INITIAL_NETWORK_TYPE = isRopsten
   ? NETWORK_TYPES.ROPSTEN
-  : isMainnet
-  ? NETWORK_TYPES.MAINNET
-  : NETWORK_TYPES.LOCAL;
+  : (isDevelopment && !isMainnet)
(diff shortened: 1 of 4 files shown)


src/Bridge.js
```diff
@@ -21,9 +21,9 @@ import GlobalErrorBoundary from 'components/GlobalErrorBoundary';
 
 const INITIAL_NETWORK_TYPE = isRopsten
   ? NETWORK_TYPES.ROPSTEN
-  : isDevelopment
-  ? NETWORK_TYPES.LOCAL
-  : NETWORK_TYPES.MAINNET;
+  : isMainnet
+  ? NETWORK_TYPES.MAINNET
+  : NETWORK_TYPES.LOCAL;
```
Fang-, line 26: I don't think we want to disregard the `isDevelopment` flag here, right? Default behavior for production builds should still be mainnet.

```suggestion
  : (isDevelopment && !isMainnet)
  ? NETWORK_TYPES.LOCAL
  : NETWORK_TYPES.MAINNET;
```

tomholford: Good catch, LGTM

Fang- (approved):
Nice! Thank you.

ngzax:
FTR, I have confirmed that this works for me. Thanks!

Issue #581, opened 2021-06-21
Ledger: doesn't work in Chrome


Fang-:
On latest firmware (2.0.0). Firefox is fine, but Chrome doesn't work for some reason. Kind of the inverse of what is described [here](https://github.com/LedgerHQ/ledgerjs/blob/master/docs/migrate_webusb.md), which made me wonder if Chrome had dropped U2F support or something, but I can't find anything about that...

For readers looking for a workaround: you can use Firefox, or add your Ledger to Metamask and then login with that instead.

eriknewton:
Just discovered that it no longer works in Safari either.

Also, while it works in Firefox, there's an additional issue using that browser: clicking on a given star in a list of stars brings up the wrong one.

tomholford:
>which made me wonder if Chrome had dropped U2F support or something

Also having trouble finding an official source, but this appears to be the case based on these related links:

https://github.com/near/near-wallet/issues/1775
https://www.reddit.com/r/ledgerwallet/comments/nqq84s/ledger_nano_s_wont_connect_to_browser_applications/

Here's an example of supporting the new standards (HID, WebUSB) while still being able to fall back to legacy U2F for now:
https://github.com/vacuumlabs/adalite/pull/1007/files

ngzax:
sorry to pile on, but even in FF it looks like Ledger login has issues. I tried to accept a planet in FF to my ledger wallet via the ledger login and received this error:

<img width="719" alt="Screen Shot 2021-08-17 at 8 13 58 AM" src="https://user-images.githubusercontent.com/87197/129735399-55975374-65d7-4350-8def-969b30154c27.png">

I subsequently connected to my ledger via the Metamask "bridge" method in Brave and the tx was processed fine.

nerveharp:
Known issue, will be addressed in the next version of Bridge. Ledger themselves recommend either using Firefox or connecting your Ledger device through Metamask.

PR #610, opened 2021-08-17
ledger, metamask: signing fixes

Fang-:
In the Metamask case, it seems we simply weren't calling the getAuthToken function correctly. Took the opportunity to also heed some of the deprecation warnings it was printing.

In the Ledger case, in addition to needing BNs instead of Buffers, latest ethereumjs-tx requires some care when doing EIP-155 style chain specification. Specifically, we can't do _both_ that and pass a `Common` object, they'll clash if we do.

Also updates the Ledger hw-app-eth to its latest version for good measure.

Fixes the thing @ngzax mentioned in https://github.com/urbit/bridge/issues/581#issuecomment-900305467.

Apologies for all the little issues here, I should've tested these things (especially the hw wallet ones) more thoroughly before approving recent changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2255,9 +2255,9 @@
       },
       "dependencies": {
         "@ethersproject/abstract-provider": {
-          "version": "5.4.0",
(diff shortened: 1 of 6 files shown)


tomholford (approved):
LGTM, nice work!

PR #614, opened 2021-08-18
api: useRoller

tomholford:
# Context

This adds a Roller API hook we can consume in future Roller development work.

# Changes

- use local copy of `roller-rpc-client` for now, until it is published to `npm` repo (Will, we may need to both put the `roller-rpc-client` package in the same dir on our machines for `package.json` to resolve it)
- add a stub `L2Header` to validate Roller API functionality by loading the config


package.json
```diff
@@ -69,6 +69,7 @@
     "react-final-form-arrays": "^3.1.0",
     "react-scripts": "4.0.3",
     "react-teleporter": "^2.2.0",
+    "roller-rpc-client": "file:~/dev/urbit/roller-rpc-client/client/typescript",
```
willbach, line 72: this can be updated to `"@urbit/roller-api": "0.1.2",`

willbach (commented):
Just the `package.json` thing, although that can be later if necessary

PR #615, opened 2021-08-18
txn: generalize fake signed transactions

Fang-:
Not all wallets support transaction signing as separate from sending.
Metamask was one such case, we handled it through special logic.
With WalletConnect, we might connect to a wallet that doesn't support
transaction signing either. As such, we generalize Metamask's special
logic.

When the wallet doesn't support transaction signing, we pretend to
succeed, but produce a "fake signed transaction" instead. This has
stubbed out serialization logic, and contains the original transaction
data plus a callback for sending that transaction.

In these cases, the user is informed that their transaction will be
signed by their wallet when sending. The signed transaction fold-out
is not shown.

image

Worth noting that, for the WalletConnect case, we cannot tell whether
the connected wallet supports transaction signing or not. So we simply
try to do it, and if it fails with METHOD_NOT_SUPPORTED, we know what's
up and proceed with the "fake signed" logic accordingly.

Additionally, this PR patches the auth token logic for the WalletConnect case to simply proceed if we cannot sign a message for reasons similar to the above.

This makes it so that Gnosis Safe can be used through WalletConnect, though it should be noted that the Bridge-side UX for sending a transaction using that is "sending progress bar forever". I don't think there's anything we can do about that.

@tomholford I tried adding in types as I went, but I couldn't get everything quite right. There's some TODOs strewn about relating to this, please advise.

Need to do another full testing pass over this once changes have settled, but so far seems to work alright. Only thing I haven't touched yet is reticketing, which is a bit special in that it does many sequential transactions. Might just work as-is, but worth verifying.

@tomholford If you could please test with some of the other WalletConnect wallets that you had success with in the past, that would be much appreciated.

### DIFF
--- a/src/lib/useWalletConnect.ts
+++ b/src/lib/useWalletConnect.ts
@@ -3,7 +3,7 @@ import QRCodeModal from '@walletconnect/qrcode-modal';
 import { ITxData } from '@walletconnect/types';
 
 import { useEffect, useMemo, useState } from 'react';
-import { Just } from 'folktale/maybe';
+import { Just, Nothing } from 'folktale/maybe';
 
 import { useWallet } from 'store/wallet';
 import { getAuthToken } from './authToken';
@@ -81,17 +81,30 @@ export const useWalletConnect = () => {
       return;
     }
 
-    const authToken = await getAuthToken({
-      address,
-      connector,
-      walletType: WALLET_TYPES.WALLET_CONNECT,
-    });
+    let authToken = Nothing();
+    try {
+      const token = await getAuthToken({
+        address,
+        connector,
+        walletType: WALLET_TYPES.WALLET_CONNECT,
+      });
+      authToken = Just(token);
+    } catch (e) {
+      if (e.message === 'METHOD_NOT_SUPPORTED') {
+        console.warn(
+          'wallet does not support message signing. proceeding without auth token.'
+        );
+      } else {
+        throw e;
+        //TODO  should errors with this *really* prevent login?
+      }
+    }
 
     const wallet: WalletConnectWallet = {
       address,
     };
 
-    setAuthToken(Just(authToken));
+    setAuthToken(authToken);
     setWallet(Just(wallet));
   };
 
--- a/src/lib/useWalletConnect.ts
+++ b/src/lib/useWalletConnect.ts
@@ -152,6 +152,37 @@ export const useWalletConnect = () => {
           data,
           nonce,
         })
+        .then((signature: string) => resolve(signature))
+        .catch((error: Error) => reject(error));
+    });
+  };
+
+  const sendTransaction = async ({
+    from,
+    to,
+    gas,
+    gasPrice,
+    value,
+    data,
+    nonce,
+  }: ITxData): Promise<string> => {
+    return new Promise((resolve, reject) => {
+      if (!connector || !isConnected()) {
+        reject(new Error('No connected wallet available for signing'));
+        return;
+      }
+
+      //REVIEW  .then path untested
+      return connector
+        .sendTransaction({
+          from,
+          to,
+          gas,
+          gasPrice,
+          value,
+          data,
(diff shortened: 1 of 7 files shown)


src/lib/useWalletConnect.ts
```diff
@@ -81,17 +81,30 @@ export const useWalletConnect = () => {
...
+      });
+      authToken = Just(token);
+    } catch (e) {
+      if (e.message === 'METHOD_NOT_SUPPORTED') {
+        console.warn(
+          'wallet does not support message signing. proceeding without auth token.'
+        );
+      } else {
+        throw e;
+        //TODO  should errors with this *really* prevent login?
```
tomholford, line 99: This is a good point, do we actually even need the auth token for any connected WC wallet? If the user is pointing their device at a QR code, or pasting the WC URL into another wallet, is it safe to assume that they also want to "authenticate"? In other words, by connecting, one is also implicitly authenticating, before actually clicking Authenticate.

To avoid refactoring, maybe we can keep the existing Authenticate flow for wallets that do support `signMessage`, and when hitting this error branch, just login anyways. It's actually a nice feature - one less popup prompt on the connected peer :)

Fang-: > do we actually even need the auth token for any connected WC wallet?

Reminder that we actually get these auth tokens for a reason!

There's a couple places we want to be able to deterministically derive new secrets. When setting networking keys, or generating invite codes for example. For custodial wallets, where we have access to a seed or private key, we can just derive from that. For non-custodial wallets, we do not have access to any secrets, so we cannot create any new (deterministic) secrets. This is why we sign an "auth token" on login: it gives us a secret to work with.

tldr: we can't just toss is out without facing ux consequences.

> maybe we can keep the existing Authenticate flow for wallets that do support signMessage, and when hitting this error branch, just login anyways.

Isn't this what we do as of e1b0301? We catch the "not supported" case and login anyway. It's _other_ errors, unexpected ones, that I'm wondering about here. For example, if we're dumb and mess up the function call, or their wallet implements it poorly, and we get a "invalid argument" error or w/e, should that prevent login? We probably don't want to _silently_ proceed, because it implies changed UX as per the above...

tomholford: Thanks for the explanation, this is much clearer now.

In this error case, since it could have failed for a variety of known and unknown reasons, and since we need an auth token for deterministically deriving new secrets, perhaps this `throw` is actually the best choice for now - it will render the message asking the user to file an issue, and then we can eventually add support for the failing WalletConnect peers.

tomholford:
@Fang- Following up on our out-of-band conversation, pulled the latest changes and re-tested with the following flows:

- Issued a Point
- Set Networking Keys
- Reset Networking Keys
- Reticket

| login method |  ropsten? | WC connect? | WC authenticate? | sign txn? | send txn? | network keys? | reticket? |
|---|---|---|---|---|---|---|---|
| mnemonic | ✅ | n/a | n/a | ✅ | ✅ | ✅ | ✅ |
| TokenPocket (Android) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| TokenPocket (iOS) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ⛔️ |
| Trust (Android) | ⛔️ | ✅ | ✅ | ✅ | ❔ | ❔ | ❔ |
| Trust (iOS) | ⛔️ | ✅ | ✅ | ✅ | ❔ | ❔ | ❔ |

Not sure if this is on us or our good friend TokenPocket: note that all flows work on Android, but iOS fails on Reticketing (the TXns all appear to be signed and sent, but keep receiving the 'nonce too low' error). Will see if I can find another Ropsten-capable wallet to test with to see if there's a pattern. Also, should be receiving HW wallets in the mail soon, might be able to add those to the testing matrix as well.

tomholford (approved):

PR #617, opened 2021-08-24
A star sends an invite

willbach:


src/components/L2/Card.tsx
```diff
@@ -0,0 +1,33 @@
...
+}: CardProps) {
+  return (
+    <Box className="card" onClick={onClick}>
+      {(!!icon || !!title) && (
+        <Row>
+          {!!icon && icon}
+          {!!title && <div className="card-title">{title}</div>}
+        </Row>
+      )}
+      {!!subtitle && <div className="card-subtitle">{subtitle}</div>}
```
tomholford, line 29: Functionally it shouldn't make a difference, but given the expected values defined by the `CardProps` it appears that `!!` is extraneous here. This is certainly a nitpick, so optional to act on it, but a change would improve the readability of this section.

src/components/L2/Dropdowns/AccountsDropdown.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const canBitcoin = Just.hasInstance(walletInfo.urbitWallet);
+
+  const points =
+    controlledPoints?.value?.value?.ownedPoints?.map((point: string) =>
+      Number(point)
+    ) || [];
+
+  const currentAddress = walletInfo?.wallet?.value?.address || '';
+  const addressLeadingFive = currentAddress.slice(0, 5);
+  const displayAddress = `${addressLeadingFive}...${currentAddress.slice(-4)}`;
```
tomholford, line 34: If it suits the needs of this use case, how about swapping this out for `abbreviateAddress` in `src/lib/utils/address.ts`? Otherwise, this is fine

src/components/L2/Dropdowns/Dropdown.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { Icon } from '@tlon/indigo-react';
+
+import './Dropdown.scss';
+
+export interface DropdownProps {
+  className: string;
+  children: any;
```
tomholford, line 8: Per [this reference](https://github.com/typescript-cheatsheets/react#useful-react-prop-type-examples) for React + Typescript, it's recommended to use `React.ReactNode` for `children` type declarations for improved type-checking

src/components/L2/Dropdowns/FeeDropdown.tsx
```diff
@@ -0,0 +1,54 @@
...
+    },
+    {
+      amount: 0.0002,
+      display: '0.0002 ETH (15 min)',
+    },
+    {
+      amount: 0.0002,
+      display: '0.0001 ETH (1 hour)',
+    },
+  ];
```
tomholford, line 32: Are these the values we will be using in production? Or is there an API where we can load updated gas prices?

src/components/L2/Headers/L2BackHeader.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const goBack = useCallback(() => {
+    if (back) {
+      back();
+    } else {
+      pop();
+    }
+  }, [back, pop]);
+
+  useEffect(() => {
+    console.log('loaded config in L2BackHeader:', config);
```
tomholford, line 37: Let's either log this only in development / tmp environment, or remove completely

src/components/L2/Headers/L2PointHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+const L2PointHeader = ({
+  hideTimer = false,
+  hideInvites = false,
+  numInvites = 0,
+}: L2PointHeaderProps) => {
+  const { config } = useRoller();
+  const { nextRoll, currentL2, pendingTransactions } = useRollerStore();
+
+  useEffect(() => {
+    console.log('loaded config in L2PointHeader:', config);
```
tomholford, line 28: Same comment as in `L2BackHeader` re: logging debug info

src/components/L2/Window/BodyPane.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+import './BodyPane.scss';
+
+export interface BodyPaneProps {}
+
+const BodyPane = ({ children }: any) => {
```
tomholford, line 7: Looks like this type got lost in the shuffle

src/lib/useInviter.js
```diff
@@ -193,6 +193,8 @@ const useInviter = () => {
         };
       }
 
+      console.log('INVITES', confirmedInvites);
+      // TODO: store these locally, and also handle the situation
```
tomholford, line 197: Has this TODO been addressed?

src/lib/useRoller.ts
```diff
@@ -43,7 +108,244 @@ export default function useRoller() {
...
+        const planet = planets[i];
+
+        const { ticket, owner } = await wg.generateTemporaryDeterministicWallet(
+          planet,
+          _authToken
+        );
+
+        const from = {
+          ship: _point, //ship that is spawning the planet
+          proxy, // TODO: check that this is either "own" or "spawn"
```
tomholford, line 146: Has this TODO been resolved?

src/views/IssueChild.js
```diff
@@ -43,6 +43,7 @@ function useIssueChild() {
   return useEthereumTransaction(
     useCallback(
       (spawnedPoint, owner) => {
+        console.log(5, spawnedPoint, owner);
```
tomholford, line 46:
```suggestion
```

src/lib/useRoller.ts
```diff
@@ -43,7 +108,244 @@ export default function useRoller() {
...
+              (p: number) =>
+                azimuth.azimuth.getPointSize(p) ===
+                  azimuth.azimuth.PointSize.Planet &&
+                availablePoints.includes(p)
+            );
+          }
+
+          // Iterate over all spawned and controlled planets
+          // If the planet is not in available invites, generate the ticket and add it
+          console.log('POSSIBLE MISSING', possibleMissingInvites);
```
tomholford, line 293: These log statements seem useful for debugging, perhaps we should keep these behind a flag for future development?

src/views/Invite/Cohort.js
```diff
@@ -1,398 +1,397 @@
-import React, { useMemo, useEffect, useCallback, useState } from 'react';
-import { Grid, Flex, Button } from 'indigo-react';
-import * as ob from 'urbit-ob';
-import { Just, Nothing } from 'folktale/maybe';
+import React, { useEffect, useCallback, useState } from 'react';
```
tomholford, line 1: This file you inherited is rather large, is it possible to refactor some of the conditional branches (e.g., the `hasPending` or `networkKeysNotSet` paths) and / or the component itself into smaller submodules? Would make this much more manageable for future reviewers and maintainers. Will defer to you on whether or not to address this now, it might be more trouble than its worth.

src/views/TransactionHistory.scss
```diff
@@ -0,0 +1,5 @@
+@import 'style/shared/variables.scss';
+
+.transaction-history {
```
tomholford, line 3: Is this still needed?

src/views/Points.js
```diff
@@ -96,24 +137,30 @@ export default function Points() {
...
+    [getDetails, controlledPoints]
+  );
+
+  const maybeLockedPoints = useMemo(
     () =>
       controlledPoints.chain(points =>
         points.matchWith({
           Error: () => Nothing(),
           Ok: c => {
+            // TODO: how to determine this?
```
tomholford, line 151: Has this TODO been resolved?

src/views/Points.js
```diff
@@ -36,26 +37,66 @@ const maybeGetResult = (obj, key, defaultValue) =>
...
+          }
+        },
+      })
+    )
+    .getOrElse([]);
+};
+
+export const isLocked = details =>
+  details.owner === '0x86cd9cd0992f04231751e3761de45cecea5d1801' ||
+  details.owner === '0x8c241098c3d3498fe1261421633fd57986d74aea';
```
tomholford, line 70: Please move these to `constants`

src/views/Points.js
```diff
@@ -349,20 +369,25 @@ export default function Points() {
...
           </Grid.Item>
         )}
 
         {allPoints.length > 0 && (
           <Grid.Item full as={Grid} gap={1}>
-            <Grid.Item full as={H5}>
+            {/* <Grid.Item full as={H5}>
               {pluralize(allPoints.length, 'ID')}
-            </Grid.Item>
+            </Grid.Item> */}
```
tomholford, line 380: Is this still needed?

src/views/Point.js
```diff
@@ -266,40 +272,34 @@ export default function Point() {
...
           full
           as={ForwardButton}
           disabled={!canBitcoin}
           className="mt1"
           detail="Bitcoin management"
           onClick={goBitcoin}>
           Bitcoin
-        </Grid.Item>
-        <Grid.Divider />
+        </Grid.Item> */}
```
tomholford, line 302: Is this still needed?

src/views/Point.js
```diff
@@ -266,40 +272,34 @@ export default function Point() {
...
-        {inviteButton}
-        <Grid.Item
-          full
-          as={ForwardButton}
-          disabled={!canManage}
-          onClick={goUrbitID}
-          className="mt1"
-          detail="Identity and security settings">
-          ID
+        {/* {inviteButton} */}
```
tomholford, line 275: Still needed?

tomholford (changes requested):
Nice work, Will, overall this looks good. Added a bunch of comments and questions in-line. Most are relatively minor, but there's at least one big question: are we going to use hardcoded gas prices in production?

Also, did not call this out inline since it's more of a general observation: in most of the view templates, the `Box` and `Row` components from `@tlon/indigo-react` are used, per the directive from Tlon's design team. However, there are a handful of instances with `div` instead of one of the aforementioned components. My understanding is we should prefer Box or Row when possible to maintain consistent stylings with the rest of Tlon & Urbit's design language.

src/components/L2/Card.tsx
willbach: I thought React had a problem with non-boolean values for conditional rendering

src/components/L2/Dropdowns/FeeDropdown.tsx
willbach: Right now I'm not using this component. I can remove these but generally yes, they should be coming from an API

src/lib/useInviter.js
willbach: yes, we don't actually use that anymore so it has been removed

src/views/Invite/Cohort.js
willbach: I agree. I might be able to do it tonight.

src/views/TransactionHistory.scss
willbach: yes, but it's still half-finished.

src/views/Points.js
willbach: yes

src/views/Points.js
willbach: nope

src/views/Point.js
willbach: nope

willbach:
@tomholford I'm going to remove the `FeeDropdown` component with the hardcoded gas prices because we won't be using it for awhile and it'll probably look different anyway. Currently gas prices for the `InlineEthereumTransaction` are coming from somewhere else and that works fine for now.

I'll replace `div` with `Box` wherever I'm using it. Otherwise I think I addressed everything going through it just now. New commit(s) incoming.

src/views/TransactionHistory.scss
tomholford: Okay NP

tomholford (approved):
LGTM, checked it out and verified locally. Nice work 💪

PR #621, opened 2021-08-29
Invite Acceptance Flow

tomholford:
# Preview

https://user-images.githubusercontent.com/16504501/133674423-dc80aeba-e12b-47b2-8f29-5b1cd75ae9cd.mov

# Context

This PR adds an all new Activation flow for new users to activate an Invite generated in Bridge.

# Changes

- New Activation flow built using `@tlon/indigo-react` component library

# Testing

Confirmed with local instance of Urbit binary running `naive/roller` branch:

1. Start the local environment as described in the latter section of `DEVELOPMENT.md`
2. Log in to Bridge, select ~wanzod
3. Click the envelope and generate some invites
4. After the next batch rolls, check the pending invites tab by logging out and in again and clicking the envelope
5. Use one of the links, format it like so, and visit to start the Accept Invite Flow:
`localhost:3000/#pitbep-salleb-nimrum-dinwer-dalnut-dovtux`
6. After a successful transfer of the point, trying to use the code again fails as expected


src/views/Invite/Cohort.js
```diff
@@ -154,6 +154,8 @@ export default function InviteCohort() {
       getPendingTransactions(point);
       pop();
     } catch (error) {
+      console.log(error);
+      // TODO: error is not used anywhere on the UI
```
yosoyubik, line 158: This is just as a reminder that we need to show any possible errors here in the UI

src/lib/useRoller.ts
```diff
@@ -225,22 +142,34 @@ export default function useRoller() {
         throw new Error('Internal Error: Missing Contracts/Web3/Wallet');
       }
 
+      if (Just.hasInstance(authMnemonic)) {
+        throw new Error(
+          "Auth key Error: A management mnemonic can't create invite codes"
+        );
+      }
```
yosoyubik, line 149: I'm capturing this here, but it could be done outside of the invite creation, for example when you click on the envelope—this depends on designs for when/where to show the error message, so tagging @jyng for feedback.

 It can be a more general question since I'm not sure what's the behavior for current Bridge when logging in with a management mnemonic and then trying to create some invites.

yosoyubik (approved):
Tested the invite flow creation/acceptance logic and all worked fine.

I left some comments as reminders of something not related to the invite acceptance flow that this PR is about, but worth noticing for creating some issues/tasks around that.

tomholford:
> Tested the invite flow creation/acceptance logic and all worked fine.
> 
> I left some comments as reminders of something not related to the invite acceptance flow that this PR is about, but worth noticing for creating some issues/tasks around that.

Thanks for the review, tracked the issue here: #633

Issue #596, opened 2021-07-21
Error: Returned error: The method personal_sign does not exist/is not available - MetaMask


rabsef-bicrym:
```
Error: Returned error: The method personal_sign does not exist/is not available
    at Object.ErrorResponse (errors.js:28)
    at index.js:303
    at XMLHttpRequest.i.onreadystatechange (index.js:98)
```

When using metamask 9.8.0 in Chrome 91.0.4472.164 on Windows 10 (corporate laptop so who knows what hijinks in configuration), I am unable to log in to Bridge 2.9.2 (hosted off urbit.org). I get the error message detailed above in the console but there are no other obvious indications of failure. One time of, say, 10 tries, I did get sent to an error page but all others fail without indication of failing (login button is light grey and unclickable but has -> arrow still).

Switching MetaMask to Ropsten then back to mainnet seemed to alleviate the issue.

@Fang- tagging as per request to post this issue in separate conversation.

Fang-:
> Switching MetaMask to Ropsten then back to mainnet seemed to alleviate the issue.

Wild. I ran into this issue today as well, and switching between connected accounts was what got it to work for me. Once it works, it works, but unclear to me why it just doesn't sometimes.

eglaysher:
I'm getting this error too, but switching between Ropsten/Mainnet and between accounts doesn't fix things.

Fang-:
@eglaysher So strange. Just had the same I described above: reproduction, but it resolving after fiddling a bit. Are you using a hardware wallet account, or just a plain metamask-generated one?

We might want to catch this specific error and proceed with login anyway if we get it, similar to the change made in e1b0301.

sjwinther:
Consistently getting this error for the past many days when trying to login with MetaMask. Changing back and forth between networks doesn't seem to help. Not really sure how to proceed?

Fang-:
Alright all, I just pushed Bridge v2.10.4, which includes an attempt at a fix for this. Please let me know if Metamask login works (more consistently) for you now.

sjwinther:
Works like a charm! Thanks 🪐

OdysLam:
When sending a transaction to accept a point, I am getting this error with metamask: 
`inpage.js:1 MetaMask - RPC Error: Error: [object Object] Object`

Fang-:
[...]

PR #630, opened 2021-09-16
metamask: strange personal_sign ritual

Fang-:
This is an attempt to do something about #596.

I can reproduce that issue fairly consistently by just pressing "cancel" when Metamask pops up for the signing request and reloading the page. Eventually it gets into the bad state where pressing the Metamask login button doesn't actually prompt you for signing the message, instead it just stays grey and gives the "personal_sign does not exist" in the browser console.

For some reason, calling `personal_sign` through `window.ethereum` this way is highly reliable, and doesn't let me reproduce using the above steps. Is it because of `window.ethereum`? Is it because we manually construct the request? We just don't know..[.](http://img.photobucket.com/albums/v695/carnilia/randomgifs/tumblr_m7swmtmv7n1rn95k2o1_250_zps880b4a1e.gif)

Notably, in `Login/Metamask.js`, we already use `window.ethereum` directly (with `.send`, which is deprecated and should be replaced by `.request`) for requesting the accounts.

The `if (window.ethereum)` is kind of superfluous in this case, because we _should_ only call this when we _do_ have it, but I want to get this live quickly so I'm opting for the extra safety.

(For the record, we should do something about Mm login's error handling in general, but I just want this out into the hands of people quickly.)

### DIFF
--- a/src/lib/authToken.ts
+++ b/src/lib/authToken.ts
@@ -57,7 +57,19 @@ type GetAuthTokenArgs =
   | DefaultAuthTokenArgs;
 
 const getMetamaskAuthToken = ({ address, web3 }: MetamaskAuthTokenArgs) => {
-  return web3.eth.personal.sign(MESSAGE, address, '');
+  if (window.ethereum) {
+    //NOTE  this doesn't _seem_ to be affected by #596,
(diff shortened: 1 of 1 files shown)


tomholford (approved):
LGTM, hopefully this helps address the immediate pain point for users

PR #636, opened 2021-09-27
Logic for setting spawn/management proxy on L2

yosoyubik:
I've changed the ganache migration script so ~marzod is on L2, so we can test changing it's management proxy (we can't do that with ~wanzod because its proxy is on L2), but to test it manually without recreating the local node, one of the other stars need to go through the migration flow.

This only handles the logic for sending the requests to the Roller, but the UI/CSS doesn't reflect the designs ([SetProxy.scss](https://github.com/urbit/bridge/pull/636/files?file-filters%5B%5D=#diff-b1027d5c9ce5ce7eecf48e5638e3d87da562bbdd8e25c92b98702f2dc62ca8bf) is just a placeholder)


src/views/UrbitID/SetProxy.js
```diff
@@ -113,7 +129,30 @@ export default function SetProxy() {
...
+  const setProxy = useCallback(async () => {
+    // setLoading(true);
+    console.log(newAddress);
+    if (newAddress === '') return;
+    try {
+      await setProxyAddress(getL2ProxyName(data.proxyType), newAddress);
+      getPendingTransactions(_point);
+      pop();
+    } catch (error) {
+      // setError(error);
```
tomholford, line 141: Do we still need the catch / finally? We an also address this in a follow-up PR

tomholford (approved):
LGTM

src/views/UrbitID/SetProxy.js
yosoyubik: Yes, let's do that when we tackle the new UI

PR #640, opened 2021-09-29
Add logic to support setting networking keys, transferring, and re-ticketing on L2

yosoyubik:
This only adds the logic to support the correct calls to the Roller, the new UI/UX should be addressed in a separate PR


src/lib/useRoller.ts
```diff
@@ -161,69 +155,76 @@ export default function useRoller() {
...
       const requests: Promise<string>[] = [];
 
-      const operator =
-        getSpawnNonce(starInfo, _wallet.address) ||
-        getOwnerNonce(starInfo, _wallet.address);
+      const proxy = isOwnerProxy(starInfo.ownership!, _wallet.address)
+        ? 'own'
+        : isSpawnProxy(starInfo.ownership!, _wallet.address)
+        ? 'spawn'
+        : undefined;
```
tomholford, line 172: Looks like this logic is repeated in multiple places. That is probably fine for now, but if we anticipate using it in more places, we should consider extracting it to its own helper.

src/lib/useRoller.ts
```diff
@@ -412,112 +410,150 @@ export default function useRoller() {
...
+    const _wallet = wallet.getOrElse(null);
+    const _details = getDetails(_point);
+    if (!_wallet || !_details) {
+      // not using need because we want a custom error
+      throw new Error('Internal Error: Missing Wallet/Details');
+    }
 
-  // }
+    const pointDetails = await api.getPoint(_point);
+    // const proxy = getProxy(pointDetails.ownership!, _wallet.address);
```
tomholford, line 432:
```suggestion
```

src/lib/useRoller.ts
```diff
@@ -161,69 +155,76 @@ export default function useRoller() {
...
+        : isSpawnProxy(starInfo.ownership!, _wallet.address)
+        ? 'spawn'
+        : undefined;
 
-      if (operator === undefined)
+      if (proxy === undefined)
         throw new Error("Error: Address doesn't match proxy");
-      console.log('operator', operator);
+
+      const nonce = await api.getNonce({
```
tomholford, line 177: I think this is a great design decision to push the nonce management to the backend. simplifies the frontend significantly

src/lib/useRoller.ts
yosoyubik: Makes sense, I can make wrappers around these calls

src/views/UrbitID/ResetKeys/ResetExecute.js
```diff
@@ -92,20 +96,33 @@ export default function ResetExecute({ newWallet, setNewWallet }) {
...
-          contracts: need.contracts(contracts),
-          networkType,
-          onUpdate: handleUpdate,
-          nextRevision: networkRevision + 1,
-          txnSigner,
-          txnSender,
-        });
+        if (currentL2) {
+          console.log(newWallet.value);
+          // FIXME: the useEffect is called twice, which is fine since the duplicate
```
tomholford, line 101: Since this is not a showstopper, we can ignore it for now, and I can take a look at this in a follow up PR

tomholford (approved):
LGTM, a few minor comments that don't necessarily require changes, so feel free to merge

package.json
```diff
@@ -32,7 +32,7 @@
     "@types/styled-components": "^5.1.10",
     "@types/styled-system": "^5.1.11",
     "@types/styled-system__css": "^5.0.15",
-    "@urbit/roller-api": "^0.1.8",
+    "@urbit/roller-api": "^0.1.9",
```
tomholford, line 35: Since this was addressed in #641, does this branch need to be rebased on `roller`?

src/lib/utils/roller.ts
```diff
@@ -73,19 +72,19 @@ export const spawn = async (
...
-  api: any,
-  _wallet: any,
-  _point: number,
+  api: RollerRPCAPI,
+  wallet: any,
+  point: Ship,
   proxy: string,
   nonce: number,
-  urbitWallet: any
+  networkSeed: any,
```
tomholford, line 80: Is there a better choice than `any` in this case?

package.json
yosoyubik: I'll look locally if there's any conflicts, although the merge message here says that all is fine

src/lib/utils/roller.ts
yosoyubik: I think we can use string here, from the `urbit-key-generation` repo [this](https://github.com/urbit/urbit-key-generation/blob/fa656d1d982681ea10c6647ec0b878e0b9309c07/src/index.js#L273) should be a hex-encoded string

yosoyubik: ...although this is a Maybe (Just/Nothing) from folktale, and that doesn't seem to support typescript? https://github.com/origamitower/folktale/issues/65

Issue #646, opened 2021-10-05
Metamask gives "eth_sendTransaction does not exist"


philipcmonk:
Have two reports that "eth_sendTransaction does not exist or is not available" shows up when you try to use metamask to set your networking keys.

Chrome on Linux

philipcmonk:
Maybe caused by #630

PR #647, opened 2021-10-05
metamask: use injected provider for sending txs

Fang-:
Similar to #630, we opt into using the injected provider directly
instead of going through Web3. While reproduction is difficult for this
case, this kind of change worked for the other one, so we go ahead and
apply it here. Certainly can't hurt us more than the bug itself already
does.

Hopefully does something about #646, but since I couldn't get a repro, no way to tell for sure.


tomholford (approved):
LGTM, hopefully this resolves the issue!

PR #696, opened 2021-11-02
UI and UX updates for the Point view and sub views

willbach:


src/lib/utils/invite.ts
```diff
@@ -2,7 +2,19 @@ import * as ob from 'urbit-ob';
...
+  } else if (planet) {
+    return ob.patp(planet);
+  }
+
+  return 'No Ticket or Ship';
+};
+
+export const generateUrlAbbreviation = (ticket: string, planet: number) => {
+  if (ticket && planet) {
+    return `bridge.urbit.org/#${ticket.slice(1, 7)}...${ob
```
tomholford, line 15: Optionally, would be nice to use the current host in the URL, especially if we also plan to deploy an instance to a mainnet staging environment.

```suggestion
    return `${location.host}/#${ticket.slice(1, 7)}...${ob
```

tomholford (approved):
One optional suggestion, otherwise LGTM ✅  Nice work!

src/lib/utils/invite.ts
willbach: Good call, will put this in

PR #753, opened 2021-11-09
point-view: animate Passport based on selected point

tomholford:
# Change

Previously this used the eth address associated with the point to generate the passport; now it uses the actual azimuth point integer value as the seed. Thus, switching points in the dropdown will update the Passport with a unique, mostly-deterministic animation.

# Preview

https://user-images.githubusercontent.com/16504501/140860093-304656d1-6dc2-4087-a422-fc0d2bcc89cf.mov

### DIFF
--- a/src/components/Passport.js
+++ b/src/components/Passport.js
@@ -48,13 +48,15 @@ function Passport({
 
   const patp = ob.patp(point.value);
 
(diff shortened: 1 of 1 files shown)


jyng:
this looks great, thanks for getting to this! lgtm

PR #754, opened 2021-11-09
tx-view: update TX history event labels & icons

tomholford:
# Change

This updates the placeholder TX event labels with clearer copy for end users, and adds updated Indigo icons.


jyng:
lgtm

PR #773, opened 2021-11-17
Complete Points & Invites Refactor for L2

willbach:
Some of the highlights:

- New Point object that has all of the point details on it
- Invites are stored by point and are checked & retrieved async
- Wallet generation is done in a web worker and is non-blocking
- Refactored the L2 calls to use a common function
- Added a "Home" button to the Point view
- Updated logic for which ID and OS fields are displayed and editable
- Galaxies can set spawn proxy to L2


src/components/L2/Dropdowns/AccountsDropdown.tsx
```diff
@@ -146,7 +133,7 @@ const AccountsDropdown = ({ showMigrate = false }: AccountsDropdownProps) => {
         <Box>Logout</Box>
         <Icon icon="LogOut" />
       </Row>
-      <Modal show={showModal} hide={() => setShowModal(false)}>
+      {/* <Modal show={showModal} hide={() => setShowModal(false)}>
```
tomholford, line 136: Okay to remove this?

tomholford (approved):
LGTM

src/components/L2/Dropdowns/AccountsDropdown.tsx
willbach: yup, removed

PR #778, opened 2021-11-22
Update login and point transfer screens

willbach:
Lots of updates in here. Mainly to login screen and the accept & cancel transfer screens.


tomholford (approved):
LGTM

Issue #792, opened 2021-11-30
Linear Star Release needs to be redesigned


jyng:
### Locked Stars — Empty

Quite often, Star Lockups are being sent to new addresses, so we should have a proper empty state here. Also, we've also never had a proper empty state here. Because they don't have any ships here, we should remove dropdown functionality and just bring the copy button out next to the address. 

![Family(Locked Stars) - Empty](https://user-images.githubusercontent.com/9557907/144106961-5c7cc107-7490-4134-adb3-f3b57a7af374.jpg)

### Family - Empty
If they don't have any locked stars, it would look the same but without the locked stars navigation button.

![Family-Empty](https://user-images.githubusercontent.com/9557907/144106946-c426cb09-0cc3-4f0d-bb6c-65fe90433c1b.jpg)

### Family w/ Stars
If they have Locked Stars, the navigation button should be sticky (just like how it is in current bridge)

![Family w:Locked Stars](https://user-images.githubusercontent.com/9557907/144107371-1a238355-1562-45c0-81e3-63880ad74807.jpg)

### Locked Stars - Empty
 
![Locked Stars - Empty](https://user-images.githubusercontent.com/9557907/144111025-518f4343-564a-4811-8378-201651276a9d.jpg)

### Locked Stars - Withdraw

- In this new design, I've removed 'Active' because that's the same as what's viewable on the family view. Instead we have a dropdown that just allows them to toggle between "Withdraw" and "Delegate".
- We're giving more visual priority to what the user can withdraw at the given time, but also displaying the number they have in lockup. Lockup mechanism can't be concisely understood on this screen, so an `info` icon should allow the user to learn about how they're designed.
- Input for Address should have a secondary "Use My Address" for convenience.

![Locked Stars - Filled](https://user-images.githubusercontent.com/9557907/144109830-73526e00-66ee-4930-bbfe-7b0f1c7380d7.jpg)

tomholford:
>. Lockup mechanism can't be concisely understood on this screen, so an info icon should allow the user to learn about how they're designed.

@jyng @nerveharp A couple questions on this:
- What copy should be used to explain the lockup mechanism?
- How should it be displayed? e.g, a modal or tooltip?

jyng:
@tomholford  Hovering over the `Info` Icon should display this tooltip.
![Locked Stars - Info Hover](https://user-images.githubusercontent.com/9557907/145908714-f70b8cc9-1a4d-4c88-907d-c7e736a9f806.jpg)

PR #832, opened 2021-12-08
points-view: empty results placeholder

tomholford:
# Changes

This PR styles the Points view to show a placeholder box when no Points are available to the connected account.

# Preview

image
image


tomholford:
@jyng FYI [the spec](https://github.com/urbit/bridge/issues/792) calls for the Accounts Dropdown to be disabled; however, while testing I realized that the user would have no option to logout gracefully, so it is still enabled in this empty state.

willbach (approved):
looks good

PR #836, opened 2021-12-09
Changed button state when insufficient ETH

willbach:
The screen:

Screen Shot 2021-12-09 at 1 27 51 PM

@jyng @tomholford @nerveharp


tomholford (approved):
LGTM

Issue #848, opened 2021-12-15
L2 planet: claim -> download keyfile -> boot -> error


nerveharp:
Booting a freshly-claimed planet via URL invite on the command line results in a pre-boot %key-mismatch with no explanation. The only recourse is logging in to Bridge, checking various countdown timers, and trying again. (See screenshots / stdout.)

This was an L2 planet freshly spawned from `~sordem` on 2021-12-14.

**To reproduce:**
1. Claim a planet via invite code.
2. Download the keyfile/passport.
3. Unzip the passport and boot Urbit using the keyfile.
4. Observe %key-mismatch.
5. Log in to Bridge at https://bridge-staging.urbit.org/
6. See yellow countdown timers.
7. Wait for however long the timers say to.
8. Try booting again, get same error.
9. Check History, wait until however long _that_ timer says to.
10. Try booting again, see same %key-mismatch error.

**Screenshots/artifacts:**

stdout:
```
~
urbit 1.8
boot: home is /home/james/Downloads/rigteg-havtuc
loom: mapped 2048MB
lite: arvo formula 27e494c5
lite: core 7b144622
lite: final state 7b144622
boot: downloading pill https://bootstrap.urbit.org/urbit-v1.8.pill
boot: retrieving ~rigteg-havtuc's public keys
boot: verifying keys
boot: invalid keys for planet '~rigteg-havtuc'
pre-boot error: %key-mismatch
```

After logging in:
![2021-12-14_18-48](https://user-images.githubusercontent.com/748181/146097583-f29a1cde-faaf-4a26-80d8-cc9a789b6df7.png)

After watching the timers clear and trying again:
![image](https://user-images.githubusercontent.com/748181/146098666-e16183fe-73f0-4d4f-b9a0-3d095037b6c0.png)

**Ethereum network:**
- Mainnet

**Desktop (please complete the following information):**
 - OS: Linux 5.15.7-arch1-1 (64 bit)
 
**Login method (please complete the following information):**
- Master ticket, address 0x8e046d073694555aa8f65aa06aea93319cb54b4d

nerveharp:
Adding context from `~wolref-podlex/next-pki`:

```
~wicdev-wisryt 12:41 AM
yes, it looks like bridge gives the wrong keyfile during activation

[...]

~datder-sonnet 2:01 AM
[...]

PR #849, opened 2021-12-15
network-keys: generated keyfiles use multikey format

tomholford:
# Changes
- [x] refactor Activate flow to unify keyfile generation and download behavior
- [x] refactor `lib/keys` to support multikey format

# Example Key
## Before
```
0w54.XImwa.fDTLw.gBfjF.zVHuz.PUp3g.dXxtu.7Ycw2.iJYNb.NepHA.0NJ6L.B9AQQ.pY45l.X61fS.5ygpk.RQfAG.m3TjJ.~zQWg.wf00U.NGu41
```

## After
```
0wa5FlF.MRsOU.x46cu.1N~9q.5XQjk.LH7ye.cbD81.B9lYo.rzDLb.kgne7.5TFZd.dJCSY.Hfore.9WpoQ.HgDfE.zmqmC.MKi5r.y41o0.92NeX.5E2zV.ZXU49.jQWo-.qTEY-.6gQ3u.Unnx~.380AH.vciYj.CqV0c.rhHVi.pdd6v.11luN.wjZxo.A6ldt.3VaBw.ZQXvU.ZeA83.M0oIq.Dx0r5
```

# Testing
This was tested using `npm run pilot-ropsten`, the following flows "work on my machine":

- downloaded keyfile from Home view
- downloaded keyfile in Activate flow

### DIFF
--- a/src/lib/keys.js
+++ b/src/lib/keys.js
@@ -76,10 +76,34 @@ export const compileNetworkKey = (pair, point, revision) => {
     noun.Atom.fromInt(point),
     noun.Atom.fromInt(revision),
     noun.Atom.fromString(bnsec.toString()),
-    noun.Atom.yes
+    noun.Atom.fromInt(0)
   );
 
   return b64(jam(sed));
+};
+
+/**
+ * @param {number} point
+ * @param {array<{revision: number, pair: object}>} keys
+ * @return {string}
+ */
+export const compileMultiKey = (point, keys) => {
+  const kyz = keys.map(k => {
+    const bnsec = new BN(createRing(k.pair), 'hex');
+    return noun.dwim(
+      noun.Atom.fromInt(k.revision),
+      noun.Atom.fromString(bnsec.toString())
+    );
+  });
+  kyz.push(noun.Atom.fromInt(0));
+
+  const fed = noun.dwim(
+    noun.dwim(noun.Atom.fromInt(1), noun.Atom.fromInt(0)), // version
+    noun.Atom.fromInt(point), // ship
+    noun.dwim(kyz) // keys
+  );
+
+  return b64(jam(fed));
 };
 
 /**
--- a/src/lib/keys.js
+++ b/src/lib/keys.js
@@ -198,8 +198,8 @@ export const attemptNetworkSeedDerivation = async ({
 
 /**
  *
- * @param {object} pair
- * @param {object} details
+ * @param {object} pair - type NetworkKeys
+ * @param {object} details - type L1Point
  * @return {boolean}
  */
 export const keysMatchChain = (pair, details) => {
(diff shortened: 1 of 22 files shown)


src/lib/keys.js
```diff
@@ -222,3 +222,30 @@ export const segmentNetworkKey = hex => {
...
 };
+
+// ctsy @yosoyubik
+/**
+ * @param {object} pair
+ * @param {number} point
+ * @param {number} revision
+ * @return {string}
+ */
+export const compileMultikey = (point, pair1, pair2, revision1, revision2) => {
```
tomholford, line 233: @Fang- Please pay special attention to this!

src/lib/useKeyfileGenerator.ts
```diff
@@ -18,20 +18,35 @@ import {
 } from './keys';
```
tomholford, line 18: This is the legacy implementation, and is no longer consumed by the application. Keeping it for comparison purposes during the review. We can remove it as a final commit to this PR, or in a follow up.

src/lib/keys.js
tomholford: `[[1 0] ship [life1 key1] [life2 key2] 0]`

src/lib/keys.js
```diff
@@ -222,3 +222,30 @@ export const segmentNetworkKey = hex => {
...
+  const bnsec1 = new BN(createRing(pair1), 'hex');
+  const bnsec2 = new BN(createRing(pair2), 'hex');
+
+  const tag = noun.dwim(noun.Atom.fromString('1'), noun.Atom.yes);
```
yosoyubik, line 237: Looking at this now, I think it should be `noun.Atom.fromInt(1)` so:

`const tag = noun.dwim(noun.Atom.fromInt(1), noun.Atom.yes);`

yosoyubik (changes requested):

nerveharp:
Fixes urbit/bridge#848

willbach (approved):
looks good, nice job with the quick turnaround

src/lib/keys.js
Fang-: Maybe I should've re-communicate as things drew closer, but I had already done multikeyfile work in 9331dd1999.

This seems to behave identically, but my version supports any number of keys in the file. (Additionally, small stylistic nit, `noun.Atom.yes` is semantically misleading, `noun.Atom.fromInt(0)` is nicer. Really there should just be a `noun.Atom.null` alias.)

nerveharp:
Anything blocking a merge here?

yosoyubik:
I think this is still not working properly (@Fang- and @tomholford will have more details about it) but using both [this](https://github.com/urbit/bridge/pull/849/commits/fdf8e6e3488fe9763e549d0831c4250c3d1c773e) and [this](https://github.com/urbit/bridge/commit/9331dd1999d043acd3d03ff9ff66e41eb116ddfd) results in a boot failure

willbach (approved):
looks solid, good job on this

src/lib/useMultikeyFileGenerator.ts
```diff
@@ -80,7 +80,7 @@ export default function useMultikeyFileGenerator({
         ? Just(seed)
         : await attemptNetworkSeedDerivation({
             urbitWallet: seedWallet || urbitWallet,
-            wallet,
+            wallet: wallet,
```
willbach, line 83: this can be shortened to just `wallet`, but also, do we need to pass this param in if `seedWallet` will always exist? We might not need `urbitWallet` either

tomholford: at this point, `seedWallet` is only passed in during the activation flow. we can refactor this to make it a required argument; for now, will just address the duplicate naming

PR #855, opened 2021-12-17
npm: add build mainnet

yosoyubik:


willbach (commented):
I think the default build should be for mainnet

yosoyubik:
> I think the default build should be for mainnet

Fixed here: https://github.com/urbit/bridge/pull/855/commits/6c9171bbe37ecc9794a7a956b7dc7df16153c75b

willbach (approved):
Looks great! Thanks for the update

PR #858, opened 2021-12-17
Add explicit height to master key animation

willbach:
Screen Shot 2021-12-17 at 6 13 38 PM


willbach:
There was an issue with the display on Safari where the animated svg didn't show up at all. Adding an explicit height fixed it

PR #877, opened 2022-01-14
keys: during Activation, use new wallet's seed

tomholford:
This fix adds conditional behavior to the MultiKeyFileGenerator.
By default, the keygen uses the logged-in wallet to generate both halves of the pair.

During the Activation flow, the generator now accepts an optional wallet arg which is used to pass in the receiving master ticket wallet. Then, the MT wallet is used to generate the 2nd half of the keypair.

This change also renames some variables for improved readability.


willbach (approved):
lgtm

tomholford:
Tested both with `npm run pilot-ropsten` and `npm run pilot-mainnet`. Confirming that i am able to boot with a key downloaded during the Activation flow.

PR #880, opened 2022-01-20
activate: new master ticket is non-deterministic

tomholford:
Previously, the new master ticket  was being generated deterministically from the invite ticket and point.
This resulted in the passport's wallet having the same master ticket as the invite.
So, this commit changes the wallet generation to use a ticket generated with non-deterministic `makeTicket` function.


yosoyubik:
Just tested it and looks good! Thanks @tomholford

PR #881, opened 2022-01-21
acceptance flow: new copy

nerveharp:
Adjusts copy to be a bit more explanatory.

(Screenshots forthcoming.)

### DIFF
--- a/src/views/Activate/ActivateCode.tsx
+++ b/src/views/Activate/ActivateCode.tsx
@@ -39,12 +39,45 @@ export default function ActivateCode() {
     <View inset hideBack>
       <ActivateView
         hideBack={hasInitialValues}
-        header={<ActivateHeader content={'Planet Code'} />}
+        header={
+          <ActivateHeader
+            content={
+              hasInitialValues
+                ? 'Welcome. This is your Urbit ID.'
+                : 'Activate a Planet'
+            }
+          />
+        }
         footer={<ActivateCodeForm afterSubmit={goToMasterKey} />}>
         <Box className="flex-col align-center justify-center w-full h-full">
-          {!hasInitialValues && (
+          {hasInitialValues ? (
+            <Box>
+              <p
+                className="mb2 sans gray5"
+                style={{
+                  fontSize: 14,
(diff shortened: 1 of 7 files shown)


e-shee:
Copy is clean. Looks good to me.

tomholford (approved):
LGTM, nice work James

jyng:
This is great. Small changes:

1) Activate -> Claim
2) Can you make the `Copy` link button, 14px?

nerveharp:
"Activate" button label now says "Claim":

image

"Copy" button label now uses 14px type:

image

Also performed a typographic cleanup on the disclaimer page:

image

jyng:
Oh, this update also includes the disclaimer, right.
The checkbox here needs to be using the indigo styling. Do you have access to those components?

nerveharp:
Unfortunately, Bridge includes an older version of Indigo _from source_ (so, not installed from npm package like Landscape or Groups, but literally just copied into the source tree).

I'm happy to address it, though?

nerveharp:
Phew, okay. Apparently we are using an older version of Indigo in Bridge, which is included in the source tree and not imported from `@tlon/indigo-react`.

I will hold off on a _wholesale component replacement_ as that's not really the priority at the moment, but I did manage to get the (latest) Indigo `Checkbox` into the `CheckboxInput` component, which is used in the Disclaimer.

image

This also helps this situation, not previously logged as a "dug" (design bug, thank you @liam-fitzgerald) in the Set Network Keys screen:

image

Now:

image

marisarow:
We still say Bridge / Activate at the top of the page. We should probably also change that to Claim.

nerveharp:
> We still say Bridge / Activate at the top of the page. We should probably also change that to Claim. 

I was going to ask about this. Does it even make sense to include? Do people need to know about what Bridge is at this point? They probably only just heard the words "Urbit" and "planet" in the past week.

e-shee:
To @nerveharp's point: I propose removing Bridge / Activate and either changing it to Claim or leaving both out altogether so that the instructional copy remains the focus.

PR #897, opened 2022-01-27
keygen: account for lib/keys revision decrement

tomholford:
This change increments the revisions used in the multiKeyfileGenerator to account for the decrement happening in libs/keys. Added a comment in-line with more detail.

Also, updated the expected args Typescript types, and improved the hook documentation.

### DIFF
--- a/src/lib/useMultikeyFileGenerator.ts
+++ b/src/lib/useMultikeyFileGenerator.ts
@@ -23,14 +23,18 @@ import { useRollerStore } from 'store/rollerStore';
 interface useMultikeyFileGeneratorArgs {
   seed?: string;
   point?: number;
-  seedWallet?: any;
-  inviteMasterTicketWallet?: any;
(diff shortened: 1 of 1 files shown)


tomholford:
An update on this PR: had an out-of-band conversation with @Fang- , I will try a different approach later today. Cross-posting his comments for future reference:

> okay, so, i'm pretty sure that what we should do is:
> - do the revision - 1 only at the kg.whatever callsites
> - only have kg.whatever callsites in lib/keys, and never ever import/use it outside of that
> - have as few unique kg.whatever callsites that take a revision argument as possible
> that way, as long as you always use keys.js, you never have to account for revision shenanigans, and just go by the chain-style revision number (ie "nth key", instead of "(n-1)th key")
> i think we already do more or less the above, so you shouldn't need to +1 the revisions. is it possible there's some off-by-one happening elsewhere?

PR #900, opened 2022-01-28
proxy: treat L1 points w/ spawn dominion as L2

### LINKED ISSUE #874: L2: incorrect display after setting spawn proxy to L2
Bridge unreliably detected a L1 star with its proxy set to L2. The user queried the back-end and got `"dominion": "spawn"`. They suspect there may be some type coercion when calculating the tx nonce (needs to be a string, not a number).

Possible fix: web+urbitgraph://group/~wolref-podlex/next-pki/graph/~wolref-podlex/general-1012/170141184505442700329412495133787553792

tomholford:
# Context
Before this change, when loading data from the roller API, points that had a dominion of 'spawn' were being treated like legacy L1 points. As such, Azimuth was queried as the source of truth, instead of the Roller API. Because of this, users were not seeing updated spawn proxy info in the Bridge UI.

Tested on Ropsten, was able to Migrate a star's spawn proxy to L2 and generate invites.

Open question: do we need another Layer indicator? Currently it shows as L1 in the Accounts Dropdown,
but L2 in the ID view

This fixes #874.

# Preview

## Account Dropdown Shows L1
image

## ID > Spawn Proxy Shows L2
image

### DIFF
--- a/src/lib/useRoller.ts
+++ b/src/lib/useRoller.ts
@@ -206,7 +206,7 @@ export default function useRoller() {
           ? await api.getRemainingQuota(pointNum)
           : 0;
 
-        const details = isL2(rawDetails?.dominion)
+        const details = isL2Spawn(rawDetails?.dominion)
           ? toL1Details(rawDetails)
           : await azimuth.azimuth.getPoint(_contracts, point);
 


nerveharp:
@arthyn We are short on devs at the moment. You mind taking a look at this?

yosoyubik:
just deployed this fixed to `bridge-dev.urbit.org`, tested it and it LGTM

PS: @tomholford networking keys are needed to create invites right? (ropsten is very slow at the moment, so i'm still waiting for that tx to be received by the roller...)

yosoyubik (approved):

Issue #888, opened 2022-01-24
copy when factory resetting a L2 planet does not tell you to wait for the timer


drbeefsupreme:
Right now, when you factory reset a L2 planet, there is no indication that you need to wait until the next batch is submitted before you can boot with the new keyfile. You'll get a key mismatch error if you try to boot before the next batch is submitted. One of the users I onboarded to L2 ran into this and was confused. So we should instruct the user to wait until the timer elapsed before booting after a factory reset.

nerveharp:
![image](https://user-images.githubusercontent.com/748181/150821122-9c582cd8-6c1b-4757-b2fc-e920126cbfd6.png)

nerveharp:
Adjusted the copy to reflect (A) our switch to the term "bootfile," (B) the important notice about waiting until the L2 batch goes through, and (C) a note about deleting your old pier before doing so.

![image](https://user-images.githubusercontent.com/748181/150847305-1c32b48f-0759-4af9-babe-e69a0c8776e5.png)

@drbeefsupreme @e-shee Comments welcome.

Issue #868, opened 2022-01-05
told to "boot with new key" instead of using |rekey when rotating keys without breaching


drbeefsupreme:
When you rotate keys without breaching, you need to import the new keys into a running ship, not boot a new one. Right now, Bridge tells you to boot with the new keys, when it should tell you to use `|rekey`. `|rekey` doesn't work with the new key format right now, but I fixed it yesterday so it will be in production soon.

nerveharp:
![image](https://user-images.githubusercontent.com/748181/150820310-a95a3278-72f0-4457-839a-0c8cb2dc6b11.png)

drbeefsupreme:
https://github.com/urbit/urbit/pull/5522

nerveharp:
I eliminated the "Download Keyfile" button in this instance, because if what we are saying is true, you don't _actually_ need your keyfile after re-keying.

![image](https://user-images.githubusercontent.com/748181/150847061-368b00c1-a7e7-4ffa-bd68-0ef7d93dd951.png)

@drbeefsupreme @e-shee Comments welcome.

drbeefsupreme:
I don't think that is correct. What you need to do is copy the contents of the keyfile, then run `|rekey 'keyfile-contents'`. `|rekey` by itself doesn't do anything, and you wouldn't be able to pull the key from Ethereum since it only has your public key, while your keyfile contains your private key.

nerveharp:
OK, so we will still need the "Download Keyfile" button. I will adjust the copy in a second pass.

nerveharp:
![image](https://user-images.githubusercontent.com/748181/150859300-91b1ab61-39af-4ae2-8759-4dba078da742.png)

drbeefsupreme:
LGTM

e-shee:
Copy is clean, lgtm.

Fang-:
>bootkey

Where did this word come from? The latest screenshot posted seems doubly confusing because it says "boot" but then doesn't actually instruct me to boot anything, to use my already-running ship instead? imo this is a strict downgrade from "keyfile".

(Not letting you download if you didn't breach is nice though. But we must be aware that this means the user cannot easily save the key for later use, if they do not have access to their dojo right this instant.)

nerveharp:
[...]

PR #904, opened 2022-02-01
bridge: more copyfixes

nerveharp:
- New copy for Factory Reset confirmation screen
- New `|rekey` instructions for normal rekey confirmation screen
- Eliminates 'Bridge / Activate' header in activation flow

Fixes urbit/bridge#868, fixes urbit/bridge#888, fixes urbit/bridge#881

### DIFF
--- a/src/views/UrbitOS/NetworkKeys.tsx
+++ b/src/views/UrbitOS/NetworkKeys.tsx
@@ -9,7 +9,7 @@ import { Just } from 'folktale/maybe';
 import { Grid, Button } from 'indigo-react';
 import * as azimuth from 'azimuth-js';
 import { randomHex } from 'web3-utils';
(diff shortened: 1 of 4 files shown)


src/views/UrbitOS/NetworkKeys.tsx
```diff
@@ -280,8 +280,41 @@ export default function UrbitOSNetworkKeys({
...
+      </Text>
+    </>
+  );
+
+  const waitMessage = () => (
+    <Text as="p" textAlign={'center'}>
+      <strong>Important:</strong> you will need to wait until the next L2 batch
+      is submitted to the roller to boot your ship. If you do not wait, then you
+      will run into a key mismatch error and will not be able to communicate
+      with other ships on the network.
```
Fang-, line 315: I think it will just prevent you from booting. We have preboot validation in place that checks the provided keys against the roller.

Issue #926, opened 2022-02-02
ops: no differentiation in resident/request copy for stars/galaxies


nerveharp:
The galaxy ops screen says "planets" in place of "stars", for the residents/requests.

![image](https://user-images.githubusercontent.com/748181/152224754-4fe08bfa-f6a7-429f-92a6-7e8424bf95ca.png)

Proposed solutionL sub for "ships" and disregard point-conditional language.

PR #927, opened 2022-02-02
galaxy ops: add conditional point subtext

nerveharp:
fixes urbit/bridge#926

image

image


tomholford (approved):
LGTM, nice work James

PR #932, opened 2022-02-03
invites-ui: Show only unclaimed L2 invites in Invites view

### LINKED ISSUE #903: invites-ui: booted L1 planets showing up in L2 invites list
After a [recent change](https://github.com/urbit/bridge/pull/900) to address #874, ~sitful-hatred reported that he is seeing booted L1 planets showing up in the Invites view:

image

To reproduce:
1. Log in with a wallet holding an L1 point with children
2. Click on Invites

Took a look at this last night, it has to do with `useControllerPointsStore` - specifically, all L1 and L2 invites are loaded into a single array in the application context. To fix this, we will need to filter out these L1 / booted points.

tomholford:
# Context

On load, Bridge queries for spawning points from both Azimuth and the Roller, and the results are concatenated into a single de-duplicated array. This data is stored in the `controlledPointsStore`, and eventually `rollerStore`.

The Invites view consumes `invites` data from the `rollerStore` to render the list of invite links, and the count of invites in the badge.

For L1 stars that had previously issued L1 invites, and then subsequently migrated spawn proxy to L2, users would see legacy L1 invites in the Invites view, along with their L2 invites:

image

# Changes

This resolves #903 by removing the legacy filter based on transferProxy, and adding one based on the spawning point's `dominion`.

# Testing

Thanks to @yapishu for reporting the original issue, and for helping QA this branch with his star as the test case.

Also verified on Ropsten locally

### DIFF
--- a/src/lib/useRoller.ts
+++ b/src/lib/useRoller.ts
@@ -1,6 +1,7 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import * as azimuth from 'azimuth-js';
 import * as need from 'lib/need';
+import * as ob from 'urbit-ob';
 import SecureLS from 'secure-ls';
 
 import { Invite } from 'lib/types/Invite';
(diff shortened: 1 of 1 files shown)


yosoyubik (approved):
LGTM

PR #935, opened 2022-02-03
transfer: safer star lockup detection

### LINKED ISSUE #908: Attempting to transfer a galaxy crashes Bridge
After logging in as a galaxy (master ticket or otherwise), navigating to the "ID" section and clicking "transfer" results in the following error:

```
23:05:52.556 Uncaught TypeError: can't access property "kind", a is null
    needLockupWarning Transfer.tsx:71
    map maybe.js:80
    needLockupWarning Transfer.tsx:71
    React 3
    l whyDidYouRender.js:161
    AdminTransfer Transfer.tsx:69
    React 12
    unstable_runWithPriority scheduler.development.js:468
    React 6
Transfer.tsx:71
```

Fang-:
In the event that star release store fails to fetch lockup details, it will
store a `Just(null)` value instead. Here, we need to account for the
possibility that `a` is null and cannot be indexed into.

Arguably it should just put `Nothing()` for the failure case? Though the points overview screen might depend on this being `Just` to determine whether it's done loading yet or not. We've had issues like this in the past though, so perhaps worth cleaning up, but out of scope for now.

(Note that the relevant lookup usually doesn't fail on mainnet, so this was really just a ropsten problem.)

Fixes #908.

### DIFF
--- a/src/views/UrbitID/Transfer.tsx
+++ b/src/views/UrbitID/Transfer.tsx
@@ -70,7 +70,9 @@ export default function AdminTransfer() {
     const az = azimuth.azimuth;
     return (
       az.getPointSize(point.value) === az.PointSize.Galaxy &&
-      starReleaseDetails.map((a: any) => a.kind).getOrElse('none') !== 'none'
+      starReleaseDetails
+        .map((a: any) => a?.kind || 'none')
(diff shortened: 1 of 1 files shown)


tomholford (approved):
LGTM :star:

PR #936, opened 2022-02-03
activate: form entry should override implied ticket

arthyn:
If you happen to get into the activate flow without a valid ticket, we kick you to a screen for manual entry. However, if you put junk in the URL to get there we were actually using that over what was entered into the form.


tomholford (approved):
LGTM, nice work :muscle:

PR #937, opened 2022-02-04
keys: generate multikeys in Activate flow, single keys in Home view

### LINKED ISSUE #895: 'pre-boot error: %not-keyed' when booting L1 planet with multikey
[this was originally reported as a comment on #892, this creates a separate issue for it]

[originally posted by](https://github.com/urbit/bridge/issues/892#issuecomment-1021705815) @nerveharp:

> @jyng was trying to boot an L1 ship
> he attempted to set his network keys for this new ship
> https://etherscan.io/tx/0x18e92441248c778d3ec70d92e3b306f5caae026de575202f63e3fa3ed148acae
> 
> the keyfile he got from bridge-staging is ~256 characters long (0wbgE.[...].Fv0r5) and although this is correct for L2 planets, this is an L1 ship and he cannot boot (see screenshot)
> he logged in to bridge-prod and retrieved the keyfile, which is ~128 characters long and works (0w1g6Ki.[...].B-BY1)
> Somebody who understands %keys please help, my family is dying

image

tomholford:
# Context
A previous refactor unified all keyfile generation to a single hook, useMultiKeyfileGenerator, to be used in all Keyfile download interactions. This turned out to cause issues booting L1 ships.

# Changes
This PR resolves #895 ~by restoring single keyfile generation for Home view touchpoints (Network Keys, Home > OS, etc.). It also updates the single keyfile generator to use the new compileMultikey implementation~

~There is also some renaming for improved readability, and simplification of the multi key generator (now that it's only used during Activation flow).~

*update*
To address PR feedback, the implementation of `useKeyfileGenerator` is refactored. It is now a base hook that can be used by itself, or composed within other React hooks. 

There are two new hooks that wrap it to provide the keyfile generation functionality, based on the application context:

In the Home view (Set Network Keys, Home > OS, Home > ID), a hook called `useHomeKeyfileGenerator` is now used.
In the Activation view, a hook called `useActivationKeyfileGenerator` is now used.

Both wrap the base hook, the difference is how the network key seeds are derived (for consumption by the base hook).

# Testing
- [x] Generate Ropsten keyfile during Activation
- [x] Generate Ropsten keyfile on Home > OS view
- [ ] Generate mainnet keyfile and verify boot


Fang- (commented):
This looks mostly good.

It's a bit weird to me to have both `useKeyfileGenerator` and `useMultiKeyfileGenerator`, when the former is really just a subset of the latter's functionality. They seem to duplicate a bunch of code as a result of that. Can we not have a flag for something like "include next pair too"?

PR #941, opened 2022-02-05
details-store: try both L1 and L2 when syncing details

arthyn:
So I'm not entirely sure this is right, but from what I could glean, it seems like the roller API was failing to find the new L1 point throwing an error from `api.getPoint`, so the L1 point details never get synced?  This just adds a backup call to try getting the point from azimuth similar to what you see in the `useRoller` hook under `initPoint`.


tomholford (approved):
LGTM, welcome to Bridge :bridge_at_night:

Issue #939, opened 2022-02-04
Rejecting a Transfer doesn't remove it from the Home view


yosoyubik:
**Describe the bug:**
Rejecting a Transfer doesn't remove it from the Home view, and looking at the point (after the transaction has been sent to the roller and processed) it seems that the proxy remains, the same, even though a change proxy transaction was sent out.

(Note: Cancelling the Transfer does reset the transfer-proxy) 

**To reproduce:**
1. Transfer a planet to an ETH address
2. Log in with that ETH address
3. Reject Transfer

**Expected behavior:**
The rejected point should disappear

**Screenshot or console log:**
<img width="523" alt="Screen Shot 2022-02-04 at 5 24 23 PM" src="https://user-images.githubusercontent.com/1033194/152565420-241e8505-e308-44df-8edc-7a8b66766aa7.png">
<img width="482" alt="Screen Shot 2022-02-04 at 5 23 04 PM" src="https://user-images.githubusercontent.com/1033194/152565423-297a66d2-0730-446d-b364-e5e2878fb7df.png">
<img width="501" alt="Screen Shot 2022-02-04 at 5 25 29 PM" src="https://user-images.githubusercontent.com/1033194/152565407-aba1801b-ca8f-4603-8759-468910296962.png">

<img width="509" alt="Screen Shot 2022-02-04 at 5 31 03 PM" src="https://user-images.githubusercontent.com/1033194/152566449-bd882eb1-4574-4ca1-a2fe-2e973a86f316.png">
<img width="935" alt="Screen Shot 2022-02-04 at 5 30 55 PM" src="https://user-images.githubusercontent.com/1033194/152566459-3cb2c654-275e-41b6-9833-fb3f93ef2636.png">

**Ethereum network:**
Ropsten

**Login method:**
Master Ticket

Issue #938, opened 2022-02-04
Clicking on Accept an Incoming point doesn't close the message


yosoyubik:
**Describe the bug:**
Accepting a transfer doesn't remove the "Incoming point" message, so the user might think that the transfer was not successful, click multiple times, which will send extra L2 txs that will later fail (and show up as such in the tx history). Note here that one of them will go through, and the others will fail—nonce here is probably reused—so this is only confusing UX/UI rather than an actual bug.

**To reproduce:**
1. Transfer a planet to an ETH address
2. Log in with that ETH address
3. Click on Accept multiple times
4. Click on History to see the Tx failures

**Expected behavior:**
After clicking on Accept, the "Incoming point" message should go away, after checking with the Roller that the TX is now in pending

**Screenshot or console log:**
<img width="482" alt="Screen Shot 2022-02-04 at 5 06 02 PM" src="https://user-images.githubusercontent.com/1033194/152563330-e40c23fe-75c2-4137-b834-a5c76b10121f.png">

<img width="511" alt="Screen Shot 2022-02-04 at 5 04 15 PM" src="https://user-images.githubusercontent.com/1033194/152563350-18a8e291-de44-45dd-a723-09bc12fc047e.png">

**Ethereum network:**
Ropsten

**Desktop (please complete the following information):**
 - OS: MacOS 10.14.6
 - Browser: Brave

**Login method:**
Master Ticket

PR #946, opened 2022-02-08
Fix grab-bag

arthyn:
This PR implements a few fixes associated with issues and a few things to just clean up:

- Fixes issue where clicking "Accept Transfer" made multiple transactions
- Fixes issue where "Incoming Point" box still shows upon going back to home screen after accepting transfer, #938 
- Hides rejected points, incoming, and outgoing points from the main list of points so they aren't duplicated, #939 
- Prevents multiple calls for `getInvites` from overlapping and generating the same wallets over and over

### DIFF
--- a/src/lib/useRoller.ts
+++ b/src/lib/useRoller.ts
@@ -1008,7 +1008,6 @@ export default function useRoller() {
       spawningPoints: number[]
     ) => {
       try {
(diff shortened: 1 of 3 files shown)


tomholford (approved):
Great work @arthyn  :muscle:  LGTM

PR #948, opened 2022-02-09
reticket: L1 points with L2 spawn proxy can reticket

### LINKED ISSUE #931: reticketing a star with L2 spawn proxy sends an L1 transaction to set spawn proxy
**Describe the bug:**
I have a Ropsten star with L2 spawn proxy on a master ticket, and reticketed to generate a new master ticket. Reticketing involves a transfer, set spawn proxy, configuring network keys, set management proxy, and another transfer. Once a star has its spawn proxy on L2, it is not eligible to make a L1 set spawn proxy action. However, this is what it does.

I believe L1 set spawn proxy is blocked by the Ecliptic on mainnet for stars with L2 spawn proxy, but that update isn't on Ropsten. However, it shouldn't be trying to do it all - it needs to send the set spawn proxy action on L2. 

**To reproduce:**
1. Login to a L1 star with L2 spawn proxy.
2. Click ID
3. Click "Reset" next to "Master Ticket"
4. Go through with the transaction.
5. Observe: https://ropsten.etherscan.io/tx/0x810e11ee48cbb7a922881d0a18859e0d2e1c497b5aa343804b74ee617c9ce1d6

**Expected behavior:**
The set spawn proxy action occurs on L2.

tomholford:
This closes #931 by changing the conditional check in `performL2reticket` so that L1 points with dominion of 'spawn' use the L2 flow.

This also updates the L2 implementation to hook into the progress bar animation as the 4 TXes are performed.


src/views/UrbitID/ResetKeys/ResetExecute.js
```diff
@@ -98,17 +128,16 @@ export default function ResetExecute({ newWallet, setNewWallet }) {
         walletType === WALLET_TYPES.WALLET_CONNECT ? wcSend : undefined;
 
       try {
-        if (isL2) {
-          console.log(newWallet.value);
+        if (isL2Spawn) {
           // FIXME: the useEffect is called twice, which is fine since the duplicate
           // L2 txs will be discarded
-          //
           await performL2Reticket({
```
yosoyubik, line 134: I think that what we want here is s "performL2SpawnReticket" that does the same as "reticketPointBetweenWallets" with the exception of doing the reset of the spawn-proxy as an L2 transaction, instead of an L1. Otherwise this is going to perform the whole reticketing flow on L2 which is not what we want for solving https://github.com/urbit/bridge/issues/931.

tomholford: Thanks for the feedback @yosoyubik, will make the change

yosoyubik:
I ran into a "Transaction has been reverted by the EVM:" so the L1 txs were not sent, but the L2 spawn proxy change was, we should probably check for that and not send the L2 tx.

(not sure if related but I keep getting "Transaction failed" for all L1 txs. this is on https://bridge-dev.urbit.org, Brave for ~navdel)

arthyn (approved):
From what I can tell the JS looks solid, no idea if these operations are right though haha

tomholford:
Per out of band conversation with @yosoyubik, just pushed one more commit that uses the `force` param in the `roller-api`'s `setProxy` function, because the L1 spawn reticket flow is a special case where we need to override the roller predicted state error.

tomholford:
An update: I made one more change per another out of band conversation with @yosoyubik re: the signing wallet for the final TX. I have been trying to test, but Ropsten is being really flaky right now. Will try again later and report back

tomholford:
Confirmed on Ropsten with ~ricnet. Was able to reticket and login with the new Master Ticket

PR #949, opened 2022-02-09
Remove birthday store

arthyn:
This showed up originally as an infinite loop seen at the end of an L1 transaction, basically anything using `useEthereumTransaction` that also has it's refetch callback depend upon either `syncExtras` or `syncDates`, which in this case was `ChangeSponsor`. I was originally going to just fix the issue inside birthday store:

```
const addToBirthdayCache = useCallback(
    entry =>
      _setBirthdayCache(cache => ({
        ...cache,
        ...entry,
      })),
    [_setBirthdayCache]
  );

  const getBirthday = useCallback(point => birthdayCache[point] || Nothing(), [
    birthdayCache,
  ]);

  const syncBirthday = useCallback(
  ...
  addToBirthdayCache({
  ...
  [contracts, web3, addToBirthdayCache, getBirthday]
```

You can maybe spot the issue, but it's kinda subtle. Basically the call to `syncBirthday` rebuilds the birthday cache every time because `addToBirthdayCache` creates a new object each time. Since `getBirthday` also has `birthdayCache` as a dependency and `getBirthday` is a dependency of `syncBirthday` that means that anytime `syncBirthday` is called, it changes itself and shows up as changed to anyone depending on it. So way down the line this causes an infinite loop. When I started looking for usages of birthday stuff I noticed we don't actually use the contents of the birthday store anywhere so I just decided to remove it since it's unnecessary.


Fang- (approved):
Less code is best code. Was kind of a meaningless feature anyway.

tomholford (approved):
wonderful! LGTM :white_check_mark:

PR #960, opened 2022-02-11
roller: personal sign issue

### LINKED ISSUE #957: L2 planet spawning: unable to spawn planets to address
From a community member:

Successfully migrated spawn proxy of ~walpub to L2, but unable to spawn planets:

```
errors.js:28 Uncaught (in promise) Error: Returned error: The method personal_sign does not exist/is not available
    at Object.ErrorResponse (errors.js:28:19)
    at index.js:302:36
    at XMLHttpRequest.i.onreadystatechange (index.js:98:13)
```

Running on macOS in chrome, ~walpub lives on a Ledger Nano X.

image

arthyn:
This potentially fixes #957 using the same solve as #647. It also makes an attempt to prevent the "execution reverted" issue seen on login when you have the MM extension installed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4720,6 +4720,11 @@
       "resolved": "https://registry.npmjs.org/@ledgerhq/logs/-/logs-6.2.0.tgz",
       "integrity": "sha512-SLyFyD7ElMhgKWPYedFGCT/ilcbGPgL5hXXYHxOM79Fs5fWi0zaUpt5oGqGMsOAAFaMa9/rbun0pokzPhEFz8A=="
     },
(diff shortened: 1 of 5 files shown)


tomholford:
LGTM, and agreed with both @arthyn and @Fang-  that we should eventually replace this, but after launch (as long as things are still working!)

tomholford:
Confirming this works for me locally, was able to do the following on Ropsten:

- Set Keys
- Spawn a new L1 point
- Spawn a new L2 point

tomholford (approved):
LGTM, the cleaner console is much appreciated :sparkles:

PR #963, opened 2022-02-12
l2transaction: update method signature

arthyn:
This updates the typings so they match and also updates the signature of the method call. Wasn't sure if force should be false or true, but it seemed defaulted to false elsewhere.


tomholford (approved):
LGTM, nice work :muscle:

PR #971, opened 2022-02-15
roller: move polling, less ddos

arthyn:
This PR adjusts the polling behavior for next batch time so that it doesn't have the capability of sending off lots of requests as soon as the roller falls behind. It also refactors the polling behavior to live in it's own file so that we're not adding to the giant useRoller file.

### DIFF
--- a/src/lib/useRollerPoller.ts
+++ b/src/lib/useRollerPoller.ts
@@ -1,0 +1,42 @@
+import { useEffect, useState } from 'react';
+import { useRollerStore } from 'store/rollerStore';
+import { useTimerStore } from 'store/timerStore';
+import { ONE_SECOND, TEN_SECONDS } from './constants';
+import useRoller from './useRoller';
+import { getTimeToNextBatch } from './utils/roller';
+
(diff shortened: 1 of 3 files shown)


src/lib/useRollerPoller.ts
```diff
@@ -0,0 +1,42 @@
...
+  const { point } = useRollerStore();
+  const { api, getInvites, getPendingTransactions } = useRoller();
+  const [time, setTime] = useState(ONE_SECOND);
+
+  useEffect(() => {
+    const interval = setInterval(async () => {
+      const nextRoll = getTimeToNextBatch(nextBatchTime, new Date().getTime());
+      setNextRoll(nextRoll);
+
+      if (nextBatchTime - ONE_SECOND <= new Date().getTime()) {
```
arthyn, line 19: I'm still unsure about this logic because it ends up firing multiple times. We could lock this down so it only fires once in that timeframe, but may be worth putting off.

tomholford (approved):
LGTM

PR #976, opened 2022-02-20
api: configurable Roller endpoint via env

### LINKED ISSUE #974: api: configurable Roller endpoint
By default, Bridge uses `roller.urbit.org` for L2 transactions.

Based on discussion in Star Squawk, some users are standing up their own rollers, so it would be nice to be able to point Bridge to one of the alternatives (e.g., via a Settings menu).

for context:
web+urbitgraph://group/~tirrel/the-marketplace/graph/~marbus/the-marketplace/170141184505500860797923642766694285312

tomholford:
Before this change, Bridge was configured to only use the L2 roller provided by Tlon.

This change allows a Bridge user to use one or more runtime env vars to connect to a custom roller.

For example:

```sh
REACT_APP_ROLLER_HOST=my-personal-roller.net npm run pilot-mainnet
```

The following are configurable, and will otherwise fallback to the defaults in `constants`:
- `REACT_APP_ROLLER_HOST` - host
- `REACT_APP_ROLLER_PORT` - port
- `REACT_APP_ROLLER_TRANSPORT_TYPE` - transport type (e.g., `http` or `https`)

Resolves #974

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,3 +54,19 @@ A proper "offline mode" for the latest iteration of Bridge is still being worked
 ## Development
 
 See [development.md](DEVELOPMENT.md).
+
+
+## Configurable Roller Endpoint
+
+By default, Bridge uses Tlon's L2 Roller, but can also be configured to use your own.
+
+For example:
+
+```sh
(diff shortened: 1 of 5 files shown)


yosoyubik (approved):
LGTM

PR #984, opened 2022-02-23
auth: use keccak256 to derive token

### LINKED ISSUE #549: signing function mismatch between Metamask / ETH toolkit

tomholford:
# Context

See the recent comments in #549 for details, but in short:

[This commit](https://github.com/urbit/bridge/commit/003c08396edb548527e8ffd56218fca90d7f2d8b#diff-ea521a1b242dfb1472db45d56b3f869a255d301f033f3c2d0aeb504c316c24bcL17) introduced a discrepancy between how auth tokens are generated between custodial and non-custodial wallets. Specifically, [per the docs](https://wallet-docs.brave.com/ethereum/use-cases/signing-data/#personal_sign) for `personal_sign`, `keccak256` is used by Ethereum Providers to sign a message.

When I first joined the Bridge project and had little insight into how things worked, I instead used the subtly different `sha256` for the default auth token case (i.e, for master tickets and mnemonics). As such, for roughly the past 6 months, any actions performed in Bridge (generating invites, setting keys, etc) using these login methods were with a logically incorrect auth token.

# Changes

This resolves #549 by:

- using `keccak256` for default auth token derivation 
- adding a "Legacy Compatibility" login flow for custodial wallets; when this is enabled, the auth token is derived using `sha256` instead of `keccak256` to allow users to re-derive deterministic keys

# Preview

## Legacy Toggle
image

## When Keyfile is Unavailable
image

## Info Modal
image

# Testing

Confirmed using the dev mnemonic on Ropsten. 

1. Log in with Recovery Phrase
2. Select ~winnet
3. Navigate to Home > OS, verify that Keyfile is unavailable
4. Logout
5. Log in with Recovery Phrase again, this time with Use Legacy Compatibility enabled
6. Select ~winnet
7. Navigate to Home > OS, verify that Keyfile can be downloaded

### DIFF
--- a/src/views/UrbitOS/Home.tsx
+++ b/src/views/UrbitOS/Home.tsx
@@ -7,6 +7,7 @@ import Window from 'components/L2/Window/Window';
 import HeaderPane from 'components/L2/Window/HeaderPane';
 import BodyPane from 'components/L2/Window/BodyPane';
 import { ReactComponent as KeyfileIcon } from 'assets/keyfile.svg';
+import { ReactComponent as InfoIcon } from 'assets/info.svg';
 
 import { useLocalRouter } from 'lib/LocalRouter';
 import { useSingleKeyfileGenerator } from 'lib/useKeyfileGenerator';
@@ -16,7 +17,6 @@ import CopiableWithTooltip from 'components/copiable/CopiableWithTooltip';
 import './UrbitOS.scss';
 import Modal from 'components/L2/Modal';
 import { useRollerStore } from 'store/rollerStore';
-import WithTooltip from 'components/WithTooltip';
 
 export default function UrbitOSHome() {
   const { point } = useRollerStore();
@@ -47,6 +47,9 @@ export default function UrbitOSHome() {
     {}
   );
   const keyfileAvailable = available && !generating;
+  const [showKeysUnavailableModal, setShowKeysUnavailableModal] = useState(
+    false
+  );
 
   const headerButton = () => {
     if (!hasSetNetworkKeys) {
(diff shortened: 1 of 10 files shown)


src/views/UrbitOS/Home.tsx
```diff
@@ -42,22 +43,40 @@ export default function UrbitOSHome() {
...
+      return <Box className="header-button keyfile">Generating...</Box>;
+    }
+
+    return keyfileAvailable ? (
+      <Button className="header-button keyfile" onClick={download}>
+        <KeyfileIcon />
+        Download Keyfile
+      </Button>
+    ) : (
+      <WithTooltip content="Either the key was derived non-deterministically, or try Login > Use Legacy Compatibility">
```
Fang-, line 66: This tooltip is super awkward lol.

Here and everywhere in the codebase, I would probably not use "legacy". Implies it was ever intentional, but it was really just a temporary buggy state. Probably "bugcompat" or w/e for variable names, not sure how to succinctly communicate this on the frontend.

...On further thinking, maybe the user-facing side is the only place "legacy" makes sense to use?

What you really want to communicate is that "if you generated a keyfile between 2021-05 and 2022-02, and you expected it here, try the login thing instead". But that feels _way_ too verbose for a tooltip...

We also probably want to put a little (i) here to indicate there is a tooltip in the first place.

arthyn: Yeah I think a an (i) icon inside the button would help here. 

I also think the wording could be different, trying to come up with something. Could we possibly just do the relogin for them on a button press? It almost seems like this is worthy of a modal just so we could fit a bit more text and maybe an action.

Just spitballing here, some bits may be wrong:
"We're unable to derive your keyfile. This is likely because it was generated using a signing method from the last six months or derived non-deterministically. Try logging in again with the previous signing method." and then have a button that does the relogin or brings them to the login form. I'm somewhat worried they won't find it behind the gear so we may need extra language to help them find it if we just redirect to the login page.

tomholford: @Fang- Thanks for the feedback. Cleaned up the UX based on this comment, but still not happy with the naming. Will think about how we can better label things more clearly for users.

tomholford: An update on this, I have looped in @jyng so we can leverage his expertise.

tomholford:
Per out of band conversations with @Fang-, @nerveharp, and @jyng, we've come up with an approach with significantly better UX. Instead of providing the option to the user upon Login, instead at the time of keyfile generation, we will derive tokens using both sha256 and keccak256 and see if either matches the on-chain keys.

PR #992, opened 2022-03-08
login: fix shard validation

arthyn:
This cleans up the validation so that we make it reactive based off of the useShards toggle.

### DIFF
--- a/src/components/L2/Headers/AdvancedOptions.tsx
+++ b/src/components/L2/Headers/AdvancedOptions.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import { Text, Box, Checkbox, Icon, Row } from '@tlon/indigo-react';
 import './AdvancedOptions.scss';
+import styled from 'styled-components';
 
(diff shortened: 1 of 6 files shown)


src/components/L2/Headers/AdvancedOptions.tsx
```diff
@@ -11,6 +12,15 @@ interface AdvancedOptionsProps {
   }[];
 }
 
+// Hide this input completely
+const HiddenInput = styled.input`
```
tomholford, line 16: There's a `HiddenInput` exported from `src/forms/inputs`, and typically I would encourage reuse. But since we're eventually moving away from Indigo, fine by me.

tomholford, line 16: There's a `HiddenInput` exported from `src/forms/inputs`, and typically I would encourage reuse. But since we're eventually moving away from Indigo, fine by me.

tomholford (approved):
LGTM :+1:

PR #993, opened 2022-03-10
invite-converter: always show button, add empty state

arthyn:
Based on some confusion about the invite button showing or not, I think it's clearer to always have it and just show an empty state if no planets are available to update.


tomholford (approved):
LGTM

PR #1000, opened 2022-03-15
keys: use sha256 token if unable to re-derive with keccak256 token

### LINKED ISSUE #549: signing function mismatch between Metamask / ETH toolkit

tomholford:
# Context

When re-deriving a keyfile and access code within the Bridge dashboard, derive using both the default auth token (derived with keccak256) and an optional fallback token (derived with sha256, if necessary).

This resolves #549.

# Testing

1. `git checkout t/549-auth-token-fallback`
2. `npm run pilot-ropsten`
3. log in with mnemonic from DEVELOPMENT.md
4. select ~winnet, navigate to Home > OS, and download Keyfile

### DIFF
--- a/src/lib/authToken.ts
+++ b/src/lib/authToken.ts
@@ -109,6 +109,23 @@ const getDefaultAuthToken = ({ wallet }: DefaultAuthTokenArgs) => {
   return token;
 };
 
+/**
+ * Heads up, this is probably not the authToken derivation function you're
+ * looking for.
+ *
+ * Due to a logical error with auth token derivation, keys set between May
+ * 2021 and March 2022 used `sha256` instead of the logically correct
+ * `keccak256`. To mitigate this for users who log back into Bridge to
+ * download their keyfile or copy their access code, the keyfile generator
+ * will first try re-deriving with `keccak256`, then will fallback to `sha256`
+ * via this function.
+ *
+ * See this issue for additional context:
+ * https://github.com/urbit/bridge/issues/549
+ *
+ * @param wallet BIP32 wallet
+ * @returns string
+ */
 export const getSha256AuthToken = ({ wallet }: DefaultAuthTokenArgs) => {
   const signature = signMessage(wallet.privateKey!, crypto.sha256);
 
--- a/src/lib/authToken.ts
+++ b/src/lib/authToken.ts
@@ -9,13 +9,13 @@ import { ledgerSignMessage } from './ledger';
 import { trezorSignMessage } from './trezor';
 import BridgeWallet from './types/BridgeWallet';
 import { Hash } from '@urbit/roller-api';
+import { keccak256 } from 'ethereumjs-util';
 
 const MESSAGE = 'Bridge Authentication Token';
 
-function signMessage(privateKey: Buffer) {
+function signMessage(privateKey: Buffer, hashFunction = keccak256) {
   const msg = '\x19Ethereum Signed Message:\n' + MESSAGE.length + MESSAGE;
-  // #ecdsaSign requires a 32-byte buffer, hence sha256
-  const hashed = crypto.sha256(Buffer.from(msg));
+  const hashed = hashFunction(Buffer.from(msg));
   const { signature } = ecdsaSign(Buffer.from(hashed), privateKey);
 
   // add key recovery parameter
@@ -103,6 +103,14 @@ const getWalletConnectAuthToken = ({
 
 const getDefaultAuthToken = ({ wallet }: DefaultAuthTokenArgs) => {
   const signature = signMessage(wallet.privateKey!);
+
+  const token = `0x${Buffer.from(signature).toString('hex')}`;
+
+  return token;
+};
+
+export const getSha256AuthToken = ({ wallet }: DefaultAuthTokenArgs) => {
+  const signature = signMessage(wallet.privateKey!, crypto.sha256);
 
   const token = `0x${Buffer.from(signature).toString('hex')}`;
 
--- a/src/lib/useKeyfileGenerator.ts
+++ b/src/lib/useKeyfileGenerator.ts
@@ -16,7 +16,7 @@ import { stripSigPrefix } from 'form/formatters';
 import useRoller from './useRoller';
 import Point from './types/Point';
 import { toL1Details } from './utils/point';
-import { ETH_ZERO_ADDR } from './constants';
+import { ETH_ZERO_ADDR, NONCUSTODIAL_WALLETS } from './constants';
 import { getSha256AuthToken } from './authToken';
 
 interface useKeyfileGeneratorArgs {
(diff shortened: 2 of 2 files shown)


src/lib/authToken.ts
```diff
@@ -109,6 +109,14 @@ const getDefaultAuthToken = ({ wallet }: DefaultAuthTokenArgs) => {
   return token;
 };
 
+export const getSha256AuthToken = ({ wallet }: DefaultAuthTokenArgs) => {
```
Fang-, line 112: I'd add a little warning comment here that this is likely not the one you want to use, and refer to the comment in `useKeyfileGenerator` for further details.

(One could argue that the main comment should live here, and the hook should refer to the lib for further details, but I don't think it matters all that much at the moment.)

src/lib/useKeyfileGenerator.ts
```diff
@@ -242,14 +253,56 @@ export const useSingleKeyfileGenerator = ({
       revision: point.keyRevisionNumber,
     });
 
-    if (Just.hasInstance(derivedSeed)) {
-      setSeeds([derivedSeed.value]);
+    if (Just.hasInstance(defaultDerivedSeed)) {
+      setDefaultSeeds([defaultDerivedSeed.value]);
+    }
```
Fang-, line 258: I think this wants an early-out? We don't want to proceed to fallback logic if we got the default one, right?

```suggestion
      setDefaultSeeds([defaultDerivedSeed.value]);
      return;
    }
```

src/lib/useKeyfileGenerator.ts
```diff
@@ -242,14 +253,56 @@ export const useSingleKeyfileGenerator = ({
...
+
+    /**
+     * Otherwise, we will need to try a different token derivation scheme.
+     *
+     * Due to a logical error with auth token derivation, keys set between May
+     * 2021 and March 2022 used `sha256` instead of the logically correct
+     * `keccak256`. To mitigate this for users who log back into Bridge to
+     * download their keyfile or copy their access code, we need to check for
+     * the existence by deriving with a fallback algorithm.
```
Fang-, line 267: The "otherwise" here doesn't flow well with the previous comment. I'd do "if we can't" or something similar.

Also, this fails to mention that this issue only affected custodial wallets. Probably worth calling out explicitly, even (or especially?) if the code below does so too.

src/lib/useKeyfileGenerator.ts
```diff
@@ -242,14 +253,56 @@ export const useSingleKeyfileGenerator = ({
...
+     *
+     * See this issue for additional context:
+     * https://github.com/urbit/bridge/issues/549
+     */
+
+    if (NONCUSTODIAL_WALLETS.has(walletType)) {
+      console.log('cannot derive fallback token for non-custodial wallets');
```
Fang-, line 274:
```suggestion
      console.log('fallback token not relevant for non-custodial wallets');
```

src/lib/useKeyfileGenerator.ts
```diff
@@ -242,14 +253,56 @@ export const useSingleKeyfileGenerator = ({
...
+      return;
     }
-  }, [authMnemonic, authToken, point, seed, urbitWallet, wallet]);
+
+    const fallbackAuthToken = getSha256AuthToken({ wallet: wallet.value });
+    const fallbackDerivedSeed = deriveNetworkSeedFromAuthToken(
```
Fang-, line 279: Don't we want `attemptNetworkSeedDerivation` here too? We want this to fail if it doesn't match the on-chain keys.

Fang- (commented):
Couple nits & questions, but otherwise looks good. Notably a much smaller diff than #984, this is clearly the better fix.

src/lib/useKeyfileGenerator.ts
tomholford: Typically we would, but in this case we want to derive both seeds if possible. The comparison to on-chain keys happens downstream in `useKeyfileGenerator` (which is wrapped by this hook). Will add some comments to reflect this behavior.

src/lib/useKeyfileGenerator.ts
tomholford: The on-chain check is happening in `useKeyfileGenerator#89`, which is downstream from this function. Will update the comments to reflect this.

src/lib/authToken.ts
tomholford: Good call, clarified the comments based on this feedback

src/lib/useKeyfileGenerator.ts
tomholford: This phrasing was not clear and potentially confusing, updated it to reflect the actual expected behavior.

src/lib/useKeyfileGenerator.ts
tomholford: At this point in the flow, we still don't know which token could potentially be valid. The `keysMatchChain` downstream will perform the validation.

tomholford:
An update on this PR: while testing this change with the Invites feature, I ran into an issue: invites generated with a SHA token cannot be re-derived with a keccak token (and thus checked to see if they have been claimed).

Had an out of band conversation with @Fang- to brainstorm ideas. Will update once we've decided the path forward.

tomholford:
cross-posting from conversation with Fang:

Did some more testing on this PR. It fixes the keys issue, but introduces a separate issue with the invites flow.

Using master, I logged in with my mnemonic, generated a few L2 invites with my
star, then logged out. these invites were generated with the sha token.
then, using the fix branch, I logged in with mnemonic again and navigated to the
invites view.
to render the list of invites, Bridge loads the outgoing points from the roller,
then filters out the pending spawns and claimed points. to determine if a point
is claimed, the invite point's inviteWallet is re-derived from the logged in
auth token; then, the inviteWallet's ownership address is compared against the
point's ownership address from the roller. if they match, it is considered
unclaimed, and a URL is rendered for the invite.
one naive approach is to try re-deriving the wallet with both types of tokens.
but generating a wallet takes time, so this will significantly slow down the
experience.

Will brainstorm other approaches with Fang and Product team

nerveharp:
> to render the list of invites, Bridge loads the outgoing points from the roller,
then filters out the pending spawns and claimed points.

Very open to a different, roller-side approach to this.

tomholford:
> > to render the list of invites, Bridge loads the outgoing points from the roller,
> > then filters out the pending spawns and claimed points.
> 
> Very open to a different, roller-side approach to this.

Good thinking, I will reach out to yosoyubik and get his thoughts.

tomholford:
Following up with another update:

Spoke with @yosoyubik about this, he mentioned that ~tirrel is working on a custom gall agent that implements the invite flow (as an alternative to Bridge). Once it's finished, perhaps we can upstream part or all of its functionality into the roller. Then we could expose an endpoint that returns just the list of unclaimed invites. Tjhis would make the frontend experience in Bridge much smoother, and also avoid the "missing" invites issue described in [this comment](https://github.com/urbit/bridge/pull/1000#issuecomment-1076354316).

At this time, issue #549 is not mission critical - a simple workaround is for the user to login with whichever method they used to generate the keyfile / invites. As such, I propose that we wait for this new functionality to be available.

tomholford:
After further conversation with @nerveharp, we have decided to close this in favor of #1012.

PR #1005, opened 2022-03-21
ux: show countdown to next roll

### LINKED ISSUE #1004: ui: timer starts countdown from 60:00:00 to next roll
Upon logging in while having pending transactions, the time to next roll is being set to 1 hour by default.

It should instead display the time to next roll.

In the below image, the timer shows 59 minutes to the next roll, but it's actually at 8:50 PM (~3 minutes from now)
image

tomholford:
On load, Bridge queries the roller for the Roller Config. One of the properties in the response is time to next batch. This value drives the countdown timer in the UI.

Before this change, it was being set in the legacy state context. This change removes the deprecated state context from rollerStore in favor of timerStore.

# Preview

image

Resolves #1004


arthyn (approved):
Thanks for jumping on this quickly!

PR #1012, opened 2022-03-26
auth: use keccak256 to derive token

### LINKED ISSUE #549: signing function mismatch between Metamask / ETH toolkit

tomholford:
# Context

This PR re-opens #984. After the initial review, we decided to explore an alternative solution (see: #1000). Unfortunately, that approach ended up not being feasible. So, here we are again :) Although the UX is not ideal, it stops the bleeding and provides a suitable fix for now. Eventually, we plan to split Bridge into separate apps, and can rethink this design when the time comes.

See the recent comments in #549 for details, but in short:

[This commit](https://github.com/urbit/bridge/commit/003c08396edb548527e8ffd56218fca90d7f2d8b#diff-ea521a1b242dfb1472db45d56b3f869a255d301f033f3c2d0aeb504c316c24bcL17) introduced a discrepancy between how auth tokens are generated between custodial and non-custodial wallets. Specifically, [per the docs](https://wallet-docs.brave.com/ethereum/use-cases/signing-data/#personal_sign) for `personal_sign`, `keccak256` is used by Ethereum Providers to sign a message.

When I first joined the Bridge project and had little insight into how things worked, I instead used the subtly different `sha256` for the default auth token case (i.e, for master tickets and mnemonics). As such, for roughly the past 6 months, any actions performed in Bridge (generating invites, setting keys, etc) using these login methods were with a logically incorrect auth token.

# Changes

This resolves #549 by:

- using `keccak256` for default auth token derivation 
- adding a "Legacy Token" login flow for custodial wallets (Master Ticket or Recovery Phrase); when this is enabled, the auth token is derived using `sha256` instead of `keccak256` to allow users to re-derive deterministic keys and view unclaimed invites created with the previous token algorithm

# Preview

## Legacy Toggle
image

## When Keyfile is Unavailable
image

## Info Modal
image

# Testing

Confirmed using the dev mnemonic on Ropsten. 

1. Log in with Recovery Phrase
2. Select ~winnet
3. Navigate to Home > OS, verify that Keyfile is unavailable
4. Logout
5. Log in with Recovery Phrase again, this time with Use Legacy Token enabled
6. Select ~winnet
7. Navigate to Home > OS, verify that Keyfile can be downloaded

### DIFF
--- a/src/views/UrbitOS/Home.tsx
+++ b/src/views/UrbitOS/Home.tsx
@@ -7,6 +7,7 @@ import Window from 'components/L2/Window/Window';
 import HeaderPane from 'components/L2/Window/HeaderPane';
 import BodyPane from 'components/L2/Window/BodyPane';
 import { ReactComponent as KeyfileIcon } from 'assets/keyfile.svg';
+import { ReactComponent as InfoIcon } from 'assets/info.svg';
 
 import { useLocalRouter } from 'lib/LocalRouter';
 import { useSingleKeyfileGenerator } from 'lib/useKeyfileGenerator';
@@ -16,7 +17,6 @@ import CopiableWithTooltip from 'components/copiable/CopiableWithTooltip';
 import './UrbitOS.scss';
 import Modal from 'components/L2/Modal';
 import { useRollerStore } from 'store/rollerStore';
-import WithTooltip from 'components/WithTooltip';
 
 export default function UrbitOSHome() {
   const { point } = useRollerStore();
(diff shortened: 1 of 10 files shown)


arthyn (approved):
LGTM!

src/views/UrbitOS/Home.tsx
```diff
@@ -42,22 +43,57 @@ export default function UrbitOSHome() {
...
+          <InfoIcon />
+          &nbsp;Keyfile Unavailable
+        </Button>
+        <Modal
+          show={showKeysUnavailableModal}
+          hide={() => setShowKeysUnavailableModal(false)}>
+          <Box className="info-modal-content">
+            <div className="fw-bold mb5">Keyfile Unavailable</div>
+            <div className="mb5">
+              If you set networking keys with an older version of Bridge, try
```
nerveharp, line 82: Good. Thank you!

nerveharp (approved):

PR #1028, opened 2022-06-15
localhost 403s on macOS

zalberico:
Using 127.0.0.1 works - I think this is probably an OS thing

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Optionally, to validate your downloaded file's integrity, compare the lines in c
 
 1. `cd` into the `bridge-$version` directory
 2. Run `python3 -m http.server 5000 --bind 127.0.0.1`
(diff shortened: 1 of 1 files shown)


Fang-:
`localhost` has always worked for me on macos. In `/etc/hosts`, for me, the very first entry is:

```
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
##
127.0.0.1 localhost
```

Are you sure this is not a local setup issue?

zalberico:
I don't think so - I have that same default config (untouched) in my hosts file. I tried Safari and Chrome/Brave to see if it was browser related. I also stopped locally running ships in case it had something to do with those. I think this is a genuine issue - it also seems to be the reason I can't access it over the network with tailscale (when using the tailscale ip I see the same. 403 related errors).

zalberico:
Yeah I'm not sure the underlying cause - I suspect whatever it is, is also why I can't access it over tailscale.

Issue #1024, opened 2022-06-01
V1 Planet Card Invite Codes No Longer Work 


bioy-casares:
8 syllable code is not being validated in the "Activate" field. 

ex: ~wacden-bardel-tidrym-filnyr  (actual code, feel free to test it)

These no longer work in the "Activate" field. The dummy text is now 12 syllables long. 

Approximately 1000 planet cards were made by Tlon with this convention. Lots were handed out this past week in NYC. 

<img width="643" alt="Screen Shot 2022-05-31 at 7 16 10 PM" src="https://user-images.githubusercontent.com/55156412/171314541-608b3abc-4814-41c6-91d5-82bccde2de6c.png">
<img width="719" alt="Screen Shot 2022-05-31 at 7 15 46 PM" src="https://user-images.githubusercontent.com/55156412/171314544-945eb39e-beef-4968-a0bd-605a7c9e024a.png">

matildepark:
The last two are the planet name I believe. Try that

bioy-casares:
Unfortunately, the cards don't contain the planet name. All there is the 8 syllable code, which also does not in any way resolve to the @p.

matildepark:
Ah. The Foundation's did have that but I guess those are V2 then.

bioy-casares:
Indeed. These ones are also L1, btw.

bioy-casares:
Bump. 

There's an important Bitcoin conference coming in August, and I'd like to be able to give away all the cards I have.

matildepark:
It’s not clear to me who owns Bridge.

bioy-casares:
@nerveharp

langalisc:
this is still broken. 

bridge.urbit.org/sunset also doesn't recognize these codes.

ilyakooo0:
The V1 invite codes work on https://bridge-sunset.urbit.org

bioy-casares:
The cards have just "bridge.urbit.org" written on them.

ilyakooo0:
@bioy-casares i know. I’m just saying what works (:

PR #1037, opened 2022-07-21
login: adds L1 activation banner for BTC conference

nerveharp:
Adds a small banner for BitBlockBoom attendees to redeem a L1 planet via `bridge-sunset`.

@bioy-casares will be distributing 500 L1 invite cards, so this fixes urbit/bridge#1024 in the least-intensive way we can at the moment.

### DIFF
--- a/src/views/Login/LoginSelector.tsx
+++ b/src/views/Login/LoginSelector.tsx
@@ -192,6 +192,13 @@ export default function LoginSelector({
         <Grid.Item full className="activate" as={Button} onClick={goToActivate}>
           Activate a Planet
         </Grid.Item>
+        <Grid.Item full className="l1-activation" as={Text}>
+          BitBlockBoom attendee?{' '}
+          <a className="underline" href="https://bridge-sunset.urbit.org">
+            Activate
+          </a>{' '}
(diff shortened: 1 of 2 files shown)


src/views/Login/LoginSelector.tsx
```diff
@@ -192,6 +192,13 @@ export default function LoginSelector({
         <Grid.Item full className="activate" as={Button} onClick={goToActivate}>
           Activate a Planet
         </Grid.Item>
+        <Grid.Item full className="l1-activation" as={Text}>
+          BitBlockBoom attendee?{' '}
```
tomholford, line 196: @nerveharp Since there are probably some other L1 invite cards floating around out threre, perhaps we could make this more generic? e.g., "Have an invite card with 8 words? "

ilyakooo0: you probably mean 4 words (8 syllables)

jamesacklin:
Looks like there's some renewed interest in this, given the chatter in #1024. I don't think anyone at Tlon has the bandwidth to re-integrate the L1 activation flow back into Bridge. I addressed the line comment above.

PR #1050, opened 2022-10-07
ux: option to skip token signing on login

tomholford:
# Context

This change introduces the option to skip eagerly generating an auth token on login using the following checkbox on the Login view:

image

Instead, a fake token is generated using random entropy. For reference, the auth token is used to derive and configure network keys, and to spawn / generate planet invites. 

Here is an excerpt from the conversation that inspired this change:

> ~wicdev-wisryt
> is there any way at all we can override the need for a signature on login from bridge?
> it's not terrible if you're working in metamask but it's terrible for anyone using a multisig or custodied address
> it means to log into bridge you use walletconnect, then go get all the other signers on the account to approve the signature, then you can do what you needed to do
> which is practically unworkable
> [...]
> at the very least we should just have a button that shows up at the same time as the signature request that fills in the signature with randomness
> a "skip signature" button

# Future Work

- Lazily generate the token on demand for the affected flows (network keys and spawning planets)

### DIFF
--- a/src/lib/useWalletConnect.ts
+++ b/src/lib/useWalletConnect.ts
@@ -30,7 +30,13 @@ type ConnectEvent = {
 };
 
 export const useWalletConnect = () => {
-  const { setWallet, setAuthToken, resetWallet }: any = useWallet();
+  const {
+    setWallet,
+    setAuthToken,
+    setFakeToken,
+    skipLoginSigning,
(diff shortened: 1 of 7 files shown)


willbach (approved):
code lgtm. Note: have not tested it out.

liam-fitzgerald (approved):

Issue #965, opened 2022-02-13
L1: transfer transaction hangs console


nerveharp:
- Transferred a star on L1 targeting a metamask wallet
- Logged in via MM and accepted the point
- See a near-infinite log of errors; console crashes
- The [transaction completes](https://ropsten.etherscan.io/tx/0xa10a7078aaad433150decc98a062981e0c3daf64c8657338068d1401fafcf952) but the front-end never responds.

Visible on Ropsten; Chrome 98, MacOS.

![image](https://user-images.githubusercontent.com/748181/153744477-f7d81ff9-722c-4bd3-84c2-2ac5ceaf4d81.png)

```
HTTPTransport.js:91          POST https://roller-tmp.urbit.org/v1/roller net::ERR_INSUFFICIENT_RESOURCES
(anonymous) @ HTTPTransport.js:91
(anonymous) @ HTTPTransport.js:46
(anonymous) @ HTTPTransport.js:27
(anonymous) @ HTTPTransport.js:21
i @ HTTPTransport.js:17
t.sendData @ HTTPTransport.js:80
(anonymous) @ RequestManager.js:96
(anonymous) @ RequestManager.js:33
(anonymous) @ RequestManager.js:14
(anonymous) @ RequestManager.js:8
r @ RequestManager.js:4
e.request @ RequestManager.js:83
(anonymous) @ Client.js:109
(anonymous) @ Client.js:33
(anonymous) @ Client.js:14
s @ Client.js:5
Promise.then (async)
l @ Client.js:7
(anonymous) @ Client.js:8
r @ Client.js:4
e.request @ Client.js:100
value @ index.ts:489
getSpawningFor @ index.ts:786
(anonymous) @ useRoller.ts:609
c @ runtime.js:63
(anonymous) @ runtime.js:294
(anonymous) @ runtime.js:119
r @ asyncToGenerator.js:3
o @ asyncToGenerator.js:25
(anonymous) @ asyncToGenerator.js:32
(anonymous) @ asyncToGenerator.js:21
(anonymous) @ useRoller.ts:597
(anonymous) @ useControlledPointsStore.js:52
c @ runtime.js:63
(anonymous) @ runtime.js:294
(anonymous) @ runtime.js:119
r @ asyncToGenerator.js:3
o @ asyncToGenerator.js:25
(anonymous) @ asyncToGenerator.js:32
(anonymous) @ asyncToGenerator.js:21
(anonymous) @ AcceptTransfer.tsx:55
[...]

PR #1052, opened 2022-10-08
accept-transfer: fix infinite render leading to roller request spam

### LINKED ISSUE #709: L1 tx confirmed but progress bar still says "sending..."
This was when logged in with Metamask. The progress bar stayed completely empty while it was awaiting confirmation (presumably due to mm not giving us the data we want until confirmation), but once the tx confirmed, the progress bar went to ~90%, and just... stayed there:

image

No relevant output in the web console.

Update: I can reproduce this consistently for the "accept transfer" flow, but not for other transactions.

pkova:
I recently decided to accept some stars. It was not a great experience, hence this PR.

The root of the problem is that Bridge is polling the roller every second, which together with some bugs relating to reference equality results in a DDOS-esque number of requests to the roller after a the transaction to accept the point completes.

Fixes #709, #965

### DIFF
--- a/src/lib/useRoller.ts
+++ b/src/lib/useRoller.ts
@@ -210,22 +210,25 @@ export default function useRoller() {
       field,
       l1Txn,
       intervalTime = TEN_SECONDS,
+      pointValue,
     }: UpdateParams) => {
+
+      point = point === undefined ? await initPoint(pointValue) : point;
+
       if (l1Txn) storePendingL1Txn(l1Txn);
 
       let interval: any;
       const check = async () => {
-        const updatedPoint = await initPoint(point);
         const changedField = !points[point]
           ? 'newPoint'
-          : updatedPoint.getChangedField(points[point], field);
+          : point.getChangedField(points[point], field);
 
         if (isDevelopment) {
-          console.log(`CHECKING FOR ${changedField} UPDATES:`, updatedPoint);
+          console.log(`CHECKING FOR ${changedField} UPDATES:`, point.patp);
         }
 
-        if (!updatedPoint.isPlaceholder && changedField) {
-          updatePoint(updatedPoint);
+        if (!point.isPlaceholder && changedField) {
+          updatePoint(point);
           if (interval) {
             clearInterval(interval);
           }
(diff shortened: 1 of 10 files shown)


src/store/pointCache.js
```diff
@@ -8,7 +9,7 @@ export function PointCacheProvider({ children }) {
   const cache = usePointStore();
 
   return (
-    <PointCacheContext.Provider value={cache}>
+    <PointCacheContext.Provider value={useDeepEqualReference(cache)}>
```
pkova, line 12: Using reference equality here causes the `AcceptTransfer` component to rerender for no reason every second.

src/views/AcceptTransfer.tsx
```diff
@@ -53,7 +53,7 @@ function useAcceptTransfer() {
     ),
     useCallback(
       () => Promise.all([syncExtras(_point), syncControlledPoints()]),
-      [_point, syncControlledPoints, syncExtras]
+      [_point]
```
pkova, line 56: Having `syncControlledPoints` and `syncExtras` as dependencies causes the crazy amount of HTTP requests, again because of reference equality checks.

src/lib/useRoller.ts
```diff
@@ -210,30 +210,33 @@ export default function useRoller() {
...
+
       if (l1Txn) storePendingL1Txn(l1Txn);
 
       let interval: any;
       const check = async () => {
-        const updatedPoint = await initPoint(point);
         const changedField = !points[point]
           ? 'newPoint'
-          : updatedPoint.getChangedField(points[point], field);
+          : point.getChangedField(points[point], field);
```
pkova, line 224: I also decided to change the `checkForUpdates` api somewhat, since I noticed that after fixing the ddos-bug this interval would keep polling every ten seconds forever, since the point transfer was already present after calling `initPoint`. Callsites can still use the old behavior with `pointValue`, but most of them have access to a proper `Point` object.

tomholford:
@pkova Thank you so much for your contribution. Legendary!

Pulled the branch to try it out locally, but ran into this issue:

image

If this is more than a quick Typescript fix, I can take a look at this later this week.

pkova:
Back in my day Bridge was Javascript!

Fixed the typescript errors, let me know if it works now.

pkova:
I noticed that `tsc` still complains, but it's about files that are unchanged from my stuff.

tomholford (approved):
pulled and tried it out locally, LGTM. Thanks @pkova

Issue #1035, opened 2022-07-16
walletconnect: invalid remainder


Fang-:
User report:

> logging in with WalletConnect via iOS Trust Wallet at first appears to work, but no operations can be performed; last time I got the error "invalid remainder" when I logged in with this method and tried to interact with a lockup contract.
>
> My sense is this is something specific to Bridge's WalletConnect code; I've tried all major browsers across iOS and MacOS, and kept getting a small set of error messages: "invalid remainder" was the most common, but also "problem reading e.length" I believe. Logging in with seed phrase completely bypasses this problem. I have not tried mobile wallets other than Trust Wallet, but it was working with the pre-L2 version of Bridge.

Partial duplicate of #1006, perhaps also resolved by #1008. We should consider just merging & releasing that.

lukestiles:
@Fang- do we have a list of supported browsers and OSes?

tomholford:
@lukestiles As of August 2021, this was the compatibility table I put together when doing a bunch of testing:

![image](https://user-images.githubusercontent.com/16504501/129429905-edc6c123-9ca6-4b2b-9184-60b569bc0e08.png)

tomholford:
Getting additional reports of WalletConnect failures. Here's someone who found a workaround:

![image](https://user-images.githubusercontent.com/16504501/193939399-a7f37216-e10e-4358-ad7a-caec5028aed5.jpeg)

Issue #1006, opened 2022-03-23
WalletConnect in new Bridge no longer works (with Gnosis Safe)


tacryt-socryp:
No transactions are able to go through.

<img width="1191" alt="Screen Shot 2022-03-23 at 4 45 37 PM" src="https://user-images.githubusercontent.com/1377557/159801235-7a4997d2-99c3-42d8-a439-fd97c12e5897.png">

tacryt-socryp:
Logging in to Bridge using WalletConnect now works properly, but now when I try to send a transaction it silently fails with no errors in the console. Nothing at all happens when clicking "Set Networking Keys" button for a star.

nerveharp:
Fixed by urbit/bridge#1008

lukestiles:
Closing as fixed per comment above.

lukestiles:
PR not merged, re-opening

PR #1057, opened 2022-10-14
txns: refactor for EIP1559 compatibility; also upgrade deps, remove dead code, fix race condition

tomholford:
# Context

This PR addresses #1006, #1035 (and potentially other tracked issues) by updating out of date deps and adding support for EIP 1559 Transactions.

While investigating and fixing the root cause, the scope of necessary changes snowballed due to the fact that other interrelated dependencies were out of date. Upgrading these deps subsequently broke the build, as the newer libs use nullish coalescing operators, which the old version of `create-react-app` did not support. I initially spiked into upgrading to CRA 5 but ran into Webpack Config Hell. So instead, I decided to switch the build tooling to Vite, which required some other changes such as migrating all React *.js --> *.jsx, and upgrading commonjs code --> es6 (hence exporting the nockjs code, converting it to es6, and publishing it to [npm](https://www.npmjs.com/package/@urbit/nockjs)).

# Changes

- upgraded deps: web3, @ethereumjs/tx, @ethereumjs/common, @walletconnect/client, @walletconnect/qrcodemodal
- replace create-react-app with Vite
- remove deprecated Hardware wallet login flows and dependencies (since Bridge 3 release, users are directed to Metamask)
- fix race condition in Point Cache that would sometimes crash Bridge on load
- refactor Transaction signing & sending flow to properly handle EIP1559 transactions

# TODO

Test the following flows:
- [x] Metamask
- [x] WalletConnect
- [ ] Ledger
- [ ] Trezor


arthyn (approved):
From what I can tell this looks solid. Do the changes here still support hardware wallets or is that coming later? That's the only thing I couldn't glean that seams important.

tomholford:
> From what I can tell this looks solid. Do the changes here still support hardware wallets or is that coming later? That's the only thing I couldn't glean that seams important.

Thanks. Hardware wallet flows used to be supported explicitly. As of v3, those flows were deprecated and users are expected to use Metamask + HW. The changes in this PR re: HW are to remove dead code that are no longer used.

philipcmonk:
This worked for me with Gnosis Safe, which is new.  Looks good to me.

PR #1069, opened 2022-11-05
NetworkKeys: fix stale keyfile download

pkova:
The roller store wasn't fetching updates to the point after rekeying on L1. [Tested in production](https://network.urbit.org/~batnup-davdys). Fixes #1068.

Issue #1068, opened 2022-11-03
Stale keyfile presented for download after breach


belisarius222:
**Describe the bug**

I recently breached my ship.  After the eth transaction said it was complete, I clicked to download the keyfile, and it downloaded "rovnys-ricfer-4.key".  I then booted a fresh ship, but it couldn't talk to anyone on the network.  I eventually logged back into bridge, clicked to download the keyfile, and it downloaded "rovnys-ricfer-5.key".  I then killed the ghost ship, booted fresh with the new keyfile, and it worked fine.

**To Reproduce**
1. Breach a ship.
2. Click on 'download keyfile'
3. waste some time trying to boot a ship with the stale keyfile
4. download the new keyfile, delete ghost ship, boot fresh ship

**Expected behavior**
The keyfile download should download the new file after a breach, not the old one.

**Desktop (please complete the following information):**
 - OS: MacOS 12.4 
 - Browser: safari

**Login method (please complete the following information):**
master ticket

tomholford:
Fix for this (v3.2.1) is in production

PR #1072, opened 2022-11-09
NetworkKeys: actually fix stale keyfile download

pkova:
My previous PR #1069 fixed the stale keyfile download issue #1068, but introduced a race condition where sometimes the point state would update prematurely, leading to the rekey never completing in the UI.

On deeper investigation I found out that the stale keyfile bug was introduced by me in #1052 when I misunderstood and changed the `checkForUpdates` function. This PR reverts the relevant changes in #1069 and #1052, fixing #1068 once and for all.

### DIFF
--- a/src/lib/useRoller.ts
+++ b/src/lib/useRoller.ts
@@ -213,7 +213,8 @@ export default function useRoller() {
     }: UpdateParams) => {
 
 
-      const p = (typeof point === 'number' ? await initPoint(point) : point) as Point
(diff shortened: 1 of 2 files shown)


tomholford:
@pkova Thanks for the fix. Was this tested on mainnet?

pkova:
Yes.

tomholford (approved):
LGTM, thanks for the fix :100:

PR #1075, opened 2022-11-17
metamask: resolve connected accounts

### LINKED ISSUE #1073: "No accounts connected" for Firefox + Metamask + Ledger
Even though the user has an account connected in MM, the extension returns an empty array when trying to get the connected accounts.

Luckily, Chromium functions as a workaround.

For context:

web+urbitgraph://group/~halbex-palheb/uf-public/graph/~halbex-palheb/general-4066/170141184505929554797657732808611725312

tomholford:
This change adds the recommended approach to get a user's connected account address, as well as fallbacks for older versions of Metamask. This resolves #1073

See:

https://docs.metamask.io/guide/sending-transactions.html#example

https://metamask.zendesk.com/hc/en-us/articles/360053147012-MetaMask-Legacy-Web3

### DIFF
--- a/src/lib/metamask.ts
+++ b/src/lib/metamask.ts
@@ -18,13 +18,22 @@ export const metamaskSignTransaction = async (txn: EIP1559Transaction) => {
 };
 
 const metamaskSendTransaction = async (txn: FakeSignableTx, web3: Web3) => {
-  const accounts = await web3.eth.getAccounts();
-  if(accounts.length === 0) {
-    throw new Error('No accounts connected');
+  let from = null;
+  if(window.ethereum) {
+    from = window.ethereum.selectedAddress;
+    if (!from) {
+      const accounts = await window.ethereum.request({ method: 'eth_requestAccounts' });
+      from = accounts[0] ?? null;
+    }
+  } else if (window.web3) {
+    from = window.web3?.eth?.defaultAccount ?? null;
+  } else {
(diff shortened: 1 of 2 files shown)


src/lib/metamask.ts
```diff
@@ -18,13 +18,22 @@ export const metamaskSignTransaction = async (txn: EIP1559Transaction) => {
...
+      from = accounts[0] ?? null;
+    }
+  } else if (window.web3) {
+    from = window.web3?.eth?.defaultAccount ?? null;
+  } else {
+    const accounts = await web3.eth.getAccounts();
+    from = accounts[0] ?? null;
   }
-  const from = accounts[0];
+
```
arthyn, line 34: should we still throw an error here if `from` is `null`?

src/lib/metamask.ts
```diff
@@ -18,13 +18,22 @@ export const metamaskSignTransaction = async (txn: EIP1559Transaction) => {
 };
 
 const metamaskSendTransaction = async (txn: FakeSignableTx, web3: Web3) => {
-  const accounts = await web3.eth.getAccounts();
-  if(accounts.length === 0) {
-    throw new Error('No accounts connected');
+  let from = null;
+  if(window.ethereum) {
```
arthyn, line 22:
```suggestion
  if (window.ethereum) {
```

PR #1083, opened 2022-12-12
metamask: ensure expected chain ID

### LINKED ISSUE #1082: Confirm network connected to when using Metamask
**Describe the bug**
Currently Metamask users can execute L2 transactions while set to Arbitrum network. 

User ran a transaction to claim the planet twice [successfully, too, since L2 addresses are the same] without realizing.

tomholford:
# Context

Before this change, it was possible to have Metamask connected to another EVM chain (e.g. Optimism), or a testnet (e.g. Goerli) while using Mainnet Bridge. This resolves #1082 by conducting a check for the expected chain ID in the connected wallet during Login, and requesting to update it if necessary.

# Preview

https://user-images.githubusercontent.com/16504501/207049240-fc6689e7-dccf-4601-a759-5b15d1df3c3c.mp4


arthyn (approved):
lgtm

PR #1097, opened 2023-02-16
removing port from suggestions

thomasurbit:


thomasurbit:
This is probably not going to run, so please help me out making this correct. The idea is: get rid of the port suggestion, make Cloud hosting set-up the main way. Below it you will see a link to the CLI install (that's already there).

tomholford:
closed in favor of #1100

PR #1100, opened 2023-03-13
activate: remove references to Port

pkova:
Port is no longer recommended and we should therefore remove it from the activation flow.


tomholford (approved):
LGTM

PR #1107, opened 2023-05-26
WalletConnect: migrate to v2

### LINKED ISSUE #1098: Migrate to WalletConnect v2
WalletConnect is [sunsetting their v1](https://medium.com/walletconnect/weve-reset-the-clock-on-the-walletconnect-v1-0-shutdown-now-scheduled-for-june-28-2023-ead2d953b595) at the end of June 2023. This will effectively stop connect support for any v1 dapp/wallet.

Effective migration is to upgrade the walletConnect hook to use the v2 client.

MarcusMiguel:
Resolves #1098

Should we deploy this already? Some wallets still uses v1 (e.g. MetaMask).

This expects a new `env` variable called `VITE_WALLET_CONNECT_PROJECT_ID`, the project ID can be obtained from https://cloud.walletconnect.com/sign-in.


shawntobin (approved):
Looks good to me.
fwiw I had issues testing with my ledger, but this doesn't seem to be related to the implementation (possibly due to accessing via localhost?). Works fine on the other wallets I tested.
I assume whoever deploys will handle the VITE_WALLET_CONNECT_PROJECT_ID.

tomholford:
FYI WC v1 was sunsetted yesterday. Receiving feedback from users that they are no longer able to connect a wallet. Once tested, this change should be merged and deployed promptly.

src/views/UrbitID/ResetKeys/ResetExecute.tsx
```diff
@@ -127,10 +127,11 @@ export default function ResetExecute({
   );
 
   const performReticket = useCallback(async () => {
+    const { isConnected } = useWalletConnect();
```
dillon, line 130: shouldnt this be outside of the useCallback hook, with `isConnected` in the useCallback dependency array?

dillon (approved):
Looks good, works for me on pilot-mainnet

PR #1108, opened 2023-05-26
NetworkKeys: note potential delay after cycling keys

### LINKED ISSUE #1104: Generating a new keyfile should note potential delay
There's consistent support requests because people try to boot their ship immediately after creating a new keyfile, but I believe it takes some time to become active (at least in the L2 case).  This should be documented on Bridge so they know to be patient.

MarcusMiguel:
Resolves #1104

Displaying a `waitMessage` not only after the user has breached but also after a simple reset. Additionally, the content of the `waitMessage` has been updated to provide more accurate and useful information.

### DIFF
--- a/src/views/UrbitOS/NetworkKeys.tsx
+++ b/src/views/UrbitOS/NetworkKeys.tsx
@@ -311,9 +311,9 @@ export default function UrbitOSNetworkKeys({
 
   const waitMessage = () => (
     <Text as="p" textAlign={'center'}>
-      <strong>Important:</strong> you will need to wait until the next L2 batch
-      is submitted to the roller to boot your ship. If you do not wait, then you
-      will not be able to boot your ship.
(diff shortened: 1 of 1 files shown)


shawntobin (approved):
Looks good

pkova:
@MarcusMiguel the very first time someone accepts an L2 invite we have the following hack in place:

When they get their networking keys, they actually get two keys concatenated into one:

1. The current networking key (belongs to the star).

2. The key that is self-sovereign after the roller batch fires.

We do this so that people don't have to wait for the roller and are able to boot immediately. This does not apply on subsequent rekeys, so we need to make sure this message is only visible on subsequent rekeys and not the initial activation flow.

pkova:
On second thought I think this is fine. The activation flow is completely self-contained and does not enter Bridge proper at all. This screen is not going to be visible for first-time L2 activators.

PR #1109, opened 2023-06-07
proxy: show appropriate actions

### LINKED ISSUE #1101: proxy: show appropriate actions when logged in as management proxy
Cross-posted a user report here:

> trying to set my voting proxy with my management proxy and I get this error:

image

image

> Looks like you actually can't set voting proxy with management proxy, so this is probably a UI problem where this option shouldn't be available.

---

Azimuth only allows certain operations for each proxy:

image

https://operators.urbit.org/manual/id/proxies

So, Bridge should reflect which actions are actually available when logged in as Management Proxy

MarcusMiguel:
Resolves #1101 

Not only the actions available to the managment proxy where incorrect, but there was also issues with the others proxies (e.g., voting proxies wouldn't be displayed at all in the point's list).


src/views/Ops/Ops.tsx
```diff
@@ -67,39 +67,50 @@ export const Ops = () => {
...
+              (point.canVote ? "" : " disabled")}
+            >
+              <Box >
                 <Box className="section-title">Proposals</Box>
                 <Box className="subtitle">
                   View or vote on proposals in the Senate
                 </Box>
               </Box>
-              <Button className="secondary" onClick={goSenate}>
+              <Button className="secondary"
```
shawntobin, line 112: This probably needs to change to canVote

shawntobin (changes requested):

shawntobin (approved):
Looks good to me.

PR #1111, opened 2023-06-27
dev: update L2 local instructions

yosoyubik:
Fakeships don't use the normal codepath triggered by a timer to start receiving azimuth updates and need to manually start the azimuth log retrieval.

There seems to be other issues with the deployment of the contracts/migration scripts (https://github.com/urbit/azimuth/issues/62) but this is necessary for both %azimuth and %roller apps to get data from chain


dillon (approved):
Ah yes this was the missing link to get things running on my end. Working on another PR to push the rest of the fixes for the migration scripts but this is certainly a step in the right direction, thank you.

dillon:
Included in #1113

PR #1112, opened 2023-07-11
Fix eslint / pre-commit hook

dillon:
Missing eslint packages were causing pre-commit hooks to fail.

Fixing the linter also surfaced 2 errors, which were fixed so that things can commits can go through now without `--no-verify`.


tomholford (approved):
LGTM

PR #1114, opened 2023-07-15
separate factory reset flow

### LINKED ISSUE #1046: ux: Factory Reset should be an atomic flow
Currently *Factory Reset* is exposed as opt-in behavior during the *Reset Network Keys* flow:
 
image

Instead it should be its own individual flow.

dillon:
Separates flow for factory reset AKA breach.

closes #1046 

Screenshot 2023-07-14 at 7 36 58 PM
Screenshot 2023-07-14 at 7 37 07 PM
Screenshot 2023-07-14 at 7 37 02 PM

### DIFF
--- a/src/components/L2/Window/BodyPane.scss
+++ b/src/components/L2/Window/BodyPane.scss
@@ -26,6 +26,9 @@
     &.error-text {
       color: $warning-red;
     }
+    &.maxwidth {
+      max-width: 300px;
+    }
   }
 
   .arrow-button {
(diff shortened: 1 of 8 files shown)


shawntobin (commented):
Looks good to me, tested locally with no issues. A couple minor notes:

- Typo in the filename "use**t**SetNetworkKeys.ts"
- We changed the waitMessage text since this PR, so that should probably be updated in this new FactoryReset component.

PR #1115, opened 2023-07-28
Dp/update dev instructions

dillon:
Provides some clarity for running development on your local network


shawntobin (commented):
Looks good.
My only other thought is that we could mention how the roller needs to be set up again if the user runs 'npm run pilot' again, but maybe it's obvious enough

PR #1139, opened 2025-02-12
Fix maxFeePerGas < maxPriorityFeePerGas transaction issue

### LINKED ISSUE #1136: Can't withdraw star from lockup
**Describe the bug**
When I try to withdraw a star from a lockup contract in Bridge, no matter what gas price level I choose, I get this error:
```
maxFeePerGas cannot be less than maxPriorityFeePerGas (The total must be the larger of the two) (tx type=2 hash=not available (unsigned) nonce=4 value=0 signed=false hf=merge maxFeePerGas=902313536 maxPriorityFeePerGas=1000000000)
```

**To Reproduce**
1. Sign in with Metamask
2. Click on Locked Stars (which btw displays the wrong number of available stars, but that's a different problem)
3. Click on Use my address
4. Input any number of stars
5. Select any gas price
6. Click "Withdraw Stars"
7. Observe the error

**Expected behavior**
I expect to be prompted to approve a transaction, but I am not.

**Screenshots**
image

**Desktop:**
 - OS: MacOS 14.4
 - Brave Version 1.68.141 Chromium: 127.0.6533.120 (Official Build) (x86_64) (Shields Down )

**Login method:**
Metamask - 0x1Bef4BaA523d47D218C5c3eFe2624dC2c3a477AF

shawntobin:
resolves #1136

Replaced legacy gas calculation that sometimes caused maxFeePerGas to be less than maxPriorityFeePerGas.

Previous code used web3.eth.getGasPrice() to calculate maxFeePerGas, which isn't suitable anymore for EIP-1559 transactions. Now fetching 'baseFeePerGas' instead.

### DIFF
--- a/src/lib/txn.ts
+++ b/src/lib/txn.ts
@@ -88,11 +88,11 @@ const signTransaction = async ({
   const from = wallet.address;
   const estimate = await estimateGasLimit({ ...txn, from }, web3);
 
-  let block = await web3.eth.getBlock('pending');
+  let block =
+    (await web3.eth.getBlock('pending')) || (await web3.eth.getBlock('latest'));
 
-  // in case pending block doesn't return anything
-  if (!block || !block.baseFeePerGas) {
-    block = await web3.eth.getBlock('latest');
+  if (!block?.baseFeePerGas) {
+    throw new Error('Unable to fetch baseFeePerGas from the network');
   }
 
   const baseFeeWei = block.baseFeePerGas;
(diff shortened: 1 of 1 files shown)


master-malwyl:
One minor suggestion: we might want to add an explicit error case if both 'pending' and 'latest' blocks fail to return baseFeePerGas:

```typescript
let block = await web3.eth.getBlock('pending');

// in case pending block doesn't return anything
if (!block || !block.baseFeePerGas) {
  block = await web3.eth.getBlock('latest');
  if (!block || !block.baseFeePerGas) {
    throw new Error('Unable to fetch baseFeePerGas from network');
  }
}
```
This would make debugging easier if we ever hit an edge case where the network isn't returning proper EIP-1559 data.

Otherwise, the changes look good and should fix the issues reported in #1136.

PR #1143, opened 2025-06-20
Fix #1141 - L1 spawn issues

### LINKED ISSUE #1141: Address shows no Urbit IDs after L1 planet spawn to own address
**Describe the bug**
While connected to Bridge with Metamask which itself was connected to my Trezor HW wallet, I spawned an L1 planet from one of my stars and specified "use my address" for the target eth address. Upon completion I returned to "home" where it lists my IDs and saw the new planet wasn't there. I logged out and logged back in and it said there were no IDs in my account at all, despite there being over 70. It also showed no transactions in history.

Logging out & in again didn't work. Incognito mode didn't work. Clearing all browser history and data didn't work. Unplugging & replugging the trezor didn't work. Switching to an entirely different browser didn't work. Uninstalling & Reinstalling Metamask and reconfiguring it from scratch didn't work.

Some time later (maybe an hour or more) after the millionth trezor replug & relogin attempt it randomly started working again.

**To Reproduce**
Don't know if it's reproducible but:
1. Login with Metamask & Trezor wallet
2. Got to a star
3. Spawn an L1 planet, setting self eth address as target
4. Go back to home, look at list of IDs, see new spawned one's not there
5. logout then log back in again
6. See that it appears you have no IDs in your wallet

**Expected behavior**
The newly spawned planet should have appeared in the list of IDs in the first place, but more importantly it should not have shown there were no IDs in the wallet after logout/login.

**Screenshots**

Image

Image

There was this error in the console, idk if it's related:

Image

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Browser: It initially happened in Chromium but persisted when I switched to Firefox too. I was using Metamask in both browsers, connected to a Trezor HW wallet.

**Login method (please complete the following information):**
How you were logged into bridge when the error occured: Metamask /w Trezor HW wallet.

What Ethereum address were you logged in with: 0xe2e765fA00584cD787dC184019FAe071a63c6E50

**Additional context**
@thelifeandtimes has also encountered this error before

shawntobin:
- The call to the roller was throwing an error for a newly spawned L1 point since it didn't yet know it existed. The try / catch block wasn't set up in a way to properly handle the error, so it wasn't checking the chain when the point couldn't be found in roller state.
- This fix checks the chain when the roller doesn't return the point, and if it exists then it returns the point using on-chain data.

### DIFF
--- a/src/lib/useRoller.ts
+++ b/src/lib/useRoller.ts
@@ -145,16 +145,17 @@ export default function useRoller() {
       }
 
       const pointNum = Number(point);
-      const rawDetails = await api.getPoint(pointNum);
-      const isL2 = isL2Spawn(rawDetails?.dominion);
-
-      const details = toL1Details(rawDetails);
 
       try {
-        if (isL2) {
-          const l2Quota = isL2 ? await api.getRemainingQuota(pointNum) : 0;
-          const l2Allowance = isL2 ? await api.getAllowance(pointNum) : 0;
+        // Try fetching from the roller
+        const rawDetails = await api.getPoint(pointNum); // this fails for newly spawned L1 point until roller updates
+        const details = toL1Details(rawDetails);
+        const isL2 = isL2Spawn(rawDetails?.dominion);
 
+        if (isL2) {
+          // Scenario 1: L2 point exists in roller state - use roller data
+          const l2Quota = await api.getRemainingQuota(pointNum);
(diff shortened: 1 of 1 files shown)


thelifeandtimes:
Thanks for looking into this. Looks like the right solution to me but will defer to @pkova to review/merge/deploy.

